PR #7, opened 2015-06-19
Add a more comprehensive pre-install check

pbrisbin:
This extends our pre-install check (formerly just `which docker`) to also check
for an inability to talk to the docker server and output a more informative
error including boot2docker-specific instructions in that case.

/cc @codeclimate/review


noahd1:
Test and it worked for me.

PR #21, opened 2015-06-23
patch writes to .codeclimate.yml file to always set the file owner as the root directory user

ABaldwinHunter:
@codeclimate/review This PR updates the `init` and `update_yaml` commands to set the `.codeclimate.yml` file `uid` and `gid` to match that of the root directory. 

After some experimenting, I found stubbing the `File.chown()` call in testing somewhat unfeasible. Meaningful tests seemed to require execution with `sudo` prefixed to `bundle exec rake`, which we probably would prefer to avoid. Happy to hear other thoughts though.

### DIFF
--- a/lib/cc/cli/engines/engine_command.rb
+++ b/lib/cc/cli/engines/engine_command.rb
@@ -17,13 +17,11 @@ module CC
         end
 
         def yaml_content
-          File.read(filesystem.path_for(CODECLIMATE_YAML)).freeze
+          filesystem.read_path(CODECLIMATE_YAML).freeze
         end
 
         def update_yaml
-          File.open(filesystem.path_for(CODECLIMATE_YAML), "w") do |f|
-            f.write(parsed_yaml.to_yaml)
-          end
+          filesystem.write_path(CODECLIMATE_YAML, parsed_yaml.to_yaml)
         end
 
         def engine_present_in_yaml?
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -30,7 +30,7 @@ module CC
 
         config["exclude_paths"] = exclude_paths(AUTO_EXCLUDE_PATHS)
 
-        File.write(filesystem.path_for(CODECLIMATE_YAML), config.to_yaml)
+        filesystem.write_path(CODECLIMATE_YAML, config.to_yaml)
       end
 
       def exclude_paths(paths)
--- a/spec/cc/analyzer/filesystem_spec.rb
+++ b/spec/cc/analyzer/filesystem_spec.rb
@@ -2,7 +2,7 @@ require "spec_helper"
 
 module CC::Analyzer
   describe Filesystem do
-    describe "#exists?" do
+    describe "#exist?" do
       it "returns true for files that exist" do
         root = Dir.mktmpdir
         File.write(File.join(root, "foo.rb"), "")
@@ -64,6 +64,17 @@ module CC::Analyzer
 
         filesystem.read_path("foo.rb").must_equal("Foo")
         filesystem.read_path("bar.rb").must_equal("Bar")
+      end
+    end
+
+    describe "#write_path" do
+      it "writes to the filesystem, given a path to a file and content" do
+        filesystem = Filesystem.new(Dir.mktmpdir)
+
+        filesystem.write_path("foo.js", "Hello world")
+
+        filesystem.exist?("foo.js").must_equal(true)
+        filesystem.read_path("foo.js").must_equal("Hello world")
       end
     end
 
--- a/spec/cc/cli/command_spec.rb
+++ b/spec/cc/cli/command_spec.rb
@@ -5,7 +5,7 @@ module CC::CLI
     describe "#require_codeclimate_yml" do
       it "exits if the file doesn't exist" do
         Dir.chdir(Dir.mktmpdir) do
-          stdoit, stderr = capture_io do
+          _, stderr = capture_io do
             lambda { Command.new.require_codeclimate_yml }.must_raise SystemExit
           end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    codeclimate (0.0.5)
+    codeclimate (0.0.7)
       activesupport (~> 4.2, >= 4.2.1)
       codeclimate-yaml (~> 0.0, >= 0.0.2)
       faraday (~> 0.9.1)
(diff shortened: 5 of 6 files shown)


ABaldwinHunter:
Heads up that this PR also updates the `.codeclimate.yml` version in Gemfile.lock.

brynary:
@ABaldwinHunter hmm... any idea why we aren't getting Code Climate results on this PR? Can we check to make sure it's setup properly and also add the badge to the README?

If someone added this repo on their local dev env, it could've broken the integration.

ABaldwinHunter:
@brynary sure. The `.codeclimate.yml` had been empty. I just merged PR updating that so we should get feedback now.

lib/cc/cli/command.rb
```diff
@@ -53,6 +57,14 @@ def rainbow
         @rainbow ||= Rainbow.new
       end
 
+      def uid
+        @uid ||= filesystem.uid
+      end
+
+      def gid
+        @gid ||= filesystem.gid
+      end
```
pbrisbin: `#uid` and `#guid` should be cheap, with `File.stat` being the only (and not really) expensive call. What do you think about `FileSystem#stat` delegating to and memoizing `File.stat`, then just calling these methods when needed?

lib/cc/cli/command.rb
```diff
@@ -43,6 +43,10 @@ def require_codeclimate_yml
         end
       end
 
+      def set_local_owner(file= ".codeclimate.yml")
```
pbrisbin: I don't think there's a lot of value in the default here. It duplicates a constant and may run into issues because we're never calling `FileSystem#path_for`. I would drop it and either always call this method with `filesystem.path_for(CODECLIMATE_YML)` or implement `FileSystem#stat` and have *it* call `#path_for` internally, maybe?

ABaldwinHunter:
@pbrisbin: thanks for feedback. I made the `set_local_owner` method refer only to the path_for(CODECLIMATE_YAML), and created a `stat` method that's memoized in `Filesystem` class, instead of memoizing the `gid` and `uid` calls.

lib/cc/cli/command.rb
```diff
@@ -43,6 +43,10 @@ def require_codeclimate_yml
         end
       end
 
+      def set_local_owner
+        File.chown(uid, gid, filesystem.path_for(CODECLIMATE_YAML))
```
pbrisbin: Sorry, I think this method should still take an argument, I meant to get rid of the defaulting.

lib/cc/cli/command.rb
```diff
@@ -43,6 +43,10 @@ def require_codeclimate_yml
         end
       end
 
+      def set_local_owner
```
pbrisbin: Is there a better name here? The purpose of the method isn't really to "set local owner" (it sets owner and group, and there's only a loose concept of "local"). What about `#correct_permissions`?

lib/cc/cli/engines/engine_command.rb
```diff
@@ -24,6 +24,7 @@ def update_yaml
           File.open(filesystem.path_for(CODECLIMATE_YAML), "w") do |f|
```
pbrisbin: Not something to change now, but worth noting. I think there's an abstraction issue here. We go through `filesystem` for things like `path_for` or `stat`, but then we break the abstraction and call things like `File.open` or `File.write` directly? If `FileSystem` had a `#write` method that would be better, and `#path_for` could probably be made private as it'd only be used internally within (methods like) `FileSystem#write`.

Again, something to consider in a separate PR -- but it provides a little context for my `FileSystem#stat` suggestion.

lib/cc/cli/command.rb
pbrisbin: What do you think about this method living on `FileSystem`?

If we eventually fix the abstraction issue I had mentioned, the calling code would look like (for example):

```rb
filesystem.write(CODECLIMATE_YAML, parsed_yaml)
filesystem.correct_permissions(CODECLIMATE_YAML)
```

And that feels nice to me.

lib/cc/cli/init.rb
```diff
@@ -30,7 +30,8 @@ def create_codeclimate_yaml
 
         config["exclude_paths"] = exclude_paths(AUTO_EXCLUDE_PATHS)
 
-        File.write(filesystem.path_for(CODECLIMATE_YAML), config.to_yaml)
+        filesystem.write(CODECLIMATE_YAML, config.to_yaml)
+        filesystem.correct_permissions(CODECLIMATE_YAML)
```
pbrisbin: Is it likely that we'll want to correct every file we write? I think probably. If so, we could move the `chown` into `FileSystem#write` itself, and remove `#correct_permissions` altogether.

spec/cc/analyzer/filesystem_spec.rb
```diff
@@ -89,5 +89,22 @@ module CC::Analyzer
...
       end
     end
+
+    describe "#write" do
+      it "writes to the filesystem, given a path to a file and content" do
+        root = Dir.mktmpdir
+        Dir.mkdir(File.join(root, "foo"))
+        Dir.mkdir(File.join(root, "bar"))
+        File.write(File.join(root, "foo.rb"), "")
+        File.write(File.join(root, "foo", "bar.rb"), "")
```
pbrisbin: Do these existing files/directories impact the spec at all?

spec/cc/analyzer/filesystem_spec.rb
```diff
@@ -89,5 +89,22 @@ module CC::Analyzer
...
+
+        filesystem = Filesystem.new(root)
+
+        filesystem.write("foo/bar.js", "Hello world")
+
+        filesystem.exist?("foo/bar.js").must_equal(true)
+        filesystem.read_path("foo/bar.js").must_equal("Hello world")
```
pbrisbin: Doh. I didn't realize `#read_path` existed. Please remove the `#read` method I added and use `#read_path` instead.

spec/cc/cli/command_spec.rb
```diff
@@ -5,7 +5,7 @@ module CC::CLI
     describe "#require_codeclimate_yml" do
       it "exits if the file doesn't exist" do
         Dir.chdir(Dir.mktmpdir) do
-          stdoit, stderr = capture_io do
+          stdout, stderr = capture_io do
```
pbrisbin: If this variable is unused (which is likely b/c you renamed it and nothing broke), please call it `_` to convey that.

pbrisbin:
Minor comments, LGTM besides.

PR #27, opened 2015-06-25
Use find to walk directories in init

brynary:
* Existing `file_paths` method was unused... Removed it.
* Renamed `all` as `file_paths` and made it only emit files (not dirs)
* Shell out to `find` instead of using `Dir.glob`... Fixes issue where CTRL+C won't kill an in-process init.
* Exclude `config` and `spec` from analysis

/c @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/filesystem.rb
+++ b/lib/cc/analyzer/filesystem.rb
@@ -23,18 +23,8 @@ module CC
         File.chown(root_uid, root_gid, path_for(path))
       end
 
-      def file_paths
-        Dir.chdir(@root) do
-          Dir["**/*.*"].select { |path| File.file?(path) }.sort
-        end
-      end
-
-      def all
-        @files ||= Dir.chdir(@root) { Dir.glob("**/*") }
-      end
-
       def any?(&block)
-        all.any?(&block)
+        file_paths.any?(&block)
       end
 
       def files_matching(globs)
@@ -46,6 +36,12 @@ module CC
       end
 
       private
+
+      def file_paths
+        @file_paths ||= Dir.chdir(@root) do
+          `find . -type f`.strip.split("\n")
+        end
+      end
 
       def path_for(path)
         File.join(@root, path)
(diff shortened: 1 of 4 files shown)


lib/cc/analyzer/filesystem.rb
```diff
@@ -47,6 +37,12 @@ def files_matching(globs)
 
       private
 
+      def file_paths
+        @file_paths ||= Dir.chdir(@root) do
+          `find . -type f`.strip.split("\n")
```
pbrisbin: Did you find this to be much faster? @noahd1 and I found that it's the actual file access that's slow, not the ruby. In other words when we ran this `find` command in boot2docker on the large test repo, it was just as slow as the `Dir.glob` approach. We weren't rigorous though so I can't cite numbers. It could be faster, it just didn't feel it.

I'm fine making this change, just FYI it might not be the silver bullet.

pbrisbin: You've also now opened yourself up to funny filenaming issues. I'd suggest `-print0` and `split("\0")`.

brynary: @pbrisbin `find` is not intended to be a speed improvement, it's an improvement because `Dir.glob` blocks signal handling, and so you can't stop it with CTRL+C.

I will use `-print0`

pbrisbin: :+1:

lib/cc/analyzer/filesystem.rb
```diff
@@ -47,6 +37,14 @@ def files_matching(globs)
 
       private
 
+      def file_paths
+        @file_paths ||= Dir.chdir(@root) do
+          `find . -type f -print0`.strip.split("\0").map do |path|
+            path.sub(/^\.//, "")
```
pbrisbin: If you use `find *` instead of `find .` you won't need to strip it in ruby.

pbrisbin: There may be caveats, see http://stackoverflow.com/questions/2596462/how-to-strip-leading-in-unix-find/2596474#2596474 for details.

brynary:
@pbrisbin actually, the tests fail. However, Circle CI is not reporting on these PRs for some reason. Adding a card.

PR #30, opened 2015-06-25
Tweak plain text formatter

brynary:
* Always display line numbers, not offsets in plain text formatter (offsets are incomprehensible to humans)
* Don't display the end line if same as the start line
* Various updates to conform with the SPEC

/c @codeclimate/review @GordonDiggs

### DIFF
--- a/lib/cc/analyzer/formatters/formatter.rb
+++ b/lib/cc/analyzer/formatters/formatter.rb
@@ -2,7 +2,8 @@ module CC
   module Analyzer
     module Formatters
       class Formatter
-        def initialize(output = $stdout)
+        def initialize(filesystem, output = $stdout)
+          @filesystem = filesystem
           @output = output
         end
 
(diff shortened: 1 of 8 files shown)


lib/cc/analyzer/formatters/formatter.rb
```diff
@@ -2,7 +2,8 @@ module CC
   module Analyzer
     module Formatters
       class Formatter
-        def initialize(output = $stdout)
+        def initialize(filesystem, output = $stdout)
```
GordonDiggs: Does this change work without any change to the JSON formatter?

lib/cc/analyzer/location_description.rb
```diff
@@ -28,23 +25,20 @@ def to_s
 
       attr_reader :location, :suffix
 
-      def render_lines
-        str = location["lines"]["begin"].to_s
-        str << "-#{location["lines"]["end"]}" if location["lines"]["end"]
-        str
+      def render_lines(begin_line, end_line)
+        if end_line != begin_line
```
GordonDiggs: :lipstick: These conditionals should be swapped so that the positive case is first

brynary:
Replaced by #99

PR #34, opened 2015-06-25
Simplify engine timeout logic

brynary:
/c @noahd1


lib/cc/analyzer/engine.rb
```diff
@@ -35,35 +35,34 @@ def run(stdout_io, stderr_io = StringIO.new)
...
           Analyzer.statsd.increment("cli.engines.result.success")
         else
           Analyzer.statsd.increment("cli.engines.names.#{name}.result.error")
           Analyzer.statsd.increment("cli.engines.result.error")
-
           raise EngineFailure, "engine #{name} failed with status #{status.exitstatus} and stderr #{stderr_io.string.inspect}"
         end
       ensure
+        t_timeout.kill if t_timeout
+
```
GordonDiggs: extra line here

GordonDiggs:
LGTM save for that extra line

pbrisbin:
LGTM, but can someone remind me why we're not using `Timeout.timeout`?

lib/cc/analyzer/engine.rb
brynary: I actually put this here intentionally because I wanted to separate this from the other threads (which are handled differently, and are grouped together)

brynary:
@GordonDiggs I have some ideas, but interested in your take. Can you field @pbrisbin's Q?

GordonDiggs:
I forget the specific reasons, but IIRC it didn't play nice with that `waitpid` call. This `Thread` implementation is nearly identical to how Timeout does it. @noahd1 is working on a proof-of-concept branch now that uses Timeout so we can see if that works.

PR #35, opened 2015-06-26
Alternative approach to handling timeouts for engines

noahd1:
Alternative to own handling of timeouts using stdlib `Timeout::timeout`.

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -17,9 +17,8 @@ module CC
       end
 
       def run(stdout_io, stderr_io = StringIO.new)
+        timed_out = false
         pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command)
-        engine_running = true
-        Analyzer.statsd.increment("cli.engines.started")
 
         t_out = Thread.new do
           out.each_line("\0") do |chunk|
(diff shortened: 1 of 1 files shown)


lib/cc/analyzer/engine.rb
```diff
@@ -35,32 +34,30 @@ def run(stdout_io, stderr_io = StringIO.new)
...
-            Analyzer.statsd.increment("cli.engines.names.#{name}.result.error.timeout")
-            raise EngineTimeout, "engine #{name} ran past #{TIMEOUT} seconds and was killed"
+        begin
+          status = nil
+          Timeout::timeout(TIMEOUT) do
+            pid, status = Process.waitpid2(pid)
           end
+        rescue Timeout::Error
+          run_command("docker kill #{container_name} || true")
+          timed_out = true
```
pbrisbin: Can we get rid of `timed_out` and do whatever happens on `timed_out = true` in the `rescue`? I think the only change in behavior is that `"cli.engines.finished"` wouldn't be ticked in the case of a timeout. If that's important, I'd just duplicate the `increment` call: once in the happy path and again in the timeout rescue.

Also, the `begin`/`rescue` is telling me to extract a `#wait_for_container` method.

noahd1: All good suggestions. Just waiting for someone (@brynary) to tell us why this is a bad idea first.

brynary: In the timeout case, the process does not get cleaned up.

pbrisbin: Calling `waitpid` again after sending the kill should reap the process.

PR #38, opened 2015-06-29
Remove image_id, rename image_name to image

pbrisbin:
The image key represents the argument given to `docker run` to run the 
engine's image. It is equally valid to use a name, an id, or any other 
identifier that `docker run` supports.

This change will allow Builder to pass exact image IDs and coordinate a 
version bump along with any change in the engine images that are run.


GordonDiggs:
LGTM

PR #39, opened 2015-06-29
Raise error if config json is too big

GordonDiggs:
/cc @codeclimate/review 

Example output:

```console
~/git_repos/codeclimate_test (master) 🐖  codeclimate analyze -f json
error: (CC::Analyzer::Engine::EngineFailure) Config for engine rubocop exceeds 64k character limit
```

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -92,12 +92,14 @@ module CC
       end
 
       def env_file
-        if @config_json.size > 64 * 1024
+        contents = "ENGINE_CONFIG=#{@config_json}"
+
+        if contents.size > 64 * 1024
           raise EngineFailure, "Config for engine #{name} exceeds 64k character limit"
         end
 
         path = File.join("/tmp/cc", SecureRandom.uuid)
-        File.write(path, "ENGINE_CONFIG=#{@config_json}")
+        File.write(path, contents)
         path
       end
 
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -92,6 +92,10 @@ module CC
       end
 
       def env_file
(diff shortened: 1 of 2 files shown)


lib/cc/analyzer/engine.rb
```diff
@@ -92,6 +92,10 @@ def docker_run_command
       end
 
       def env_file
+        if @config_json.size > 64 * 1024
```
pbrisbin: The limit is line length, I think it's `"ENGINE_CONFIG=#{@config_json}".size`, right?

GordonDiggs: Oh yeah good call

GordonDiggs: Updated!

lib/cc/analyzer/engine.rb
```diff
@@ -92,8 +92,14 @@ def docker_run_command
       end
 
       def env_file
+        contents = "ENGINE_CONFIG=#{@config_json}"
+
+        if contents.size > 64 * 1024
+          raise EngineFailure, "Config for engine #{name} exceeds 64k character limit"
```
pbrisbin: Do you think there's anything more valuable we can say here? Unfortunately, I'm having trouble thinking of a concrete improvement, but I find this message as-is a little lacking from the end-user's perspective.

GordonDiggs: Agreed. We could suggest excluding fewer files, since that is the main contributor to this size, but that doesn't seem like a good UX either.

If it helps, this will only be around until next week when we move to a different solution.

pbrisbin:
LGTM - I'm fine with the error given that it's temporary and I can't think of anything better. @brynary may have an opinion since it's user facing.

PR #41, opened 2015-06-30
update makefile

ABaldwinHunter:
@codeclimate/review updates make file.


GordonDiggs:
LGTM

pbrisbin:
Eventually we should automate this so updating the `VERSION` file is all that's needed. But until, then LGTM too.

PR #46, opened 2015-07-03
WIP: extract EngineProcess, new timeout logic

pbrisbin:

This is a WIP / Discussion PR. If we think the approach is sound, I will follow
up with better/more tests before submitting the PR as non-WIP for re-review

Motivation:

We've identified a case where our calls to `Thread#join` for the reading
threads (where we're reading stdout/stderr of the engine) is where builder
"hangs".

This could be doing actual work: if an engine generates a lot of output at the
very end (which some are known to do) we may spend a large amount of time after
the engine's finished processing this data.

We call this `#join` and wait for this work unnecessarily in the case of engine
timeout or failure.

With that in mind, this PR intends to:

1. Clarify the timeout logic

  Extract it to its own class and layer the methods such that it's clearer
  what's going on. (This was the intent, I don't claim to have succeeded).

  Use Timeout.timeout. The less we call `Thread.new` directly, the better. If
  the concern is that the spawned process is not cleaned up, that's on us. The
  `rescue` logic I've written should correctly handle process (and thread)
  cleanup.

  I also elected to use `Process.kill` rather than spawning a `docker kill`
  command. I believe this to be equivalent.

1. Include our processing in an overall timeout

  This ensures that if our work (mainly `ConstantAggregator`) is taking too
  long (5 minutes beyond the 15 we give the engine), we timeout rather than
  waiting "forever" and appearing as a "hanging yellow". I'm open to a better
  value here than the arbitrary 5 minutes I picked.

1. Have a single `rescue` with correct cleanup logic

  If we encounter any exception, including either timeout, we kill any running
  threads and kill/clean up the spawned process.

### DIFF
--- a/lib/cc/analyzer/engine_process.rb
+++ b/lib/cc/analyzer/engine_process.rb
@@ -41,18 +41,19 @@ module CC
         pid = read_out = read_err = nil
 
         status = Timeout.timeout(ENGINE_TIMEOUT, engine_timeout_error) do
-          pid, _, out, err = POSIX::Spawn.popen4(*@engine.command)
+          pid, stdin, stdout, stderr = POSIX::Spawn.popen4(*@engine.command)
+          stdin.close
 
           increment("started")
 
           read_out = Thread.new do
-            out.each_line("\0") do |chunk|
+            stdout.each_line("\0") do |chunk|
               @stdout_io.write(chunk.chomp("\0"))
             end
           end
 
           read_err = Thread.new do
-            err.each_line do |line|
+            stderr.each_line do |line|
               @stderr_io.write(line)
             end
           end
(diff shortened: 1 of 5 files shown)


lib/cc/analyzer/engine_process.rb
```diff
@@ -0,0 +1,101 @@
...
+      # timeout for the overall operation as well as runnning the engine.
+      TIMEOUT = ENGINE_TIMEOUT + 5 * 60
+
+
+      def initialize(engine, stdout_io, stderr_io = StringIO.new)
+        @engine = engine
+        @stdout_io = stdout_io
+        @stderr_io = stderr_io
+
+        EngineTimeout.runtime_message = "engine #{@engine.name} ran past #{ENGINE_TIMEOUT} seconds and was killed"
```
pbrisbin: I'm pretty sure this approach won't work, but I have other ideas. I can flesh them out if we green-light this PR in general.

pbrisbin:
Closing this but will leave the branch. I think it's a good refactor, but using `Thread.abort_on_exception` seems to have solved the hung builders for now.

PR #47, opened 2015-07-03
Do no run unregistered engines

ivantsepp:
I was taking a look at https://github.com/codeclimate/codeclimate-fixme and how I could create an engine. I noticed that it has a [`.codeclimate.yml`](https://github.com/codeclimate/codeclimate-fixme/blob/2bea75025808e35182cc574b3ef9672dbe9d2ff0/.codeclimate.yml) file so I attempted to run codeclimate on it. But I received the following error: 

```
codeclimate-fixme[master] $ codeclimate analyze
Starting analysis
Running fixme: Done!
error: (NoMethodError) undefined method `[]' for nil:NilClass
```

First, `codeclimate-fixme` isn't in `config/engines.yml` nor could I find it in the docker registry. Are there plans to add it? 

Second, I guess we should not try to run engines if they're not in `config/engines.yml`. I took a similar approach as [this command](https://github.com/codeclimate/codeclimate/blob/c341ccd457916a2b24a75f18f44bf64d6b8e42ad/lib/cc/cli/engines/install.rb#L24) in that we'll ignore unregistered engines and output a warning.

Let me know what you think!

### DIFF
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -70,14 +70,19 @@ module CC
 
       def engines
         @engines ||= config.engine_names.map do |engine_name|
-          Engine.new(
-            engine_name,
-            registry_entry(engine_name),
-            path,
-            engine_config(engine_name),
-            SecureRandom.uuid
-          )
-        end
+          entry = registry_entry(engine_name)
+          if entry.nil?
+            fatal("unknown engine name: #{engine_name}")
+          else
+            Engine.new(
+              engine_name,
+              entry,
+              path,
+              engine_config(engine_name),
+              SecureRandom.uuid
+            )
+          end
+        end.compact
       end
 
       def formatter
--- a/spec/cc/cli/analyze_spec.rb
+++ b/spec/cc/cli/analyze_spec.rb
@@ -1,0 +1,59 @@
+require "spec_helper"
+
+module CC::CLI
+  describe Analyze do
+    describe "#run" do
+      before { CC::Analyzer::Engine.any_instance.stubs(:run) }
+
+      describe "when no engines are specified" do
+        it "exits and reports no engines are enabled" do
+          within_temp_dir do
+            create_yaml(Factory.create_yaml_with_no_engines)
+
+            stdout, stderr = capture_io do
+              lambda { Analyze.new(args = []).run }.must_raise SystemExit
+            end
+
+            stderr.must_match("No engines enabled. Add some to your .codeclimate.yml file!")
+          end
+        end
+
+        describe "when engine is not in registry" do
+          it "reports that no engines are enabled" do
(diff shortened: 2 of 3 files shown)


pbrisbin:
Hi, thanks for the report and PR!

> First, codeclimate-fixme isn't in config/engines.yml nor could I find it in the docker registry. Are there plans to add it?

If a particular engine is not yet in `engines.yml` that means it's not ready for use . We do plan to incorporate this engine (and many others) when it's done, at which time we'll also add it to `engines.yml`.

> Second, I guess we should not try to run engines if they're not in config/engines.yml

In general, that's correct. An engines developer does need to be able to run engines not yet in `engines.yml` to test their work. For this we have the undocumented (I think?) `--dev` flag which allows you to give any engine name and (if a docker image for it can be found on your system) `codeclimate analyze` will run it regardless of `engines.yml`.

You can see the implementation of that [here](https://github.com/codeclimate/codeclimate/blob/87a57601689d0cf97f3bba6ebe4aac7cec6337a7/lib/cc/cli/analyze.rb#L101).

> ... we'll ignore unregistered engines and output a warning.

Thanks for this PR, we should have time to review it and potentially merge it this week.

ivantsepp:
> We do plan to incorporate this engine (and many others) when it's done, at which time we'll also add it to engines.yml.

Is there an engine underway for scss-lint? If not I'd love to help with that.

GordonDiggs:
@ivantsepp there is not an engine in the works for that yet! Take a look at [the spec](https://github.com/codeclimate/spec) and let me know if you have questions!

lib/cc/cli/analyze.rb
```diff
@@ -70,14 +70,19 @@ def exclude_paths
...
-            engine_name,
-            registry_entry(engine_name),
-            path,
-            engine_config(engine_name),
-            SecureRandom.uuid
-          )
-        end
+          entry = registry_entry(engine_name)
+          if entry.nil?
+            warn("unknown engine name: #{engine_name}")
```
GordonDiggs: We have a [fatal method](https://github.com/codeclimate/codeclimate/blob/f3451ba424525ad60ef3508f81fd2e7e866bb751/lib/cc/cli/command.rb#L35-L38) that I think would be a better fit here

GordonDiggs:
@ivantsepp looks good save for the one comment I added and the failing tests. Ping me when you have a chance to address them and I can merge this! Thanks!

ivantsepp:
@GordonDiggs Made the changes! Not sure what that build failure is about though. Looks like it's something to do with gcloud not authenticating?

GordonDiggs:
Ah good call - I need to address that separately.

GordonDiggs:
@ivantsepp this will go out in the next release!

PR #52, opened 2015-07-08
Abh update wrapper script

ABaldwinHunter:
@codeclimate/review This PR updates the codeclimate-wrapper script to run `docker --tty` only if stdout is a `tty`.

Impact: piping to a file with `codeclimate analyze` will result in plain text. Running `codeclimate analyze` will still produce normal colorful output and spinner.

Sample output post `codeclimate analyze > test.txt`:

terminal: nothing

file: 
screen shot 2015-07-08 at 4 25 59 pm

### DIFF
--- a/codeclimate-wrapper
+++ b/codeclimate-wrapper
@@ -41,10 +41,18 @@ else
     && invalid_docker_host "$DOCKER_HOST"
 fi
 
-exec docker run \
-  --interactive --tty --rm \
-  --env CODE_PATH="$PWD" \
-  --volume "$PWD":/code \
-  --volume /tmp/cc:/tmp/cc \
-  --volume /var/run/docker.sock:/var/run/docker.sock \
-  codeclimate/codeclimate "$@"
+docker_run() {
+  exec docker run \
+    --rm \
+    --env CODE_PATH="$PWD" \
+    --volume "$PWD":/code \
+    --volume /tmp/cc:/tmp/cc \
+    --volume /var/run/docker.sock:/var/run/docker.sock \
+    "$@"
+}
+
+if [ -t 1 ]; then
+  docker_run --interactive --tty codeclimate/codeclimate "$@"
+else
+  docker_run codeclimate/codeclimate "$@"
+fi
--- a/lib/cc/analyzer/formatters/spinner.rb
+++ b/lib/cc/analyzer/formatters/spinner.rb
@@ -7,6 +7,7 @@ module CC
         end
 
         def start
+          return unless $stdout.tty?
           @thread = Thread.new do
             loop do
               @spinning = true


codeclimate-wrapper
```diff
@@ -41,10 +41,20 @@ else
     && invalid_docker_host "$DOCKER_HOST"
 fi
 
-exec docker run \
+if [ -t 1 ]; then
+  exec docker run \
   --interactive --tty --rm \
```
pbrisbin: Can you shift these two spaces so the continuations are still indented beyond what they continue?

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -56,7 +56,11 @@ def finished
 
         def engine_running(engine)
           @active_engine = engine
-          with_spinner("Running #{engine.name}: ") do
+          if $stdout.tty?
+            with_spinner("Running #{engine.name}: ") do
```
GordonDiggs: What do you think of putting this logic inside of `Spinner`? The `start` method could become:

        def start
          if $stdout.tty?
            @thread = Thread.new do
              loop do
                @spinning = true
                spinner.spin
                sleep 0.075
              end
            end
          end
        end

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -56,7 +56,11 @@ def finished
...
+              yield
+            end
+          else
             yield
```
pbrisbin: I think I suggested this before, but you can change the syntax here:

```rb
def engine_running(engine, &block)
  # ...
  with_spinner("...", &block)
end
```

In this case I think it'd help because you're pushing another level of indentation.

codeclimate-wrapper
```diff
@@ -41,10 +41,20 @@ else
...
   --volume /var/run/docker.sock:/var/run/docker.sock \
   codeclimate/codeclimate "$@"
+else
+  exec docker run \
+    --rm \
+    --env CODE_PATH="$PWD" \
+    --volume "$PWD":/code \
+    --volume /tmp/cc:/tmp/cc \
+    --volume /var/run/docker.sock:/var/run/docker.sock \
+    codeclimate/codeclimate "$@"
```
pbrisbin: I'm wondering if it's worth DRYing this so the two don't drift out of since. One way to do that would be to exctract a function:

```sh
docker_run() {
  exec docker run --all --the --shared --options "$@"
}

if ...; then
  docker_run --interactive --tty codeclimate/codeclimate "$@"
else
  docker_run codeclimate/codeclimate "$@"
fi
```

WDYT?

GordonDiggs:
This is a nice solution - much better than turning Rainbow on and off

lib/cc/analyzer/formatters/plain_text_formatter.rb
ABaldwinHunter: that makes sense. Added that refactor

lib/cc/analyzer/formatters/spinner.rb
```diff
@@ -7,11 +7,13 @@ def initialize(text)
         end
 
         def start
-          @thread = Thread.new do
-            loop do
-              @spinning = true
-              spinner.spin
-              sleep 0.075
+          if $stdout.tty?
+            @thread = Thread.new do
```
pbrisbin: Does something try to kill this thread later? Is it a problem that `@thread` will be `nil` in the non-tty case?

lib/cc/analyzer/formatters/spinner.rb
```diff
@@ -7,11 +7,13 @@ def initialize(text)
         end
 
         def start
-          @thread = Thread.new do
-            loop do
-              @spinning = true
-              spinner.spin
-              sleep 0.075
+          if $stdout.tty?
```
pbrisbin: I support rubocop in its `return unless $stdout.tty?` suggestion, FWIW.

lib/cc/analyzer/formatters/spinner.rb
GordonDiggs: @pbrisbin it happens in `stop` below but is guarded

codeclimate-wrapper
ABaldwinHunter: sure.

ABaldwinHunter:
@pbrisbin @GordonDiggs Thanks for the feedback! And @brynary thanks for the feedback earlier. 

I believe I responded to all comments: 
1) updated wrapper script to abstract out a`docker_run` method
2) move conditional spinner use out of `formatter` into `spinner#start` method
3) refactor unless statement

pbrisbin:
What a great diff! LGTM

codeclimate-wrapper
```diff
@@ -41,10 +41,18 @@ else
...
-  --volume /var/run/docker.sock:/var/run/docker.sock \
-  codeclimate/codeclimate "$@"
+docker_run() {
+  exec docker run \
+    --rm \
+    --env CODE_PATH="$PWD" \
+    --volume "$PWD":/code \
+    --volume /tmp/cc:/tmp/cc \
+    --volume /var/run/docker.sock:/var/run/docker.sock \
+    "$@"
```
GordonDiggs: Have you tested that this works correctly? My understanding was that wrapping this in quotes makes it one argument

ABaldwinHunter: I tested locally by running `./codeclimate-wrapper analyze > test.txt` and `./codeclimate-wrapper analyze`, and both worked correctly. That said, I'm pretty new to bash scripting - mostly started with Pat's suggestion above - so welcome other suggestions!

GordonDiggs: Ok cool - if it works then I'm good!

GordonDiggs:
LGTM!

codeclimate-wrapper
```diff
@@ -41,10 +41,18 @@ else
...
+}
+
+if [ -t 1 ]; then
+  docker_run --interactive --tty codeclimate/codeclimate "$@"
```
brynary: Hmm, `--interactive` is about _STDIN_, not _STDOUT_. I wonder if it should not be affected.

ABaldwinHunter: @brynary I think you're correct. I could make that change and bump version?

brynary: Sounds good

PR #53, opened 2015-07-08
always keep --interactive as arg in docker run command

ABaldwinHunter:
@codeclimate/review @brynary PR refactors wrapper script to keep `--interactive` as arg in docker run command no matter what. (Since it's related to `STDIN`, I shouldn't have made its inclusion conditional in recent update to wrapper script.)

Tested locally by running `./codeclimate-wrapper analyze` and `./codeclimate-wrapper analyze > test.txt`

### DIFF
--- a/codeclimate-wrapper
+++ b/codeclimate-wrapper
@@ -43,7 +43,7 @@ fi
 
 docker_run() {
   exec docker run \
-    --rm \
+    --interactive --rm \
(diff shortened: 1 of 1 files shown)


GordonDiggs:
LGTM

pbrisbin:
I wouldn't call it a "refactor", but LGTM ;)

PR #55, opened 2015-07-13
update gem file, particularly cc-yaml version bump

ABaldwinHunter:
@codeclimate/review Bump `cc-yaml` version to latest release, to get updated warnings for incompatibility of `languages` and `engines` keys.


GordonDiggs:
:+1:

PR #56, opened 2015-07-13
bump to cc-yaml version 0.0.7 - fix languages key parsing

ABaldwinHunter:
@codeclimate/review bumps to `cc-yaml 0.0.7`, which fixes `languages` parsing to use traditional format:

```
languages:
  Ruby: true
  JavaScript: true
```


ABaldwinHunter:
see https://github.com/codeclimate/codeclimate/pull/57 for updates to gem file

PR #58, opened 2015-07-16
bump cc-yaml to v0.0.8

ABaldwinHunter:
@codeclimate/review bumps `cc-yaml` to `v0.0.8`, [which] (https://github.com/codeclimate/codeclimate-yaml/commit/03b144bfd1b49b17ba4049b5957c93dbd773beb6):

1) Adds warning if no `engines` or `languages` key is found.

```
$ codeclimate validate-config
WARNING: No languages or engines key found. Must have analysis key. 
```
2) Updates language in warning for when both `engines` and `languages` keys are present.

old warning:
```
$ codeclimate validate-config
WARNING: engines key already found, dropping key: languages. Analysis settings for Languages and 
Engines are both valid but mutually exclusive. Note: command line analysis requires an Engines 
configuration.
```
new warning:
```
$ codeclimate validate-config
WARNING: Use either a Languages key or an Engines key, but not both. They are mutually exclusive.
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     codeclimate (0.0.25)
       activesupport (~> 4.2, >= 4.2.1)
-      codeclimate-yaml (~> 0.0, >= 0.0.7)
+      codeclimate-yaml (~> 0.0, >= 0.0.8)
       faraday (~> 0.9.1)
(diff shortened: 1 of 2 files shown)


ABaldwinHunter:
The error parsing doesn't work appropriately. Fixing bug.

Issue #60, opened 2015-07-20
-f mode is exclusive in option parsing in analyze command


bobbytables:
Hi!

I'm trying to use the `--dev` option with the `-f` flag, but it seems they are mutually exclusive. I did some code spelunking and found this:

https://github.com/codeclimate/codeclimate/blob/master/lib/cc/cli/analyze.rb#L38

It seems if you pass the `-f` flag, the `--dev` flag is completely ignored.

PR #61, opened 2015-07-20
Apply all arguments to analyze command.

bobbytables:
This is an attempt to fix: https://github.com/codeclimate/codeclimate/issues/60

### DIFF
--- a/spec/cc/cli/analyze_spec.rb
+++ b/spec/cc/cli/analyze_spec.rb
@@ -2,6 +2,20 @@ require "spec_helper"
 
 module CC::CLI
   describe Analyze do
+    describe "#initialize" do
+      it "sets dev mode for --dev" do
+        instance = Analyze.new(["--dev"])
+
+        instance.dev_mode?.must_equal true
+      end
+
+      it "sets dev mode with formatters defined" do
+        instance = Analyze.new(%w[-f json --dev])
+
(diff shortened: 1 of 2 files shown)


spec/cc/cli/analyze_spec.rb
```diff
@@ -2,6 +2,21 @@
 
 module CC::CLI
   describe Analyze do
+    describe '#initialize' do
+      it 'sets dev mode for --dev' do
+        instance = Analyze.new(['--dev'])
+
+
+        assert instance.dev_mode?, "in dev mode"
```
GordonDiggs: We use the `must_` syntax - can you reformat these lines? Should be `instance.dev_mode?.must_equal true`

pbrisbin:
This LGTM with two notes:

- We double-quote all strings (also, `%w[ ]` might be good for the spec arguments)
- Adding `#dev_mode?` only for the spec is a smell to me, but we have plans to upgrade to a real options parser soon, so I'm OK not worrying about it for now.

/cc @GordonDiggs WDYT?

GordonDiggs:
@pbrisbin Agreed!

bobbytables:
@pbrisbin @GordonDiggs - Style comments applied. I agree `dev_mode?` is an odd method but it allowed me quickly try this out.

A real option parser would help these problems.

PR #64, opened 2015-07-22
Refactors during implementation of ignore-check

pbrisbin:
These are some behavior-neutral changes I have in my branch for ignoring
checks. I'm opening this separately so the eventual diff for the feature is
clearer.

- Change whitespace in JSONFormatter

  The document itself doesn't use spaces between tokens, so we shouldn't have
  whitespace after/before the overall brackets.

- Fix JSONFormatter spec
  - Include Factory to cut down on noise
  - Actually send in empty data for empty data spec
  - Use must_equal not must_match which always succeeds if given the empty
    string (causing pass-by-coincidence)

- Centralize engine_running behavior into superclass
  - Sub classes shouldn't need to duplicate this
  - Ensures a consistent instance variable (current_engine) is used
  - Provide an attr_reader for it
  - Paves the way for formatter-centralized issue exlusion based on engine
    configuration

/cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/formatters/formatter.rb
+++ b/lib/cc/analyzer/formatters/formatter.rb
@@ -2,7 +2,6 @@ module CC
   module Analyzer
     module Formatters
       class Formatter
-        attr_reader :current_engine
 
         def initialize(output = $stdout)
           @output = output
@@ -28,6 +27,10 @@ module CC
         end
 
         InvalidFormatterError = Class.new(StandardError)
+
+        private
+
+        attr_reader :current_engine
       end
     end
   end
--- a/lib/cc/analyzer/formatters/formatter.rb
+++ b/lib/cc/analyzer/formatters/formatter.rb
@@ -2,6 +2,8 @@ module CC
   module Analyzer
     module Formatters
       class Formatter
+        attr_reader :current_engine
+
         def initialize(output = $stdout)
           @output = output
         end
@@ -13,7 +15,10 @@ module CC
         end
 
         def engine_running(engine)
+          @current_engine = engine
           yield
+        ensure
+          @current_engine = nil
         end
 
         def finished
--- a/lib/cc/analyzer/formatters/formatter.rb
+++ b/lib/cc/analyzer/formatters/formatter.rb
@@ -2,7 +2,6 @@ module CC
   module Analyzer
     module Formatters
       class Formatter
-
         def initialize(output = $stdout)
           @output = output
         end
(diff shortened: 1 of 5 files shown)


lib/cc/analyzer/formatters/formatter.rb
```diff
@@ -2,6 +2,8 @@ module CC
   module Analyzer
     module Formatters
       class Formatter
+        attr_reader :current_engine
```
GordonDiggs: Should this be `protected` so that subclasses can get it, but outsiders can't?

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -14,8 +14,8 @@ def started
         def write(data)
           if data.present?
             json = JSON.parse(data)
-            if @active_engine
-              json["engine_name"] = @active_engine.name
+            if current_engine
```
GordonDiggs: hah should we remove this conditional re: conversation yesterday?

pbrisbin: Yeah, I think I will in the feature branch. Leaving it seems more in the spirit of behavior-neutral for this PR.

lib/cc/analyzer/formatters/formatter.rb
pbrisbin: Ruby's method security is weird and that's not what `protected` means. But it could be `private`, sure.

GordonDiggs: I thought `private` stuff wasn't accessible to subclasses but I could have been wrong.

My main goal here is that this is accessible to subclasses but not outside.

pbrisbin: Nope, that's the traditional definition of these keywords. In ruby `private` just means you can't call it with a receiver (`foo` works, `anything.foo`, `self.foo` errors). This means within a subclass you can call `current_engine` which is private and it works because there's no method receiver.

`protected` means you can only call it with a receiver that is your or your subclasses. It's meant for this scenario:

```rb
class Foo
  def ==(other)
    name == other.name
  end

  protected

  # this should be private, but because we need to invoke other.name 
  # within #==, we use protected
  attr_reader :name
end
```

> My main goal here is that this is accessible to subclasses but not outside.

Yup, done.

GordonDiggs: Ah I get that now  - thanks for the explanation!

pbrisbin:
Pulled some similar changes over, this time for the PlainTextFormatter

lib/cc/analyzer/formatters/plain_text_formatter.rb
pbrisbin: Went ahead with this anyway now that I refactored the spec such that it doesn't matter

GordonDiggs:
LGTM

PR #65, opened 2015-07-22
Add EngineOutputFilter, filter by check name

pbrisbin:
Engines are instantiated with a string a json that corresponds to a modified
version of the YAML object under the engines key for the given engine. Within
this object, we'll now support a "checks" key that can hold check-specific
configuration including an enabled boolean.

This commit introduces an EngineOutputFilter that accepts this engine 
configuration and decides if the output of the engine should be written to the
formatter.

There is a lot of JSON generation and re-parsing going on that can seem 
redundant; I'm leaving this in place for now because orchestrating the various
parts that would need to change to address that is difficult.

One possible approach would be to enrich the interfaces involved and not pass
strings around so much. Engine would take a Hash of configuration and
stringify it only for the purposes of writing the config.json file. Formatter
could lose it's IO-like write interface and receive more structured data
centralizing the output parsing and filtering further upstream.

/cc @codeclimate/review

Note: the refactor PR from earlier was actually in support of an alternative
approach that I didn't end up going with. Oh well, I still think it was
valuable.

### DIFF
--- a/lib/cc/analyzer/engine_output_filter.rb
+++ b/lib/cc/analyzer/engine_output_filter.rb
@@ -24,7 +24,7 @@ module CC
       end
 
       def issue?(json)
-        json.is_a?(Hash) && json["type"] == ISSUE_TYPE
+        json["type"] == ISSUE_TYPE
       end
 
       def ignore_issue?(json)
--- a/lib/cc/analyzer/engine_output_filter.rb
+++ b/lib/cc/analyzer/engine_output_filter.rb
@@ -32,8 +32,8 @@ module CC
       end
 
       def check_config(check_name)
-        checks = @config.fetch("checks", {})
-        checks.fetch(check_name, {})
+        @checks ||= @config.fetch("checks", {})
+        @checks.fetch(check_name, {})
       end
     end
   end
--- a/lib/cc/analyzer/engine_output_filter.rb
+++ b/lib/cc/analyzer/engine_output_filter.rb
@@ -1,0 +1,40 @@
+module CC
+  module Analyzer
+    class EngineOutputFilter
+      ISSUE_TYPE = "issue".freeze
+
+      def initialize(config = {})
+        @config = config
+      end
+
+      def filter?(output)
+        if (json = parse_as_json(output))
+          issue?(json) && ignore_issue?(json)
+        else
+          false
+        end
+      end
+
+      private
+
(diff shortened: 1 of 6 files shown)


pbrisbin:
Going to ignore the CC check, it's `Engine` being too long, something this PR doesn't contribute to in a meaningful way IMO.

lib/cc/analyzer/engine_output_filter.rb
```diff
@@ -0,0 +1,40 @@
...
+      private
+
+      def parse_as_json(output)
+        JSON.parse(output)
+      rescue JSON::ParserError
+        nil
+      end
+
+      def issue?(json)
+        json.is_a?(Hash) && json["type"] == ISSUE_TYPE
```
GordonDiggs: When would the first condition be false?

lib/cc/analyzer/engine_output_filter.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      def ignore_issue?(json)
+        !check_config(json["check"]).fetch("enabled", true)
+      end
+
+      def check_config(check_name)
+        checks = @config.fetch("checks", {})
```
GordonDiggs: Could be good to memoize this since this object is somewhat long running.

lib/cc/analyzer/engine_output_filter.rb
pbrisbin: If the output is a JSON array (which is valid for JSON). See [here](https://github.com/codeclimate/codeclimate/pull/65/files#diff-1a8dd8bd9de2daee94952f7f6d88199eR15) for a test that fails without this condition in place.

GordonDiggs: That's valid for JSON, but not our spec, right? Is there a situation where an engine will output an array?

lib/cc/analyzer/engine_output_filter.rb
pbrisbin: Good call, done.

lib/cc/analyzer/engine_output_filter.rb
pbrisbin: Yeah, I guess that's true. If the filter doesn't blow up on that, the formatter would anyway. Will remove.

GordonDiggs:
LGTM!

PR #67, opened 2015-07-23
Extract shareable classes, improve interfaces

pbrisbin:
- Extract EnginesRunner
- Remove unused EngineProcess autoload
- Engine accepts non-string json configuration
- Sharable PathPatterns class
- Support dev_mode directly in EngineRegistry

/cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -5,8 +5,8 @@ module CC
     autoload :Engine,             "cc/analyzer/engine"
     autoload :EngineClient,       "cc/analyzer/engine_client"
     autoload :EngineOutputFilter, "cc/analyzer/engine_output_filter"
-    autoload :EngineProcess,      "cc/analyzer/engine_process"
     autoload :EngineRegistry,     "cc/analyzer/engine_registry"
+    autoload :EnginesRunner,      "cc/analyzer/engines_runner"
     autoload :Filesystem,         "cc/analyzer/filesystem"
     autoload :Formatters,         "cc/analyzer/formatters"
     autoload :IssueSorter,        "cc/analyzer/issue_sorter"
@@ -17,12 +17,20 @@ module CC
     autoload :UnitName,           "cc/analyzer/unit_name"
 
     class DummyStatsd
-      def method_missing(*args)
+      def method_missing(*)
+        yield if block_given?
+      end
+    end
+
+    class DummyLogger
+      def method_missing(*)
         yield if block_given?
       end
     end
 
     cattr_accessor :statsd
+    cattr_accessor :logger
     self.statsd = DummyStatsd.new
+    self.logger = DummyLogger.new
   end
 end
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -28,8 +28,7 @@ module CC
       end
     end
 
-    cattr_accessor :statsd
-    cattr_accessor :logger
+    cattr_accessor :statsd, :logger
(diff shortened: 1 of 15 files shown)


brynary:
@pbrisbin can we fixup the codeclimate issues detected (with the possible exception of the "long method" one... that's more subjective)?

pbrisbin:
Always.

lib/cc/cli/analyze.rb
```diff
@@ -1,11 +1,9 @@
-require "securerandom"
-
 module CC
   module CLI
     class Analyze < Command
       include CC::Analyzer
 
-      def initialize(args = [])
+      def initialize(_args = [])
```
GordonDiggs: What's the value of this change?

lib/cc/analyzer.rb
```diff
@@ -5,23 +5,32 @@ module Analyzer
...
+    end
+
+    class DummyLogger
+      def method_missing(*)
         yield if block_given?
       end
     end
 
     cattr_accessor :statsd
+    cattr_accessor :logger
```
GordonDiggs: Any reason not to declare these on the same line?

lib/cc/cli/analyze.rb
pbrisbin: It was a drive-by fix: `args` is unused, if this were written today, it would be a rubocop issue.

lib/cc/analyzer.rb
pbrisbin: No reason, didn't think about it.

lib/cc/analyzer/engines_runner.rb
```diff
@@ -0,0 +1,80 @@
...
+          engine_config = config.merge!(exclude_paths: exclude_paths)
+
+          Engine.new(name, metadata, @source_dir, engine_config, label)
+        end
+      end
+
+      def enabled_engines
+        {}.tap do |ret|
+          @config.engines.each do |name, config|
+            if config["enabled"]
```
pbrisbin: I think this should be `config.enabled?`, it's a true `CC::Yaml` object now.

lib/cc/analyzer/path_patterns.rb
```diff
@@ -0,0 +1,26 @@
...
+      private
+
+      def expand
+        results = Dir.chdir(@root) do
+          @patterns.map do |pattern|
+            Dir.glob(pattern)
+          end
+        end
+
+        results.flatten.sort.uniq
```
pbrisbin: Could remove `flatten` by using `flat_map` above.

brynary: Someone should add a rubocop check for flat_map :)



Typed with thumbs.

lib/cc/analyzer/engine_registry.rb
```diff
@@ -3,18 +3,29 @@
...
-      def initialize
+      def initialize(dev_mode = false)
         @path = File.expand_path("../../../../config/engines.yml", __FILE__)
         @config = YAML.safe_load_file(@path)
+        @dev_mode = dev_mode
       end
 
       def [](engine_name)
-        @config[engine_name]
+        if dev_mode?
```
GordonDiggs: Just putting this out there, but I don't think it warrants a change just yet. I'm nervous about this being included in builder because of the potential for arbitrary docker containers to be run. WDYT of overriding this method within builder when we include this to always be false?

GordonDiggs:
LGTM

PR #69, opened 2015-07-24
WIP: Replace custom Config object with CC::Yaml

pbrisbin:
All codeclimate.yml access should go through CC::Yaml to get the proper 
validations and content access. Custom Config objects and direct hash access
should be avoided at all costs.

Included spec improvements:

- Centralize repeated within_temp_dir definitions in helper
- Centralize repeated create_yml definitions in helper
- Replace Factory MysteryGuests with in-place HEREDOCs

/cc @codeclimate/review

I'm opening this as WIP to make it visible so someone else doesn't duplicate
this work by mistake. It is close enough that reviewing it in this WIP state
would still be valuable (I don't have any big changes planned).

TODO:

- config.foo = bar, or config["foo"] = bar explodes with an unknown class error
  meaning the in-place config updates don't work yet. This will probably have
  to be fixed in cc-yaml (sigh).
- Analyze blows up if given an entirely empty config. It's not clear if this
  should be a validation error or no-enabled-engines error, either way it
  should probably be better handled.
- Related, should the require_codeclimate_yaml assertion used for the engines
  and analyze commands require a *valid* codeclimate.yaml too?


lib/cc/cli/analyze.rb
```diff
@@ -83,7 +73,7 @@ def engines
               engine_name,
               entry,
               path,
-              engine_config(engine_name),
+              config.merge!(exclude_paths: exclude_paths).to_json,
```
GordonDiggs: Does this actually mutate the `CC::Yaml` object we pass around? Would it be better to use `merge` here?

lib/cc/cli/engines/engine_command.rb
```diff
@@ -12,32 +11,29 @@ def engine_name
...
-        def engine_enabled?
-          parsed_yaml.engine_enabled?(engine_name)
+        def write_config
+          filesystem.write_path(CODECLIMATE_YAML, config.as_json.to_yaml)
         end
 
-        def engine_exists?
-          engines_registry_list.keys.include?(engine_name)
+        def valid_engine?(name = engine_name)
+          registry.list.keys.include?(name)
```
GordonDiggs: `registry.list[name].present?` should be a little more performant here. I realize you didn't make the change but since you're already changing this line.

lib/cc/cli/engines/install.rb
```diff
@@ -27,15 +27,13 @@ def pull_docker_images
         end
 
         def engine_names
-          @engine_names ||= parsed_yaml.engine_names
-        end
-
-        def engine_exists?(engine_name)
-          engines_registry_list.keys.include?(engine_name)
+          p config.engines
```
GordonDiggs: Should these `p` statements be here?

lib/cc/cli/engines/remove.rb
```diff
@@ -7,13 +7,12 @@ class Remove < EngineCommand
         def run
           require_codeclimate_yml
 
-          if !engine_exists?
+          if !valid_engine?
             say "Engine not found. Run 'codeclimate engines:list' for a list of valid engines."
```
GordonDiggs: hmm should this and the one on line 13 be `fatal`?

lib/cc/cli/analyze.rb
pbrisbin: I have no idea! Agreed merge is probably better either way

lib/cc/cli/engines/install.rb
pbrisbin: Nope. Just putting the p in WIP

lib/cc/cli/engines/remove.rb
pbrisbin: Had the same thought and no strong opinion so left it as is. Fatal makes sense to me.

lib/cc/cli/engines/enable.rb
```diff
@@ -27,7 +26,10 @@ def pull_docker_images
         end
 
         def enable_engine
-          parsed_yaml.enable_engine(engine_name)
+          config.engines[engine_name] ||= {}
+          config.engines[engine_name].enabled = true
```
ABaldwinHunter: It looks like right now our `codeclimate-yaml` gem throws an error when this is called. It's:
`NameError: uninitialized constant CC::Yaml::Parser::Ruby`

shows up in [two] (https://github.com/codeclimate/codeclimate-yaml/blob/master/lib/cc/yaml/nodes/sequence.rb#L97) [spots] (https://github.com/codeclimate/codeclimate-yaml/blob/master/lib/cc/yaml/nodes/mapping.rb#L80)

I think it's a mistake in `cc/yaml`. Looking into fixing.

brynary:
@pbrisbin where did this shake out?

pbrisbin:
It's probably mergeable (after dealing with the conflicts), the major blocking TODO has been resolved. We've just been too strapped for time to pick it back up. I'll close and re-open later.

brynary:
Can you describe briefly in layman's terms the value of this?

pbrisbin:
We have a lot of desperate places across our systems that need content that lives in the codeclimate.yml. I've said, and I know I've heard you say, basically, "anywhere that needs to read codeclimate.yml should go through this gem". It doesn't make sense (to me) to do anything else. The only reason that's not the case here already is that the gem didn't exist when it was written.

When places (like this one) do their own `Yaml.parse` instead of using the gem you can have all sorts of problems and unexpected behavior.

- Using data from an unvalidated config for unexpected results
- Forgetting to use `SafeYaml` in your custom reader
- Getting no-method for `nil` instead of meaningful errors the gem provides on mis-use
- In one place "no engines" might mean `[]` but in another it means `nil`
- New developers don't realize the gem exists and start building custom validation and warning logic that duplicates what the gem already provides
- We find a bug in cc-yaml logic and fix the gem, but that logic is duplicated here and goes unfixed
- etc, etc

Some of these have happened and some of these are just general reasons for DRYing code.

brynary:
Makes perfect sense. Thanks.

PR #74, opened 2015-07-31
Temporarily handle codeclimate-yaml's config default

GordonDiggs:
/cc @codeclimate/review 

Eventually, all engines will be updated to use the new hash format, but until then we need to pass a string `config` value to engines if that is the only thing present.


pbrisbin:
LGTM

PR #75, opened 2015-07-31
invalid engines

ABaldwinHunter:
@codeclimate/review 

This PR adds check for invalid engine names to `codeclimate validate-config`

Its implementation is to create a `validate` subcommand under `enges` (which means that users can also run `codeclimate engines:validate`) which I kind of like.

Feedback welcome.

### DIFF
--- a/lib/cc/cli/engines.rb
+++ b/lib/cc/cli/engines.rb
@@ -9,6 +9,7 @@ module CC
       autoload :Install, "cc/cli/engines/install"
       autoload :List, "cc/cli/engines/list"
       autoload :Remove, "cc/cli/engines/remove"
+      autoload :Validate, "cc/cli/engines/validate"
     end
   end
 end
--- a/lib/cc/cli/engines/validate.rb
+++ b/lib/cc/cli/engines/validate.rb
@@ -1,0 +1,36 @@
+require "cc/analyzer"
+
+module CC
+  module CLI
+    module Engines
+      class Validate < EngineCommand
+        def run
+          if invalid_engines.any?
+            warn
+            true
+          end
+        end
+
+        private
+
+        def warn
+          invalid_engines.each do |engine|
+            puts colorize("WARNING: unknown engine <#{engine}>", :red)
+          end
+        end
+
+        def invalid_engines
+          @invalid_engines ||= engine_names.reject { |engine_name| engine_exists? engine_name }
+        end
+
+        def engine_names
+          @engine_names ||= parsed_yaml.engine_names
+        end
+
+        def engine_exists?(engine_name)
+          engines_registry_list.keys.include?(engine_name)
+        end
+      end
+    end
+  end
+end
--- a/lib/cc/cli/validate_config.rb
+++ b/lib/cc/cli/validate_config.rb
@@ -17,7 +17,9 @@ module CC
         if any_issues?
           display_issues
         else
-          puts colorize("No errors or warnings found in .codeclimate.yml file.", :green)
+          unless validate_engines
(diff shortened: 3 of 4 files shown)


lib/cc/cli/engines.rb
```diff
@@ -9,6 +9,7 @@ module Engines
       autoload :Install, "cc/cli/engines/install"
       autoload :List, "cc/cli/engines/list"
       autoload :Remove, "cc/cli/engines/remove"
+      autoload :Validate, "cc/cli/engines/validate"
```
pbrisbin: I thought we wanted this to be incorporated into the existing `validate-config` command. Might be a question for the team, both ways make some sense.

.codeclimate.yml
```diff
@@ -4,6 +4,8 @@ engines:
     enabled: true
   bundler-audit:
     enabled: true
+  hippo:
+    enabled: true
```
pbrisbin: accidental?

brynary:
Seems like this should be part of the `codeclimate validate-config` experience. Pat -- What's your thinking on why a separate subcommand may make sense?

lib/cc/cli/engines/validate.rb
```diff
@@ -0,0 +1,36 @@
+require "cc/analyzer"
+
+module CC
+  module CLI
+    module Engines
+      class Validate < EngineCommand
+        def run
+          if invalid_engines.any?
+            warn
+            true
```
pbrisbin: We would want this command to exit non-zero to signify failure in this case right? I'm not sure the mechanism commands have for controlling that, but if it's returning `true` here, that seems backwards.

pbrisbin:
@brynary I agree with you and made that comment [myself](https://github.com/codeclimate/codeclimate/pull/75#discussion_r35993171)

pbrisbin:
@brynary scratch that -- I forgot I said both may make some sense earlier... The sense I see in this way is:

- It still is part of `validate-config` experience, `engines:validate` is a bonus
- I wouldn't normally say it was worth it except this implementation is quite small and actually easier this way than the other (because of our current inheritance-based architecture)

Up to me: I'd refactor the inheritance in the `Engines` commands to use composition for the interaction with the config that needs to now be shared. Then I'd use that shared object to implement this only in the `validate-config` command.

ABaldwinHunter:
@brynary @pbrisbin that sounds pretty good.

As an FYI or reminder for consideration- I think the reasons we went with inheritance originally was that at least half of the current collection of engines subcommands (`remove`, `enable`, `disable`) reference  only one engine. (engines:enable <x>, engines:remove <x>). These subcommands are [initialized] (https://github.com/codeclimate/codeclimate/blob/master/lib/cc/cli/engines/engine_command.rb#L11) with an `@engine_name` variable that reduces argument passing. 

That structure might not make sense as our need for `EngineRegistry` and engines stuff in other places expands, but wanted to raise that thought.

lib/cc/cli/engines/validate.rb
brynary: I don't think we have that exit code behavior currently, so if so we should do that separately. Probably want a CommandResult object so we don't litter exit calls in the classes

brynary:
Ok. Yeah I thinkI agree with Pat's proposal

ABaldwinHunter:
@brynary @pbrisbin okay that implementation sounds good to me. Thanks for the discussion.

ABaldwinHunter:
See https://github.com/codeclimate/codeclimate/pull/78/files for composition implementation.

PR #76, opened 2015-07-31
Read exclude paths from .gitignore if present

pbrisbin:
Ended up with free time in a pairing interview and decided to work on this.

### DIFF
--- a/lib/cc/analyzer/engines_runner.rb
+++ b/lib/cc/analyzer/engines_runner.rb
@@ -85,7 +85,15 @@ module CC
       end
 
       def exclude_paths
-        PathPatterns.new(@config.exclude_paths || []).expanded
+        PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
+      end
+
+      def gitignore_paths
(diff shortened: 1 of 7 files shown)


lib/cc/analyzer/engines_runner.rb
```diff
@@ -85,7 +85,15 @@ def registry_lookup(engine_name)
       end
 
       def exclude_paths
-        PathPatterns.new(@config.exclude_paths || []).expanded
+        PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
+      end
+
+      def gitignore_paths
+        if File.exist?(".gitignore")
+          `git ls-files --others -i --exclude-from .gitignore`.split("\n")
```
GordonDiggs: Do you know if `ls-files` has an argument to delimit these with `\0`? Other git calls have that

GordonDiggs: Also, does this work with symlinks?

GordonDiggs: Update: this works with symlinks

PR #77, opened 2015-07-31
Change into the filesystem dir before running engines

GordonDiggs:
/cc @codeclimate/review 

Will fix the exclude paths


pbrisbin:
LGTM, understood that this is really hard to test so we're deferring, but we should try to come back to that.

PR #78, opened 2015-08-03
composition

ABaldwinHunter:
@codeclimate/review This PR:

1) adds warning for invalid engines to the `validate-config` command
2) makes `engine_exists?` and `engine_registry_list` accessible via the `Engines` module through composition.

### DIFF
--- a/lib/cc/cli/engines/install.rb
+++ b/lib/cc/cli/engines/install.rb
@@ -1,5 +1,3 @@
-require "cc/analyzer"
-
 module CC
   module CLI
     module Engines
@@ -17,7 +15,7 @@ module CC
 
         def pull_docker_images
           engine_names.each do |name|
-            if engine_exists?(name)
+            if engine_registry.exists?(name)
               image = engine_image(name)
               pull_engine_image(image)
             else
@@ -30,12 +28,8 @@ module CC
           @engine_names ||= parsed_yaml.engine_names
         end
 
-        def engine_exists?(engine_name)
-          engines_registry_list.keys.include?(engine_name)
-        end
-
         def engine_image(engine_name)
-          engines_registry_list[engine_name]["image"]
+          engine_registry_list[engine_name]["image"]
         end
 
         def pull_engine_image(engine_image)
--- a/lib/cc/cli/validate_config.rb
+++ b/lib/cc/cli/validate_config.rb
@@ -6,6 +6,8 @@ module CC
       include CC::Analyzer
       include CC::Yaml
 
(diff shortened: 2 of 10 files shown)


ABaldwinHunter:
P.S. First approach using `engines:validate` subcommand and discussion found [here] (https://github.com/codeclimate/codeclimate/pull/78/files).

lib/cc/cli/engines.rb
```diff
@@ -9,6 +9,12 @@ module Engines
       autoload :Install, "cc/cli/engines/install"
       autoload :List, "cc/cli/engines/list"
       autoload :Remove, "cc/cli/engines/remove"
+
+      REGISTRY_LIST ||= CC::Analyzer::EngineRegistry.new.list
```
GordonDiggs: Running code when defining a constant can cause some weird things (related to the way that classes are loaded). Can you move this to a method? Alternatively, what do you think of being able to call `CC::Analyzer::EngineRegistry.exists?(engine_name)`?

ABaldwinHunter: I like the suggestion of using `CC::Analyzer::EngineRegistry.exists?(engine_name)`.

ABaldwinHunter:
@codeclimate/review Ready for re-review. 

Refactored approach to use `EngineRegistry` class to check whether an engine exists.
Adds `engine_registry` to `Command` class.

lib/cc/cli/engines/install.rb
```diff
@@ -3,6 +3,8 @@
 module CC
   module CLI
     module Engines
+      include CC::Analyzer
```
GordonDiggs: This `include` should probably be nested under `Install` so that this doesn't override the defaults

ABaldwinHunter: Ah nice catch. `CC::Analyzer` is already included in the `EngineCommand` class so `Install` has access. Not sure why I added `Analyzer` here.

lib/cc/cli/validate_config.rb
```diff
@@ -48,28 +50,47 @@ def errors
...
       end
 
       def display_warnings
         warnings.each do |warning|
-          puts colorize("WARNING: " + warning, :red)
+          puts colorize("WARNING: #{warning}", :red)
         end
       end
+
+      def validate_engines
```
GordonDiggs: What do you think of naming this `display_invalid_engines` to match the convention elsewhere?

GordonDiggs:
One minor thing - LGTM otherwise

lib/cc/cli/validate_config.rb
ABaldwinHunter: Okay, yeah I thought about that too. 

Sounds good.

PR #87, opened 2015-08-13
Improve Engine spec

pbrisbin:
- Replace OpenStruct with mocha stub
- Load anything in spec/support
- Nested describe to avoid namespacing

/cc @codeclimate/review

These are just some zero-impact improvements to the spec that I find useful in
another branch I'm working on (specifically the namespacing), and I'd like to
merge it first here to keep that eventual diff easier to review.


GordonDiggs:
LGTM

PR #88, opened 2015-08-14
Support container logging within Engine

pbrisbin:
We want to log the actions of starting, finishing, or timing out containers
into the database via Kafka/Wally. We will initially expose this information in
Admin and eventually to users via some kind of Builds dashboard.

Unfortunately, these actions occur very far away from the code that is actually
able to perform the logging:

    CC::Builder  |  CC::Analyzer
    =============|============================
    Runner      -|-> EnginesRunner -> Engine
      |                                 |
      V                                 V
    Kafka                           docker run

The approach implemented here passes a ContainerLogger object into
EnginesRunner and onto each Engine. This object responds to event methods which
the Engine invokes as it starts and stops containers.

Within this code base, the object is a NullContainerLogger and does nothing for
these events, but Builder will be able to pass a real ContainerLogger that
writes the information to the database through Kafka.

Other approaches considered and ruled out or deferred:

- Reifying a Container object to hold the docker run and (optionally) event
  logging logic

  This class could have been reusable for the (eventually container-based)
  Clone logic in Builder and taken advantage of the homogeneity that provides.
  Unfortunately, we would still be invoking Container from within Engine as
  it's very useful to share the EngineRunner and Engine objects between Builder
  and Analyzer, so it wouldn't help us in our immediate goal.

  When we do have the cloning performed in a container, creating such a class
  to hold the Posix::Spawn and timeout logic would make sense.

- Customizable life-cycle methods on Engine

  This would bloat Engine with new concerns and not be significantly more
  convenient for the objects that wanted to do logging.

  The object in charge of the actions to take only knows about EnginesRunner,
  so these actions would need to be passed through to Engine somehow.

  This would mean treating functions as values (setting and calling lambdas for
  the container life cycle events), and that is not idiomatic Ruby.

/cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -16,7 +16,9 @@ module CC
         @label = label.to_s
       end
 
-      def run(stdout_io, stderr_io = StringIO.new)
+      def run(stdout_io:, stderr_io: StringIO.new, container_log: NullContainerLog.new)
+        container_log.started(@metadata["image"])
+
         timed_out = false
         pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command)
         Analyzer.statsd.increment("cli.engines.started")
@@ -55,8 +57,13 @@ module CC
           Analyzer.statsd.increment("cli.engines.result.error.timeout")
           Analyzer.statsd.increment("cli.engines.names.#{name}.result.error")
           Analyzer.statsd.increment("cli.engines.names.#{name}.result.error.timeout")
+          container_log.timed_out
           raise EngineTimeout, "engine #{name} ran past #{TIMEOUT} seconds and was killed"
-        elsif status.success?
+        end
+
+        container_log.finished(status, stderr_io.string)
+
+        if status.success?
           Analyzer.statsd.increment("cli.engines.names.#{name}.result.success")
           Analyzer.statsd.increment("cli.engines.result.success")
         else
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -12,6 +12,7 @@ module CC
     autoload :IssueSorter,        "cc/analyzer/issue_sorter"
     autoload :LocationDescription,"cc/analyzer/location_description"
     autoload :NullConfig,         "cc/analyzer/null_config"
+    autoload :NullContainerLog,   "cc/analyzer/null_container_log"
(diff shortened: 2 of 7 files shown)


pbrisbin:
Yes CC, `Engine` is a craaazy class, but it's not my fault and I don't plan on fixing it here :)

lib/cc/analyzer/engine.rb
```diff
@@ -16,7 +16,9 @@ def initialize(name, metadata, code_path, config, label)
         @label = label.to_s
       end
 
-      def run(stdout_io, stderr_io = StringIO.new)
+      def run(stdout_io, stderr_io = StringIO.new, container_log = NullContainerLog.new)
+        container_log.started(@metadata["image"])
```
noahd1: Thoughts on creating an initializer which takes an image name? It just seems strange that only one of the log methods takes an image but they all relate to the same image.

lib/cc/analyzer/engine.rb
```diff
@@ -16,7 +16,9 @@ def initialize(name, metadata, code_path, config, label)
         @label = label.to_s
       end
 
-      def run(stdout_io, stderr_io = StringIO.new)
+      def run(stdout_io, stderr_io = StringIO.new, container_log = NullContainerLog.new)
```
GordonDiggs: What do you think of making these named params?

```ruby
def run(stdout_io:, stderr_io: StringIO.new, container_log: NullContainerlog.new
```

Then you could call it below with just

```ruby
run(stdout_io: @formatter, container_log: container_log)
```

and not have to worry about the stderr value

lib/cc/analyzer/engine.rb
pbrisbin: The object that would initialize this log instance wouldn't know the image name. Since we've got the `EnginesRunner` class in the middle, the original invoking doesn't know anything engine-specific unless it's passed into the life-cycle methods themselves (like this).

lib/cc/analyzer/engine.rb
pbrisbin: I like it.

pbrisbin:
@GordonDiggs suggestion implemented. @noahd1 let me know if me reasoning is satisfactory on your comment.

Ready for re-review.

GordonDiggs:
LGTM

PR #89, opened 2015-08-14
WIP: Sketch for improving Engine

pbrisbin:
/cc @codeclimate/review

This is as sketch. It's built on top of pb-container-log. If we think this is a
good direction, I'll wait for that to merge and open a new PR from this branch
after rebasing, adding/fixing tests etc.

---

New responsibility breakdown:

**Container**: launch a docker container, give its stdout to the caller,
capture its stderr, monitor its process, handle timeouts, call life-cycle
events on the container-log given.

**Engine::ContainerLog**: perform the statsd calls on container life-cycle
events and then forward those events to the log its been given; sort of a
ContainerLog middleware stack idea.

**Engine**: launch and run a Container for the engine, specify how to shuttle
null-delimited stdout to the formatter, compose its container-log with the one
given.

Benefits:

- Engine is smaller, more focused
- Container is reusable for Clone in builder
- Container is easier to reason about (i.e. the complex timeout logic is
  clearer IMO)

Notable change:

Rather than spawning `docker kill` I'm using Ruby's `Process.kill`, the two
should be equivalent and this was more convenient with the way the code is now
(e.g. no need to track container-name).

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -1,0 +1,70 @@
+module CC
+  module Analyzer
+    class Container
+      TIMEOUT = 15 * 60 # 15m
+
+      def initialize(image, command = nil, log = nil)
+        @image = image
+        @command = command
+        @log = log || NullContainerLog.new
+
+        @output_delimeter = "\n"
+        @on_output = ->(*)
+
+        @timed_out = false
+        @stderr_io = StringIO.new
+      end
+
+      def on_output(delimeter = "\n", &block)
+        @output_delimeter = delimeter
+        @on_output = block
+      end
+
+      def run(options)
+        @log.started(@image)
+
+        pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command)
+
+        t_out = read_stdout(out)
+        t_err = read_stderr(err)
+        t_timeout = timeout_thread(pid)
+
+        _, status = Process.waitpid2(pid)
+
+        t_timeout.kill
+      ensure
+        t_timeout.kill if t_timeout
+
+        if @timed_out
+          t_out.kill if t_out
+          t_err.kill if t_err
+        else
+          t_out.join if t_out
+          t_err.join if t_err
+        end
+      end
+
+      private
+
+      def read_stdout(out)
+        Thread.new do
+          out.each_line(@output_delimeter, &@on_output)
+        end
+      end
+
+      def read_stderr(err)
+        Thread.new do
+          err.each_line { |line| stderr_io.write(line) }
+        end
+      end
+
+      def timeout_thread(pid)
+        Thread.new do
+          sleep TIMEOUT
+          Process.kill("KILL", pid)
+          @timed_out = true
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/cc/analyzer/container.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      def on_output(delimeter = "\n", &block)
+        @output_delimeter = delimeter
+        @on_output = block
+      end
+
+      def run(options)
+        @log.started(@image)
+
+        pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command)
```
pbrisbin: I forgot to implement `docker_run_command` but it'll be `docker run` + default options + `options`

lib/cc/analyzer/container.rb
```diff
@@ -0,0 +1,70 @@
...
+    class Container
+      TIMEOUT = 15 * 60 # 15m
+
+      def initialize(image, command = nil, log = nil)
+        @image = image
+        @command = command
+        @log = log || NullContainerLog.new
+
+        @output_delimeter = "\n"
+        @on_output = ->(*)
```
pbrisbin: Should be `->(*) { }` basically a no-op default.

lib/cc/analyzer/container.rb
```diff
@@ -0,0 +1,70 @@
...
+      def run(options)
+        @log.started(@image)
+
+        pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command)
+
+        t_out = read_stdout(out)
+        t_err = read_stderr(err)
+        t_timeout = timeout_thread(pid)
+
+        _, status = Process.waitpid2(pid)
```
pbrisbin: Forgot to call `@log.finished(status, @stderr_io.string)`

lib/cc/analyzer/engine.rb
```diff
@@ -16,83 +17,28 @@ def initialize(name, metadata, code_path, config, label)
...
-        container_log.started(@metadata["image"])
-
-        timed_out = false
-        pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command)
-        Analyzer.statsd.increment("cli.engines.started")
+      def run(stdout_io, container_log = NullContainerLog.new)
+        container = Container.new(
+          @metadata["image"],
+          @metadata["command"],
+          ContainerLogLog.new(name, container_log)
```
GordonDiggs: `ContainerLogLog`?

lib/cc/analyzer/engine.rb
```diff
@@ -113,21 +61,48 @@ def config_file
...
-      end
-
       def output_filter
         @output_filter ||= EngineOutputFilter.new(@config)
       end
 
-      CommandFailure = Class.new(StandardError)
-      EngineFailure = Class.new(StandardError)
-      EngineTimeout = Class.new(StandardError)
+      class ContainerLog
```
pbrisbin: FYI: I'll move this to its own file in the real PR.

lib/cc/analyzer/engine.rb
pbrisbin: Heh, would clearly catch stuff like this when I do the real PR with tests.

pbrisbin:
FYI: because of the rebase on the base branch, this is now showing some differences that aren't really intended. Sorry.

pbrisbin:
Spoke with Gordon IRL -- we think this is promising enough to move forward. I'll open a new PR with tests and things shortly.

PR #90, opened 2015-08-14
Extract Container class from Engine

pbrisbin:
New responsibility breakdown:

**Container**: launch a docker container, give its stdout to the caller,
capture its stderr, monitor its process, handle timeouts, call life-cycle
events on the container-log given.

**Engine::ContainerLog**: perform the statsd calls on container life-cycle
events and then forward those events to the log its been given; sort of a
ContainerLog middleware stack idea.

**Engine**: launch and run a Container for the engine, specify how to shuttle
null-delimited stdout to the formatter, compose its container-log with the one
given.

Benefits:

- Engine is smaller
- Engine#run is smaller
- Container is singly-responsible and reusable for Clone in builder
- Easier to reason about (complex timeout logic is clearer IMO)
- Better tested

Notable change:

Rather than spawning `docker kill` I'm using Ruby's `Process.kill`, the two
should be equivalent and this was more convenient with the way the code is now
(no need to track container-name).

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -1,0 +1,98 @@
+require "posix/spawn"
+
+module CC
+  module Analyzer
+    class Container
+      DEFAULT_TIMEOUT = 15 * 60 # 15m
+
+      def initialize(
+        image:,
+        name:,
+        command: nil,
+        log: NullContainerLog.new,
+        timeout: DEFAULT_TIMEOUT
+      )
+        @image = image
+        @name = name
+        @command = command
+        @log = log
+        @timeout = timeout
+
+        @output_delimeter = "\n"
+        @on_output = ->(*) { }
+
+        @timed_out = false
+        @stderr_io = StringIO.new
+      end
+
+      def on_output(delimeter = "\n", &block)
+        @output_delimeter = delimeter
+        @on_output = block
+      end
+
+      def run(options = [])
+        @log.started(@image, @name)
+
+        pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command(options))
+
+        t_out = read_stdout(out)
+        t_err = read_stderr(err)
+        t_timeout = timeout_thread(pid)
+
+        _, status = Process.waitpid2(pid)
+
+        @log.finished(status, @stderr_io.string)
+
+        t_timeout.kill
+      ensure
+        t_timeout.kill if t_timeout
+
+        if @timed_out
+          @log.timed_out(@timeout)
+          t_out.kill if t_out
+          t_err.kill if t_err
+        else
+          t_out.join if t_out
+          t_err.join if t_err
+        end
+      end
+
+      private
+
+      def docker_run_command(options)
+        [
+          "docker", "run",
+          "--rm",
+          "--name", @name,
+          options,
+          @image,
+          @command,
+        ].flatten.compact
+      end
+
+      def read_stdout(out)
+        Thread.new do
+          out.each_line(@output_delimeter) do |chunk|
+            output = chunk.chomp(@output_delimeter)
+
+            @on_output.call(output)
+          end
+        end
+      end
+
+      def read_stderr(err)
+        Thread.new do
+          err.each_line { |line| @stderr_io.write(line) }
+        end
+      end
+
+      def timeout_thread(pid)
+        Thread.new do
+          sleep @timeout
+          @timed_out = true
+          Process.kill("KILL", pid)
+        end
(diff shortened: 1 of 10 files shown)


lib/cc/analyzer/container.rb
```diff
@@ -0,0 +1,98 @@
...
+      DEFAULT_TIMEOUT = 15 * 60 # 15m
+
+      def initialize(
+        image:,
+        name:,
+        command: nil,
+        log: NullContainerLog.new,
+        timeout: DEFAULT_TIMEOUT
+      )
+        @image = image
```
GordonDiggs: Is there a reason not to use `attr_reader` for these?

lib/cc/analyzer/engine/container_log.rb
```diff
@@ -0,0 +1,48 @@
+module CC
+  module Analyzer
+    class Engine
+      EngineFailure = Class.new(StandardError)
+      EngineTimeout = Class.new(StandardError)
+
+      class ContainerLog
+        def initialize(name, inner_log)
+          @name = name
+          @inner_log = inner_log
```
GordonDiggs: same `attr_reader` question here

lib/cc/analyzer/container.rb
pbrisbin: I don't have a strong preference between making (private) `attr_reader`s or using instance variables directly. I think I advocated the latter at one point but because I see the former more prevalent our code bases I've been starting to do that -- which is what I've done here.

GordonDiggs: >  I see the former more prevalent our code bases I've been starting to do that -- which is what I've done here.

I don't see `attr_reader` in this class

pbrisbin: Sorry, got my former and latters mixed up. Editing my original comment to hopefully clarify...

GordonDiggs: Cool. I haven't seen much of using instance variables directly (probably because I prefer the `attr_reader` approach).

Functionally, they're the same so I don't see a big win either way.

lib/cc/analyzer/container.rb
```diff
@@ -0,0 +1,98 @@
+require "posix/spawn"
+
+module CC
+  module Analyzer
+    class Container
+      DEFAULT_TIMEOUT = 15 * 60 # 15m
+
+      def initialize(
+        image:,
```
ABaldwinHunter: @pbrisbin this is probably a ruby syntax I'm not familiar with, but just wanted to make sure the absence of a value here was intentional.

GordonDiggs:
LGTM. 2 questions:

1. Do any of the new issues on CC need to be addressed?
2. Does this work locally?

lib/cc/analyzer/container.rb
pbrisbin: Yup, makes it a required keyword argument.

pbrisbin:
> Do any of the new issues on CC need to be addressed?

The issues on CC are because the `#run` method which I moved from `Engine` to `Container`, while reduced and simplified greatly (IMO), is still flagged -- (hence 5 fixed in Engine and 5 new in Container). IMO the class/method is improved and there's not much obvious work I can do to go further such that CC is also happy so I'm good shipping it as is. WDYT?

> Does this work locally?

Will definitely test locally before shipping.

lib/cc/analyzer/container.rb
ABaldwinHunter: Cool. thanks for explanation!

GordonDiggs:
I'm cool with that. Should we disable the `Space inside empty braces detected.` rule?

lib/cc/analyzer/container.rb
pbrisbin: > I haven't seen much of using instance variables directly

I'd have to do some kind of audit to be sure, but at least in this PR `Engine` also accesses its instance variables directly.

I was going to switch these over, but it's actually kind of annoying and error prone so I'm going to leave them as is (if that's OK with you). If you want to push for the `attr_reader` approach can you open a PR on the style guide?

pbrisbin:
I'm not sure what that'd do. I do want to catch anyone writing `{ }` for a Hash or `[ ]` for an Array, but a I don't want to write `{}` for a lambda...

pbrisbin:
FYI: `codeclimate analyze` works locally

lib/cc/analyzer/container.rb
GordonDiggs: Will do!

PR #91, opened 2015-08-14
Pass container image and name to all logged events

pbrisbin:
More than just the start event needs to know about container-specific
information (so far the container image and name). One solution to this would
be to instantiate a new log per container and pass in those values at
construction. This does not work for the following reasons:

- The object instantiating the log does not know these values
- The log also needs to know *build*-specific values (e.g. host and snapshot),
  so we really want to instantiate a single log for the build with these values
  passed in at construction (and the constructing object does know these)

The solution chosen is to pass in the container-specific values for every
life-cycle event, not just started.

/cc @codeclimate/review and h/t @noahd1 for spotting this originally

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -41,14 +41,14 @@ module CC
 
         _, status = Process.waitpid2(pid)
 
-        @log.finished(status, @stderr_io.string)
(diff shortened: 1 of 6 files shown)


GordonDiggs:
LGTM

PR #93, opened 2015-08-15
Add support for new docker version output

ablyler:
- Docker recently changed the output of the `docker version` command

### DIFF
--- a/bin/check
+++ b/bin/check
@@ -4,7 +4,7 @@ command -v docker > /dev/null 2>&1 || {
   exit 1
 }
 
-docker version | grep -Fq "Server version" || {
+docker version | grep -q "Server version\|Server:" || {
   echo "Unable to run \`docker version', the docker daemon may not be running" >&2
 
   if command -v boot2docker > /dev/null 2>&1; then


bin/check
```diff
@@ -4,7 +4,7 @@ command -v docker > /dev/null 2>&1 || {
   exit 1
 }
 
-docker version | grep -Fq "Server version" || {
+docker version | egrep -q "(Server version|Server:)" || {
```
pbrisbin: Since this check was never meant to be super strict anyway, what do you think about `grep -Fq "Server"`?

If you'd like to keep the or-pattern, I would just escape the pipe since it's only the one character (and the parens are redundant): `grep -q "Server version\|Server:"`.

If you'd like to keep the extended option, please switch the deprecated `egrep` out for `grep -E`.

pbrisbin:
Thanks so much for addressing this. I had a minor comment, but once address we should be able to get this out today.

ablyler:
@pbrisbin good call, updated

GordonDiggs:
Thank you @ablyler!

PR #94, opened 2015-08-17
Refactor ContainerLog objects

pbrisbin:
- Rename container "log" to container "listener"
- Make the Null listener usable as a base class
- Except a value object for event methods
- Send duration with finished event
- Send stderr with timed_out event
- Separate distinct listener classes
- Compose with a Composite listener
- Move all logging/statsd calls into listeners

/cc @codeclimate/review @noahd1

Confirmed local analysis still works, and even tested by setting a real logger
and seeing the start/finished events log during analysis.


.rubocop.yml
```diff
@@ -22,6 +22,9 @@ Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*'
 
+Style/GuardClause:
+  Enabled: false
```
GordonDiggs: :clap:

lib/cc/analyzer/container_listener.rb
```diff
@@ -0,0 +1,22 @@
+module CC
+  module Analyzer
+    class ContainerListener
+      ContainerData = Struct.new(
```
GordonDiggs: What's the benefit of nesting this under `ContainerListener` instead of `Container`? Nothing in this class actually uses the struct directly. You could even do this `Struct.new` call right within `container_data` above

pbrisbin: Didn't think too hard about it. `Container` seems like a good home too. `Container::ContainerData` is a little redundant, but that's probably OK.

I'd prefer not to do `Struct.new(...).new(...)` or use an `OpenStruct` at the call site.

pbrisbin:
@GordonDiggs moved. @noahd1 thoughts?

GordonDiggs:
LGTM

PR #95, opened 2015-08-17
Add phpmd engine

ablyler:
@brynary, I'm not sure what needs to be done to publish the docker image for codeclimate-phpmd.  But once that is done I think this will work.


noahd1:
We pushed the image to Docker Hub.

PR #97, opened 2015-08-19
Update URL for Code Climate's Docker Hub Page

eliotsykes:
Updated to https://hub.docker.com/u/codeclimate/ (previously: https://registry.hub.docker.com/repos/codeclimate/)


pbrisbin:
Thanks!

PR #99, opened 2015-08-25
Revive pull request #30

jacobi007:


ABaldwinHunter:
@jacobi007 Thanks for your work! Someone from our team will review this PR soon.

brynary:
@jacobi007 I think I fixed the build issue in master. Can you pleasemerge master in here and see if it builds green?

jacobi007:
@brynary it didn't help :/

brynary:
Have another idea. Working on it.

brynary:
OK, can you please try one more time?

brynary:
OK, tried another thing. Another merge please? :)

lib/cc/cli/analyze.rb
```diff
@@ -43,7 +43,7 @@ def registry
       end
 
       def formatter
-        @formatter ||= Formatters::PlainTextFormatter.new
+        @formatter ||= Formatters::PlainTextFormatter.new(filesystem)
```
brynary: We need to adjust the usage of `Formatters.resolve` so it also passes a `filesystem` param

lib/cc/analyzer/formatters/formatter.rb
```diff
@@ -2,7 +2,8 @@ module CC
   module Analyzer
     module Formatters
       class Formatter
-        def initialize(output = $stdout)
+        def initialize(filesystem, output = $stdout)
```
brynary: The `initialize` method in the `JSONFormatter` will need adjustment to accept `filesystem` (currently takes no args)

brynary:
Build fixed. There's a couple (pre-existing) bugs that need to be resolved, however.

brynary:
(Added line comments about said bugs)

jacobi007:
Noted, thanks!

jacobi007:
I've modified the JSONFormatter and it takes a filesystem argument now. On the other hand I'm wondering if that's the best solution because the variable remains unused.

brynary:
@jacobi007 looks good to me.

We'd eventually need the `filesystem` object if we wanted to make the JSON data more closely match the plain text output, but I think it's ok like this for now.

PR #102, opened 2015-08-26
Generate default engine config on init

grstk:


brynary:
@grstk some feedback:

* I think this infers a bit too much. The default config for an engine will not always be `".#{engine_name}.yml"
* I'd prefer not to embed the default configs into `engines.yml`
* Some engines may have more than one config file we want to generate

Given all this, I propose that we create a directory structure which is like: `config/#{engine_name}/...` and any files we put in there are the default configs. For example `config/rubocop/.rubocop.yml` and `config/eshint/.eshint.json` (or whatever their config file is called, I forgot).

That resolves all of the issues, and I think would be cleaner. WDYT?

grstk:
@brynary sounds reasonable. I'll be back with the improvements soon.

brynary:
Also if you merge master, the build should pass

brynary:
LGTM

PR #103, opened 2015-08-27
Add default eslint config

grstk:
Based on https://github.com/codeclimate/eslint/blob/master/conf/eslint.json. 
Equivalent to `extends: 'eslint:recommended'`.


brynary:
Would it work for us to use a JSON-formatted `.eslintrc`?

grstk:
Yes, eslint accepts both formats. I used YAML only because I find it cleaner and easier to read than JSON.

brynary:
OK, sounds good

PR #106, opened 2015-09-09
Enable Duplication Engine by default if there are ruby files

jacobi007:


pbrisbin:
I'd assume we don't want to merge something like this until the very end, once we've done some testing of the new engine(s)?

brynary:
@jacobi007 I think in addition to enabling the engine in this case, we should also configure the `languages` to be "Ruby". Thoughts on how we could do that?

jacobi007:
@brynary Check out my solution.

lib/cc/cli/init.rb
```diff
@@ -21,7 +21,8 @@ def create_codeclimate_yaml
         config = { "engines" => {} }
         eligible_engines.each do |engine_name, engine_config|
           config["engines"][engine_name] = {
-            "enabled" => true
+            "enabled" => true,
+            "config" => (additional_duplication_options if engine_name == 'duplication'),
```
brynary: I don't love hardcoding specific engine names here. Rather than doing that, what if we add:

```yaml
duplication:
  # ...
  default_config:
    languages:
      - ruby
```

To `engines.yml` and use that?

That would work for now, and we can revisit when we run into the issue of: "How do we make the `languages` variable, based on what we find when we init?"

lib/cc/cli/init.rb
```diff
@@ -80,6 +81,19 @@ def eligible_engines
           end
         end
       end
+
+      def additional_duplication_options
+        { "languages" => ["ruby"] }
```
brynary: BTW, should this be "ruby" or "Ruby"? Do both work?

jacobi007:
Should be ok now, @brynary

jacobi007:
Hmmm some tests failed, let me get back to you once I fix it

ABaldwinHunter:
@jacobi007 @codeclimate/review is this change related to the duplication work that @BlakeWilliams is working on or waiting on something else?

ABaldwinHunter:
like arbitrary configs permitted in codeclimate-yaml, [which we now have](https://github.com/codeclimate/codeclimate-yaml/pull/23) :)

jacobi007:
Thanks for the info @ABaldwinHunter . I've reverted the last commit as it was referencing code from my old PR to yaml-repo which is no longer relevent in the light of [mxie's pull request](https://github.com/codeclimate/codeclimate-yaml/pull/23) . I think there is nothing else stopping us here.

jacobi007:
Thanks for the info @ABaldwinHunter . I've reverted the last commit as it was referencing code from my old PR to yaml-repo which is no longer relevent in the light of [mxie's pull request](https://github.com/codeclimate/codeclimate-yaml/pull/23) . I think there is nothing else stopping us here.

BlakeWilliams:
I forgot to mention it, but is this the same thing? https://github.com/codeclimate/codeclimate/pull/130

ABaldwinHunter:
@BlakeWilliams @jacobi007 Hm, yeah those look pretty similar to me.

BlakeWilliams:
Yeah, we should close this PR unless there's any objections.

PR #107, opened 2015-09-09
Emit include_paths for use by engines.

fhwang:
The path reduction goes top-down recursively, plucking unreadable files etc. FYI it takes less than a second on the Discourse repo.


lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,140 @@
+module CC
+  module Analyzer
+    class IncludePathsBuilder
+      def initialize(cc_excludes)
+        @cc_excludes = cc_excludes
```
GordonDiggs: I think naming this `exclude_paths` might be clearer. WDYT?

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+
+      def included_files
+        includable_files - ignored_files
+      end
+
+      def includable_files
+        tracked_in_git = `git ls-files -z`.split("\0")
+        untracked_in_git = `git ls-files -zo`.split("\0")
+        tracked_in_git + untracked_in_git
```
GordonDiggs: Is it possible for things to show up in both? Would a union be better here?

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,140 @@
...
+            result
+          else
+            []
+          end
+        end
+
+        protected
+
+        def ensure_hashified(obj)
+          if obj.is_a?(Array)
```
GordonDiggs: What in the above calls could make `obj` be of different types?

lib/cc/analyzer/include_paths_builder.rb
fhwang: Maybe `cc_exclude_paths`? I wanted to differentiate it from patterns in `.gitignore`, which aren't ever technically called `exclude_paths`, but are usually a list of paths to exclude.

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,140 @@
...
+            obj
+          end
+        end
+
+        def files_all_included?
+          file_entries.all? { |e| @included_files[e] }
+        end
+
+        def file_entries
+          unless @_file_entries
```
GordonDiggs: This method could be 

```ruby
@_file_entries ||= relevant_full_entries.reject do |e|
  File.directory?(e)
end
```

lib/cc/analyzer/include_paths_builder.rb
fhwang: My understanding is that the two sets are always mutually exclusive. `git ls-files` only shows tracked files and `git ls-files -o` only shows untracked files.

spec/cc/analyzer/engines_runner_spec.rb
```diff
@@ -76,6 +81,28 @@ module CC::Analyzer
       EnginesRunner.new(registry, formatter, "/code", config).run
     end
 
+    context "when the source directory contains all readable files, and there are no ignored files" do
```
GordonDiggs: We use `describe` everywhere else. What does `context` give us?

lib/cc/analyzer/include_paths_builder.rb
GordonDiggs: Cool - makes sense

lib/cc/analyzer/include_paths_builder.rb
GordonDiggs: I like `cc_exclude_paths`

lib/cc/analyzer/include_paths_builder.rb
fhwang: Basically, there's a necessary step of turning an Array into a Hash as an optimization. In practice it's only run once, at the top directory, and then the hash is passed down recursively.

I opted to do this inside `Directory`, because it felt like the responsibility of `Directory` to know how it's going to use that list and thus optimize it.

lib/cc/analyzer/include_paths_builder.rb
fhwang: Yeah, I just didn't feel like finding out if our CodeClimate setup would complain about that too ;) So we consider these sorts of or-equals assignments okay?

I also sometimes do `||= begin ... end`, which some people hate but I think make sense if you're into the functionalish side of Ruby. Also, it makes a lot of editors (including my Vim setup) lose their mind w.r.t. indenting.

```
@_memoized_var ||= begin
   some_code
   var
end
```

GordonDiggs: We definitely use `||=` generously for memoization. I'm not sure about the `begin...end` pattern. I don't think it makes sense in this method but maybe elswhere

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,140 @@
...
+
+        def included_subdirectory_results
+          result = []
+          subdirectories.each do |subdirectory|
+            result = result.concat(subdirectory.included_paths)
+          end
+          result
+        end
+
+        def readable?
```
pbrisbin: Originally the requirements of this solution were such that, if there were files/directories that were not readable and weren't intended to be excluded, those were still supposed to bubble up as permissions errors.

It was only when the engines encountered unreadable files *that the user had intended to exclude* that is "the permissions  bug" we're trying to fix.

It might be a fine 95% solution, or the best we can do given all the subtle hurdles we've encountered, but if the requirement were boiled down to: "just skip unreadable stuff", there may have been something simpler that didn't involve all the git tom-foolery:

- Recursively expand the source dir
- Rescue permissions errors and ignore
- Filter excluded
- Present the result as `include_paths`

Right?

Also, we have to keep in mind that the CLI (this code) runs as root (technically whatever user is UID:0), while the engines run as an underprivileged user (app:9000), so anything you do permissions-related in this code isn't guaranteed to translate to what the engine is ultimately able to access.

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,140 @@
...
+          if @_readable.nil?
+            begin
+              Dir.entries(@path)
+              @_readable = true
+            rescue Errno::EACCES, Errno::EPERM
+              @_readable = false
+            end
+          end
+          @_readable
```
pbrisbin: :lipstick: we don't use this pattern of underscored "private" (I assume?) instance variables

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,140 @@
...
+        end
+
+        def relevant_full_entries
+          unless @_relevant_full_entries
+            raw_entries = Dir.entries(@path).reject do |e|
+              %w(. .. .git).include?(e)
+            end
+            @_relevant_full_entries = raw_entries.map do |e|
+              @path == "." ? e : File.join(@path, e)
```
pbrisbin: :lipstick: no ternary please

pbrisbin:
I'm going to hold off on additional review, in case there's any large changes to come out my [inline comment](https://github.com/codeclimate/codeclimate/pull/107/files#r39071075).

lib/cc/analyzer/include_paths_builder.rb
brynary: Is this the same as `Dir['**/*']` or something similar? Why do we need `git` at all?

lib/cc/analyzer/include_paths_builder.rb
fhwang: Is it house style at CC not to use ternaries? Personally I'm fine with them as long as they're not too complex.

fhwang:
OK, this is now ready for review again. The logic regarding unreadable files and directories is:

* If a directory can't be read by everybody, we exclude it from the whitelist we pass to the engines. This mimics the way that Git silently ignores unreadable directories.
* If a file can't be read by everybody, we check for its presence in `.gitignore` or `.codeclimate.yml`
  * If it matches a pattern in one of those, we exclude it from the whitelist we pass to the engines
  * If it doesn't match any pattern in those files, we raise an error. The user needs to fix it locally before continuing.

cc @codeclimate/review

spec/cc/analyzer/engines_runner_spec.rb
```diff
@@ -66,9 +72,11 @@ module CC::Analyzer
       registry = registry_with_engine("rubocop")
       formatter = null_formatter
 
+      FileUtils.stubs(:readable_by_all?).at_least_once.returns(true)
       expected_config = {
         "enabled" => true,
-        :exclude_paths => %w[ .ignorethis ]
+        :exclude_paths => %w[ .ignorethis ],
+        :include_paths => %w[.gitignore]
```
mxie: Use the same spacing for both lines.

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,156 @@
...
+            subdirectories_all_included?
+        end
+
+        def included_paths
+          if all_included?
+            [@path + "/"]
+          elsif readable_by_all?
+            result = []
+            result = result.concat(included_file_entries)
+            result = result.concat(included_subdirectory_results)
```
mxie: Is this style preferred over `+=` or even just `included_file_entries + included_subdirectory_results`?

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,156 @@
...
+
+        protected
+
+        def ensure_hashified(obj)
+          if obj.is_a?(Array)
+            result = {}
+            obj.each do |included|
+              result[included] = true
+            end
+            result
```
mxie: You could replace the above 5 lines with something like this:

```ruby
obj.each_with_object({}) do |included, result|
  result[included]
end
```

and it'll return the resulting hash at the end.

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,156 @@
...
+          else
+            obj
+          end
+        end
+
+        def files_all_included?
+          !file_entries.any? { |e| @excluded_files[e] }
```
mxie: How about using `none?` over `!<whatever>.any?`?

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,156 @@
...
+        def included_file_entries
+          file_entries.reject { |file_entry| @excluded_files[file_entry] }
+        end
+
+        def included_subdirectory_results
+          result = []
+          subdirectories.each do |subdirectory|
+            result = result.concat(subdirectory.included_paths)
+          end
+          result
```
mxie: Similar trick here as https://github.com/codeclimate/codeclimate/pull/107/files#r39193403 using `each_with_object` but with a `[]`.

lib/cc/analyzer.rb
```diff
@@ -33,5 +34,8 @@ def method_missing(*)
     cattr_accessor :statsd, :logger
     self.statsd = DummyStatsd.new
     self.logger = DummyLogger.new
+
+    class UnreadableFileError < StandardError
```
pbrisbin: :lipstick: we use the `UnreadableFileError = Class.new(StandardError)` form

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,156 @@
...
+        @cc_exclude_paths = cc_exclude_paths
+      end
+
+      def build
+        root = Directory.new('.', ignored_files)
+        paths = root.included_paths
+        paths.each do |path|
+          raise_on_unreadable_files(path)
+        end
+        paths
```
pbrisbin: I'm 99% sure `Enurable#each` returns the collection.

lib/cc/analyzer/include_paths_builder.rb
pbrisbin: https://github.com/codeclimate/styleguide#general

The reasoning is that ternaries (regardless of complexity) "hide" conditionals, so expanding them is a way to make the conditional logic of your program more visible and painful so you're motivated to remove it rather than work around it by compressing to a ternary.

lib/file_utils_ext.rb
```diff
@@ -0,0 +1,7 @@
+require 'fileutils'
+
+module FileUtils
+  def self.readable_by_all?(path)
```
pbrisbin: Why did you go for a monkey-patch over just putting a `#readable_by_all?` on the object that needs this logic?

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,156 @@
...
+        tracked_in_git = `git ls-files -z`.split("\0")
+        untracked_in_git = `git ls-files -zo`.split("\0")
+        tracked_in_git + untracked_in_git
+      end
+
+      def raise_on_unreadable_files(path)
+        if File.directory?(path)
+          raise_on_unreadable_files_in_directory(path)
+        elsif !FileUtils.readable_by_all?(path)
+          raise(CC::Analyzer::UnreadableFileError, "Can't read #{path}")
```
pbrisbin: I'm noticing the "exception handling for control flow" smell here. Did you consider any non-exception-based alternatives?

pbrisbin: Oh nevermind, this is actually raising out to the end-user? Sorry, ignore my comment.

pbrisbin: :lipstick: no parens on `raise`

lib/file_utils_ext.rb
fhwang: 1. The object under consideration is a `String`, which seems like a class we don't want to be adding methods to. Unless you're suggesting assigning `readable_by_all?` as a singleton method on any given `String` that could end up being used as a path inside `IncludePathsBuilder::Directory`, but that seems brittle in practice.

2. I could use `Pathname` instead of `String`, which was in Bryan's initial code sample. Generally I found that the switching back and forth between the two caused a lot of bookkeeping overhead. For example, you're dealing with `git ls-files` output, should you convert them all into `Pathname` before moving on?

3. I also tried making this a method on `IncludePathsBuilder`, but hit problems due to testing. In a lot of the tests in this repo, we use `Dir.mktmpdir`, which creates a temporary directory. Only problem is, it sets permissions of that temp directory to be `0700`, which caused erroneous bugs when testing code that's so dependent on file and directory permissions. I wanted to centralize this logic in a method that was stubbable, and I didn't want that method to live on `IncludePathsBuilder` since I disliked the idea of stubbing a method on the class under test.

pbrisbin: > I also tried making this a method on IncludePathsBuilder

This is what I was suggesting. Thanks for the explanation.

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,150 @@
...
+          file_entries.none? { |e| @excluded_files[e] }
+        end
+
+        def file_entries
+          unless @_file_entries
+            @_file_entries = relevant_full_entries.reject do |e|
+              File.directory?(e)
+            end
+          end
+          @_file_entries
```
pbrisbin: How about `@file_entries ||= relevant_full_entries.reject do |e|`?

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,150 @@
...
+            result.concat(subdirectory.included_paths)
+          end
+        end
+
+        def readable_by_all?
+          FileUtils.readable_by_all?(@path)
+        end
+
+        def relevant_full_entries
+          unless @_relevant_full_entries
```
pbrisbin: We don't use this `@_` convention anywhere else. Could you avoid it to keep things consistent?

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -0,0 +1,150 @@
...
+          tmp << @cc_exclude_paths.join("\n")
+          tmp.close
+          tracked_and_ignored = `git ls-files -zi -X #{tmp.path}`.split("\0")
+          untracked_and_ignored = `git ls-files -zio -X #{tmp.path}`.split("\0")
+          tracked_and_ignored + untracked_and_ignored
+        end
+      end
+
+      def raise_on_unreadable_files(path)
+        if File.directory?(path)
```
pbrisbin: At what point are you actually looking to see if directories themselves are "unreadable"?

fhwang: When `Directory#include_paths` calls `Directory#readable_by_all?`.

pbrisbin: Gotcha

pbrisbin:
OK. I'm going to call this LGTM without nitpicking further for the following reasons:

- The specs are comprehensive
- The code is for the most part confined to the new class

This is probably worth some manual QAing as well, which we should do before releasing.

PR #109, opened 2015-09-09
Fix #108

### LINKED ISSUE #108: Support Docker Machines not named "default"
I have updated my Mac to replace boot2docker with Docker Machine. Unfortunately it does not seem to work with the codeclimate wrapper script.

```
$ codeclimate                                                                                                                                                                          Your Docker setup does not support the codeclimate wrapper script:

  > /var/run/docker.sock must exist as a Unix domain socket

We require a local Docker daemon that supports communication via the default
socket path.

Please use `docker run' to run the `codeclimate/codeclimate' image directly.

See https://github.com/codeclimate/codeclimate for more details.
```

I do have the Docker env variables set correctly.

```
$ env | grep DOCKER                                                                                                                                                                  DOCKER_TLS_VERIFY=1
DOCKER_HOST=tcp://192.168.99.100:2376
DOCKER_CERT_PATH=/Users/daniel/.docker/machine/machines/docker-dev
DOCKER_MACHINE_NAME=docker-dev
```

squaresurf:
Why:

* Docker Machine has support for many machines which results in
  codeclimate needing to know which machine to ssh.

This change addresses the need by:

* Use the `$DOCKER_MACHINE_NAME` environment variable that gets set by
* `docker-machine env`.

### DIFF
--- a/codeclimate-wrapper
+++ b/codeclimate-wrapper
@@ -29,10 +29,10 @@ invalid_docker_host() {
 }
 
 if command -v docker-machine > /dev/null 2>&1; then
-  docker-machine ssh default -- \
+  docker-machine ssh $DOCKER_MACHINE_NAME -- \
     test -S /var/run/docker.sock > /dev/null 2>&1 || socket_missing
 
-  docker-machine ssh default -- \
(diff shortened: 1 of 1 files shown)


GordonDiggs:
Thank you!

PR #113, opened 2015-09-16
Let users specify engines and paths as options to analyze command for easier debugging

ABaldwinHunter:
@codeclimate/review 

Caveat: At the moment, specifying the path part only works for `codeclimate-eslint` - which uses include paths instead of exclude paths internally.

### DIFF
--- a/lib/cc/analyzer/engines_runner.rb
+++ b/lib/cc/analyzer/engines_runner.rb
@@ -6,11 +6,12 @@ module CC
       InvalidEngineName = Class.new(StandardError)
       NoEnabledEngines = Class.new(StandardError)
 
-      def initialize(registry, formatter, source_dir, config, container_label = nil)
+      def initialize(registry, formatter, source_dir, config, requested_paths = [], container_label = nil)
         @registry = registry
         @formatter = formatter
         @source_dir = source_dir
         @config = config
+        @requested_paths = requested_paths
         @container_label = container_label
       end
 
@@ -27,6 +28,8 @@ module CC
       end
 
       private
+
+      attr_reader :requested_paths
 
       def run_engine(engine, container_listener)
         @formatter.engine_running(engine) do
@@ -72,12 +75,12 @@ module CC
         @registry[engine_name]
       end
 
-      def exclude_paths
-        PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
+      def include_paths
+        IncludePathsBuilder.new(@config.exclude_paths || [], requested_paths).build
       end
 
-      def include_paths
-        IncludePathsBuilder.new(@config.exclude_paths || []).build
+      def exclude_paths
+        @exclude_paths ||= PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
       end
 
       def gitignore_paths
@@ -87,6 +90,7 @@ module CC
           []
         end
       end
+
     end
   end
 end
--- a/lib/cc/analyzer/engines_runner.rb
+++ b/lib/cc/analyzer/engines_runner.rb
@@ -80,7 +80,7 @@ module CC
       end
 
       def exclude_paths
-        @exclude_paths ||= PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
+        PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
       end
 
       def gitignore_paths
@@ -90,7 +90,6 @@ module CC
           []
         end
       end
-
     end
   end
 end
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -5,15 +5,18 @@ module CC
 
       def initialize(_args = [])
         super
+        @engine_options = []
+        @path_options = []
 
         process_args
+        apply_config_options
       end
 
       def run
         require_codeclimate_yml
 
         Dir.chdir(ENV['FILESYSTEM_DIR']) do
-          runner = EnginesRunner.new(registry, formatter, source_dir, config)
+          runner = EnginesRunner.new(registry, formatter, source_dir, config, path_options)
           runner.run
         end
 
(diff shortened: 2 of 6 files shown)


lib/cc/analyzer/engines_runner.rb
```diff
@@ -72,12 +75,16 @@ def metadata(engine_name)
...
+        if paths && paths.any?
+          paths
+        else
+          IncludePathsBuilder.new(@config.exclude_paths || []).build
+        end
       end
 
-      def include_paths
-        IncludePathsBuilder.new(@config.exclude_paths || []).build
+      def exclude_paths
```
ABaldwinHunter: Put `include_paths` at higher level than `exclude_paths` method because the former uses the latter, but maybe noisy change here.

lib/cc/analyzer/engines_runner.rb
```diff
@@ -72,12 +75,16 @@ def metadata(engine_name)
         @registry[engine_name]
       end
 
-      def exclude_paths
-        PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
+      def include_paths
+        if paths && paths.any?
```
mxie: If `paths` is any array, can we get away with just doing `if paths.present?` instead?

lib/cc/cli/analyze.rb
```diff
@@ -25,13 +28,22 @@ def run
...
+      attr_reader :engine_options, :path_options
+
       def process_args
         while arg = @args.shift
           case arg
           when '-f'
             @formatter = Formatters.resolve(@args.shift)
+          when '-e'
+            @engine_options << @args.shift
+          when '--engine'
```
mxie: Can we combine these cases? i.e. `when '-e', '--engine'`

ABaldwinHunter: Yes that sounds good! thanks!

lib/cc/analyzer/engines_runner.rb
ABaldwinHunter: okay that sounds good

lib/cc/analyzer/engines_runner.rb
```diff
@@ -28,6 +29,9 @@ def run(container_listener = ContainerListener.new)
 
       private
 
+      attr_accessor :paths
+      attr_reader :exclude_paths
```
andyw8: I think the `exclude_paths` method defined on line 87 will override this?

ABaldwinHunter: :+1: good call

ABaldwinHunter:
@codeclimate/review @brynary Ready for re-review.

Updates: 
1) let a user pass an engine not in `.codeclimate.yml` config (fixed cc-yaml node bug)
2) accept multiple file and dir paths as arguments
3) leverage permissions-checking of `IncludePathsBuilder` and honor `exclude_paths` from `.codeclimate.yml` when analyzing requested file paths.

Part three is the trickiest IMO, and any feedback appreciated!

Re codeclimate issues: I think a refactoring might be in order, but outside the scope of this PR. (These changes seem like jenga pieces.)

Example:

```
Ashleys-MacBook-Pro-2:boggle ashleybaldwin-hunter$ codeclimate analyze -e rubocop app/models
Starting analysis
Running rubocop: Done!

== app/models/boggle_board.rb (17 issues) ==
1: Missing top-level class documentation comment. [rubocop]
18: Pass `&:shuffle!` as an argument to `each` instead of a block. [rubocop]
6: Prefer single-quoted strings when you don't need string interpolation or special symbols. [rubocop]
24-25: Extra empty line detected at class body end. [rubocop]
6: Prefer single-quoted strings when you don't need string interpolation or special symbols. [rubocop]
31-36: 5 trailing blank lines detected. [rubocop]
18: Space between { and | missing. [rubocop]
22: Space between { and | missing. [rubocop]
18: Space missing inside }. [rubocop]
22: Surrounding space missing for operator `+`. [rubocop]
22: Surrounding space missing for operator `%`. [rubocop]
22: Surrounding space missing for operator `+`. [rubocop]
22: Prefer single-quoted strings when you don't need string interpolation or special symbols. [rubocop]
22: Space missing inside }. [rubocop]
22: Line is too long. [102/80] [rubocop]
22: Prefer single-quoted strings when you don't need string interpolation or special symbols. [rubocop]
22: Prefer single-quoted strings when you don't need string interpolation or special symbols. [rubocop]

== app/models/score.rb (1 issue) ==
2: Final newline missing. [rubocop]

== app/models/word.rb (1 issue) ==
2: Final newline missing. [rubocop]

Analysis complete! Found 19 issues.
```

ABaldwinHunter:
Fixing a bug.

lib/cc/analyzer/engines_runner.rb
```diff
@@ -6,11 +6,12 @@ class EnginesRunner
       InvalidEngineName = Class.new(StandardError)
       NoEnabledEngines = Class.new(StandardError)
 
-      def initialize(registry, formatter, source_dir, config, container_label = nil)
+      def initialize(registry, formatter, source_dir, config, requested_paths = [], container_label = nil)
```
fhwang: Not a blocker, but this argument is getting long enough that keyword arguments are starting to look like a good option.

https://robots.thoughtbot.com/ruby-2-keyword-arguments

fhwang:
Code-wise this LGTM. However:

When we start adding all these subtle types of includes and excludes some of the edge case start to get tricky. What happens if my `.codeclimate.yml` excludes a path but I explicitly include it in the command line? If I include engines A and B in `.codeclimate.yml`, but I specify engine C in the command line, am I running one engine or three?

The answers to these aren't obvious and should be documented somewhere. I tried running `codeclimate help analyze` but I got no love. Is there an easy way to put this in the CLI help? Or failing that can we make sure to update `docs.codeclimate.com` when this is released?

wfleming:
Improving the CLI help and doing per-command detailed help is definitely something we need to do as we start to build out more features like this (though it means we'll need a new interview pairing problem :wink:).

I do think that functionality should be in the CLI, but since it's a wider-reaching thing that will impact all commands, I also think we can do it separately.

fhwang:
Makes sense. Personally I'd be fine with this PR as-is if a related card about CLI documentation made its way into the backlog.

brynary:
I'm good with this.

PR #120, opened 2015-09-21
Update check message to use sudo when appropriate

pbrisbin:
If bin/check is run as root (i.e. via sudo make install), and the user is
unable to connect to docker (as root), the error message is confusing since it
instructs them to ensure docker version succeeds (which it likely does). In
fact, *sudo* docker version is the failing command that must succeed before
sudo make install would.

To address this confusion, we include sudo in the example command if the check
script is being run as root (UID=0).

Note: we're purposely ignoring the edge case where a user is running the check
script as root but not via sudo (e.g. su root) and our example command
incorrectly instructs them to try sudo docker version.


brynary:
Lgtm

PR #122, opened 2015-09-23
Add executables to gemspec.

fhwang:
These are needed in the gemspec by Bundler for binstubs.


GordonDiggs:
LGTM

codeclimate.gemspec
```diff
@@ -14,6 +14,7 @@ Gem::Specification.new do |s|
   s.files         = Dir["lib/**/*.rb"] + Dir["bin/*"] + Dir["config/*"]
   s.require_paths = ["lib"]
   s.bindir        = "bin"
+  s.executables   = %w[check codeclimate release]
```
wfleming: I don't think `release` should be part of the built gem.

PR #125, opened 2015-09-25
Fix formatter instantiation

### LINKED ISSUE #111: Undefined method `source_buffer_for'  for "/code":String
```
jordan$ codeclimate analyze -f text
Starting analysis
Running eslint: Done!

== Brocfile.js (7 issues) ==
error: (NoMethodError) undefined method `source_buffer_for' for "/code":String
This is especially confusing because there is no method (or even text string) source_buffer_for anywhere in my project, which leads me to believe this error is coming from codeclimate.
```

/cc @elwayman02

pbrisbin:
Formatters are meant to be instantiated with an instance of FileSystem,
however when handling the -f option we were instantiating it with the raw
directory. To fix this, Formatters.resolve now returns the class itself and
the caller instantiates it with the FileSystem instance it holds already.

Fixes #111

/cc @codeclimate/review

Note: my initial attempts at getting a meaningful failing spec around this
before implementing the fix were not successful, but I should probably figure
something out along those lines before merging.

### DIFF
--- a/spec/cc/cli/analyze_spec.rb
+++ b/spec/cc/cli/analyze_spec.rb
@@ -17,89 +17,98 @@ module CC::CLI
             stderr.must_match("No enabled engines. Add some to your .codeclimate.yml file!")
           end
         end
+      end
 
-        describe "when engine is not in registry" do
-          it "ignores engine, without blowing up" do
-            within_temp_dir do
-              create_yaml(<<-EOYAML)
-                engines:
-                  madeup:
-                    enabled: true
-                  rubocop:
-                    enabled: true
-              EOYAML
+      describe "when engine is not in registry" do
+        it "ignores engine, without blowing up" do
+          within_temp_dir do
+            create_yaml(<<-EOYAML)
+              engines:
+                madeup:
+                  enabled: true
+                rubocop:
+                  enabled: true
+            EOYAML
 
-              _, stderr = capture_io do
-                Analyze.new.run
-              end
-
-              stderr.must_match("")
+            _, stderr = capture_io do
+              Analyze.new.run
             end
+
+            stderr.must_match("")
           end
         end
+      end
 
-        describe "when user passes engine options to command" do
-          it "uses only the engines provided" do
-            within_temp_dir do
-              create_yaml(<<-EOYAML)
-                engines:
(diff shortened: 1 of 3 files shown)


wfleming:
:+1:

pbrisbin:
I'm having a really tough time coming up with a valuable test here. If anyone has any ideas, please let know. Otherwise, I'm going to merge without it shortly.

mxie:
@pbrisbin You could potentially do a `stub_const` on `FORMATTERS` to make sure that we're only ever passing in something that's in that array (and raise otherwise).

pbrisbin:
Thank's @mxie , it's actually not the `resolve` method that needs testing (IMO). The original bug was in how we were instantiating the formatter (and only presented when it ultimately gets used a number of objects away), so what I'm having trouble with is finding a spec that would expose that bug...

mxie:
Ah, I see...so it's on the `Analyze` class instead then? Could you spy on the `resolve` method?

fhwang:
Did you consider putting an expectation on `JSONFormatter#new` to ensure it gets the filesystem argument it's expecting when a `-f` flag is passed?

pbrisbin:
@fhwang i had originally discounted that as brittle and not meeting the "valuable" bar; it's probably better than nothing though, so I'll explore that route a bit further.

pbrisbin:
Cool, spec added. Note: I also fixed some invalid block nesting in that spec as part of adding my new example at the end.

spec/cc/cli/analyze_spec.rb
```diff
@@ -17,91 +17,100 @@ module CC::CLI
...
+            CC::Analyzer::EnginesRunner.expects(:new).with(anything, anything, anything, anything, paths).returns(engines_runner)
 
-              analyze.run
-            end
+            analyze.run
           end
         end
       end
+
+      describe "when a formatter argument is passed" do
```
pbrisbin: This is the only actual change in this file.

fhwang:
Yeah I'd imagine a more full-scale spec would be something on the integration level, as this bug seems to be in part due to a misunderstanding as to how the classes are supposed to interact.

Still, on its own I think this LGTM

PR #127, opened 2015-09-29
Skip .gitignore if it's a directory.

fhwang:
IncludePathsBuilder was erroring on repos where .gitignore was a
directory (most likely due to developer error). Now it just skips
.gitignore instead.


GordonDiggs:
LGTM

PR #130, opened 2015-09-30
Add `default_config` key with nested values

BlakeWilliams:
This adds the `default_config` option which is used to add default
configuration options to an engine. This also marks the duplication
engine as a default engine when ruby files are present.

### DIFF
--- a/lib/cc/cli/config.rb
+++ b/lib/cc/cli/config.rb
@@ -1,0 +1,32 @@
+module CC
+  module CLI
+    class Config
+      delegate :to_yaml, to: :config
+      def initialize
+        @config = {
+          "engines" => {},
+          "ratings" => { "paths" => [] }
+        }
+      end
+
+      def add_engine(engine_name, engine_config)
+        config["engines"][engine_name] = { "enabled" => true }
+
+        if engine_config["default_config"].present?
+          config["engines"][engine_name]["config"] = engine_config["default_config"]
+        end
+
+        config["ratings"]["paths"] |= engine_config["default_ratings_paths"]
+      end
+
+      def add_exclude_paths(paths)
+        config["exclude_paths"] ||= []
+        config["exclude_paths"] += paths.map { |path| "#{path}/**/*" }
+      end
+
+      private
+
+      attr_reader :config
+    end
+  end
+end
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -1,3 +1,5 @@
+require 'cc/cli/config'
+
 module CC
   module CLI
     class Init < Command
@@ -18,19 +20,13 @@ module CC
       private
 
       def create_codeclimate_yaml
-        config = { "engines" => {} }
-        eligible_engines.each do |engine_name, engine_config|
-          config["engines"][engine_name] = {
-            "enabled" => true
-          }
-          config["ratings"] ||= {}
-          config["ratings"]["paths"] ||= []
+        config = CC::CLI::Config.new
 
-          config["ratings"]["paths"] |= engine_config["default_ratings_paths"]
+        eligible_engines.each do |(engine_name, engine_config)|
+          config.add_engine(engine_name, engine_config)
         end
 
-        config["exclude_paths"] = exclude_paths(AUTO_EXCLUDE_PATHS)
-
+        config.add_exclude_paths(auto_exclude_paths)
         filesystem.write_path(CODECLIMATE_YAML, config.to_yaml)
       end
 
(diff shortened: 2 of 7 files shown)


mxie:
Tagging @codeclimate/review for a review. (Can someone also just add Blake on this team?)

lib/cc/cli/init.rb
```diff
@@ -18,14 +18,13 @@ def run
...
-          }
-          config["ratings"] ||= {}
-          config["ratings"]["paths"] ||= []
+        config = {
+          "engines" => {},
+          "ratings" => { "paths" => [] },
+        }
 
+        eligible_engines.each do |engine_name, engine_config|
+          config["engines"][engine_name] = build_engine_config(engine_config)
```
mxie: What do you think about doing something like...:

```ruby
def create_codeclimate_yaml
  # ...
  config["engines"][engine_name]["config"] = build_engine_config(engine_config)
  # ...
end
```

...and then just have `build_engine_config` return the appropriate config hash?

mxie: Oh wait...nevermind.

BlakeWilliams: It also needs to set `"enabled" => true` on the engine so I don't think this will work unless we extract that as well.

lib/cc/cli/init.rb
```diff
@@ -65,6 +64,16 @@ def exclude_paths(paths)
...
       end
 
+      def build_engine_config(engine_config)
+        config = { "enabled" => true }
+
+        if engine_config["default_config"]
+          config["config"] = engine_config["default_config"]
+        end
+
+        config
```
mxie: It feels like there should be a cleaner way to do this...but I can't think of anything. Shouldn't be a blocker though.

mxie:
LGTM, but maybe get a set of eyes from a CC dev just in case.

lib/cc/cli/init.rb
wfleming: I think maybe what makes it feel unclean is naming: we're passing around an `engine_config` which is really the "default" or "prototype" engine config, and referring to the actual current engine config as just `config`. (I think: that's my reading of the code.) Agree it's not a blocker, but it does seem it could be made clearer.

BlakeWilliams: I agree. I just used what it was called in the calling method. I think it would be nice to refactor the entire file.

wfleming:
LGTM. I agree with @mxie that some of it feels like it could be a bit cleaner, so if you agree I'd ask you to take a couple minutes to think about if it could be improved, but I don't consider it blocking.

BlakeWilliams:
@mxie @wfleming I cleaned it up a bit. One more look?

lib/cc/cli/config.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      def add_engine(engine_name, engine_config)
+        config["engines"][engine_name] = { "enabled" => true }
+
+        if engine_config["default_config"].present?
+          config["engines"][engine_name]["config"] = engine_config["default_config"]
+        end
+
+        config["ratings"]["paths"] |= engine_config["default_ratings_paths"]
```
BlakeWilliams: Hah, just put two and two together on this one. Union equals.

lib/cc/cli/init.rb
```diff
@@ -55,28 +51,27 @@ def available_configs
...
-        expanded_paths = []
-        paths.each do |dir|
-          if filesystem.exist?(dir)
-            expanded_paths << "#{dir}/**/*"
-          end
+      def engine_eligible?(engine)
+        !engine["community"] && engine["enable_regexps"].present? && has_files(engine)
+      end
+
+      def has_files(engine)
```
mxie: Maybe name this with `?` since it's returning a boolean.

lib/cc/cli/config.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      def add_exclude_paths(paths)
+        config["exclude_paths"] ||= []
+        config["exclude_paths"] += paths.map { |path| "#{path}/**/*" }
+      end
+
+      attr_reader :config
```
mxie: I think `attr_reader`s generally belong near the top of the class definition.

lib/cc/cli/config.rb
```diff
@@ -0,0 +1,33 @@
...
+    class Config
+      def initialize
+        @config = {
+          "engines" => {},
+          "ratings" => { "paths" => [] }
+        }
+      end
+
+      def to_yaml
+        config.to_yaml
```
mxie: Alternatively: `delegate :to_yaml, to: :config`, but *shrug*.

mxie:
Just a few nitpicks, but this refactor looks awesome. Great job!

lib/cc/cli/config.rb
BlakeWilliams: I accidentally deleted `private` above it. :stuck_out_tongue:

lib/cc/cli/init.rb
BlakeWilliams: Good catch!

PR #131, opened 2015-10-01
EnginesRunner doesn't assume formatter is closable

fhwang:


pbrisbin:
Discussed IRL that `IOProcessor` (builder's formatter, the object raising because it lacks a `#close`) should've probably  subclassed `Formatter`. That would've given it an empty `#close` that would've also prevented this bug -- we came to the conclusion that over-paranoia is warranted though, and even if we did go ahead and do that, having the `#responds_to?` check here is valuable over-protection.

So... LGTM

PR #132, opened 2015-10-03
add confidence reporting

ABaldwinHunter:
@codeclimate/review @mrb 

Experiment with having Code Climate CLI accept a confidence key on issue.

Only applicable type of analysis that I'm aware of is Brakeman, so it may not be clean design idea to make a custom key for that quality yet. 

screen shot 2015-10-02 at 8 18 47 pm

### DIFF
--- a/lib/cc/analyzer/formatters/plain_text_formatter.rb
+++ b/lib/cc/analyzer/formatters/plain_text_formatter.rb
@@ -7,6 +7,17 @@ module CC
     module Formatters
       class PlainTextFormatter < Formatter
 
+        SEVERITY_LEVELS = [
+          INFO = "info",
+          NORMAL = "normal",
+          CRITICAL = "critical",
+        ].freeze
+
+        SEVERITY_HUES = {
+          INFO => "#1a0dab",
+          CRITICAL => "#FF4500",
+        }.freeze
+
         def started
           puts colorize("Starting analysis", :green)
         end
@@ -40,7 +51,13 @@ module CC
               end
 
               print(issue["description"])
+
+              if issue["severity"] && issue["severity"] != NORMAL
+                report(issue["severity"])
+              end
+
               print(colorize(" [#{issue["engine_name"]}]", "#333333"))
+
               puts
             end
             puts
@@ -67,6 +84,14 @@ module CC
         end
 
         private
+
+        def report(severity)
+          color = SEVERITY_HUES[severity] || "#333333"
+          str = colorize(" [", "#333333")
+          str += colorize("#{severity}", color)
+          str += colorize("]", "#333333")
+          print str
+        end
 
         def spinner(text = nil)
           @spinner ||= Spinner.new(text)
--- a/spec/cc/analyzer/formatters/plain_text_formatter_spec.rb
+++ b/spec/cc/analyzer/formatters/plain_text_formatter_spec.rb
@@ -36,6 +36,39 @@ module CC::Analyzer::Formatters
         stdout.must_match("Missing top-level class documentation comment")
         stdout.must_match("[cool_engine]")
       end
+
+      it "reports issue severity when info" do
+        engine = stub(name: "cool_engine")
+
+        stdout, _ = capture_io do
+          write_from_engine(formatter, engine, sample_issue_with_severity("info"))
+          formatter.finished
+        end
+
+        stdout.must_match("[info]")
+      end
+
+      it "reports issue severity when critical" do
+        engine = stub(name: "cool_engine")
+
+        stdout, _ = capture_io do
+          write_from_engine(formatter, engine, sample_issue_with_severity("critical"))
+          formatter.finished
+        end
+
+        stdout.must_match("[critical]")
+      end
+
+      it "does not report severity when normal" do
+        engine = stub(name: "cool_engine")
+
+        stdout, _ = capture_io do
+          write_from_engine(formatter, engine, sample_issue_with_severity("normal"))
+          formatter.finished
+        end
+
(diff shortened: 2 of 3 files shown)


lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -68,6 +74,19 @@ def failed(output)
 
         private
 
+        def confidence_report(confidence)
+          colors_map = {
```
mxie: Can we make this into a constant instead? This seems unlikely to change.

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -68,6 +74,19 @@ def failed(output)
...
+            "weak" => "#905E3B",
+            "medium" => "#1a0dab",
+            "high" => "#FF4500",
+          }
+
+          color = colors_map[confidence] || "#333333"
+
+          print " with confidence: "
+          print(colorize("#{confidence}", color))
```
mxie: Could be combine these 2 lines to make it into a single `print` statement?

lib/cc/analyzer/formatters/plain_text_formatter.rb
ABaldwinHunter: :+1:

ABaldwinHunter:
@codeclimate/review @mrb @brynary 

Updated to use `severity` instead of `confidence` in plaintextformatter.

JSON formatter already works as is.

Plaintext formatter sample output:

screen shot 2015-10-07 at 3 59 09 pm

brynary:
I would omit normal and only print info and severity. We an put it in brackets next to the engine name. Critic in red, info in another color.



Typed with thumbs.

ABaldwinHunter:
@brynary :+1:  

For a clarification: 

1) I think we decided that order of severity atm is
`info < normal < critical`, translated from `weak < medium < high` in confidence terms.

I'm assuming we want to print the two most severe, so did you want a name swap so that `medium => info` and `weak => normal`, or am I reading too much into your comment :) ?

2) When the issue severity is weakest, should we omit that issue from CLI output entirely, or only the severity description?

mrb:
@ABaldwinHunter - Use `info < normal < critical` but only display the level when it is `normal` or `critical`. Display the issue in all cases.

brynary:
Actually, I'd maybe display the issue when it is `[info]` or `[critical]` and leave it off for normal

ABaldwinHunter:
:+1:

ABaldwinHunter:
@codeclimate/review cc/ @brynary @mrb Ready for re-review. Formatting updated.

I've currently stuck with this style:
screen shot 2015-10-12 at 12 26 23 pm

More concise alternative:
screen shot 2015-10-12 at 12 25 53 pm

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -40,7 +51,13 @@ def finished
               end
 
               print(issue["description"])
+
+              if issue["severity"] && issue["severity"] != NORMAL
```
ABaldwinHunter: This setup defaults to inclusively reporting the severity if it does not fit one of our expected levels [info normal critical].

Alternative: exclude severity unless `critical` or `info`

ABaldwinHunter:
@brynary @mrb updated post design feedback:

screen shot 2015-10-12 at 5 31 43 pm

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -68,6 +85,14 @@ def failed(output)
 
         private
 
+        def report(severity)
+          color = SEVERITY_HUES[severity] || "#333333"
+          str = colorize(" [", "#333333")
+          str += colorize("#{severity}", color)
```
GordonDiggs: Using `<<` instead of `+=` here will be a tiny bit faster, as it will not make a new String object every time.

```
[1] pry(main)> require 'benchmark/ips'
=> true
[2] pry(main)> 
[3] pry(main)> str1 = ""
=> ""
[4] pry(main)> str2 = ""
=> ""
[5] pry(main)> 
[6] pry(main)> Benchmark.ips do |x|
[6] pry(main)*   x.report("+=") { str1 += "a" }  
[6] pry(main)*   x.report("<<") { str2 << "a" }  
[6] pry(main)*   x.compare!  
[6] pry(main)* end  
Calculating -------------------------------------
                  +=     5.054k i/100ms
                  <<   104.102k i/100ms
-------------------------------------------------
                  +=     16.325k (±52.8%) i/s -     70.756k in   5.550502s
                  <<      3.788M (± 5.4%) i/s -     18.947M

Comparison:
                  <<:  3788143.5 i/s
                  +=:    16325.3 i/s - 232.04x slower
```

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -7,6 +7,17 @@ module Analyzer
     module Formatters
       class PlainTextFormatter < Formatter
 
+        SEVERITY_LEVELS = [
+          INFO = "info",
```
GordonDiggs: Does our spec allow arbitrary values for severity or do we need to validate these values?

lib/cc/analyzer/formatters/plain_text_formatter.rb
pbrisbin: `232.04x slower` ha!

ABaldwinHunter:
Going to close this branch for now to reduce noise.

PR #134, opened 2015-10-06
Test checking formatter for failure before sending finished event

dblandin:
This PR prevents the running engine from sending off a second container `finished` event after the formatter has already handled an engine output failure.

Depends upon the `#failed` formatter method added in codeclimate/builder#168 to block correctly.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -18,13 +18,15 @@ module CC
         name:,
         command: nil,
         listener: ContainerListener.new,
-        timeout: DEFAULT_TIMEOUT
+        timeout: DEFAULT_TIMEOUT,
+        stdout_io: StringIO.new
       )
         @image = image
         @name = name
         @command = command
         @listener = listener
         @timeout = timeout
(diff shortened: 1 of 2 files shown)


pbrisbin:
Can we spend some time pairing together on this and the builder PR? Coupling the formatter and the listeners together would be really bad, so I want to be absolutely sure we have to do it before signing off. I think I just need to spend time actually digging with you to get the needed context to feel comfortably one way or another. Sorry to be a blocker on this work...

dblandin:
@pbrisbin No worries! That sounds good to me. After we talked yesterday, I dug into the issue a bit more and realized that we did in fact have a race condition between the exception raise and the `ContainerListener#finished` event firing. I was able to remove the event that I added in the `IOProcessor`, replace it with a `sleep 5`, and see the same "second" finished event come through on its own. So it seems that by adding that `finished` event in Builder, I've given codeclimate/codeclimate just enough time for it to send off its non-errored `finished` event.

Let me know how you would like to move forward :+1:

dblandin:
Ended up taking a different approach. Closing this PR.

PR #135, opened 2015-10-07
update gemfile - unless we intentionally switched back to 0.6.3

ABaldwinHunter:
@codeclimate/review 

Not sure what's best protocol for handling bundler differences. 

I think we want `0.6.4` for CC in gemfile.lock though. Looks like the last update might not have used a bundle.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    codeclimate (0.6.3)
+    codeclimate (0.6.4)
       activesupport (~> 4.2, >= 4.2.1)
       codeclimate-yaml (~> 0.2.3)
       faraday (~> 0.9.1)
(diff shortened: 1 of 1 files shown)


pbrisbin:
I think this happens if someone skips the `bundle` step when deploying. The version in the lockfile should always match the version in the gemspec, AFAIK.

Gemfile.lock
```diff
@@ -49,7 +49,7 @@ GEM
       metaclass (~> 0.0.1)
     multipart-post (2.0.0)
     posix-spawn (0.3.11)
-    pry (0.10.2)
+    pry (0.10.1)
```
BlakeWilliams: Wanna bump this back up to `0.10.2`?

ABaldwinHunter: :+1:

BlakeWilliams:
LGTM

PR #136, opened 2015-10-08
enable bundler-audit when Gemfile.lock present

wfleming:
Looks like this was using the wrong yaml key: I don't see any reference
to `enable_patterns` in the code.

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -104,8 +104,8 @@ bundler-audit:
   community: false
   upgrade_languages:
     - Ruby
-  enable_patterns:
-    - Gemfile.lock
+  enable_regexps:
+    - Gemfile\.lock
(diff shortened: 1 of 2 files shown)


pbrisbin:
LGTM, is there a good way to confirm this works? I assume there's no good way to get it into the test suite.

wfleming:
@pbrisbin I'll alter the tests for it, but since I can't foresee that merging at all cleanly with #137, I'm going to wait until that's done & then I'll rebase this & update it. Make sense?

pbrisbin:
Fine with me, although I don't see a merge conflict on the horizon.

wfleming:
@pbrisbin spec added.

pbrisbin:
Awesome. LGTM

PR #137, opened 2015-10-08
`init --upgrade`

wfleming:
This adds an `--upgrade` flag the `init` command will recognize to upgrade classic configuration.

In the upgrade case, the logic for each individual engine should be:

* Enable it if its language was supported by the classic platform & was enabled in the user's existing classic config.
* Enable it if its language was not supported by classic platform, but it meets the normal criteria for inclusion.
* Do not enable it if its language was supported by the classic platform but it was not enabled in the user's classic config.

Additionally, if upgrading a classic config, do not alter `exclude_paths`, since this key existed for classic config & should be respected.

Basically, do what we would normally do unless it contradicts existing config.

There's definitely still opportunity for me clean this up some, but I'm going to open this for review now since I think the cleaning will make the diff harder to read. cc @codeclimate/review

### DIFF
--- a/lib/cc/cli/config_generator.rb
+++ b/lib/cc/cli/config_generator.rb
@@ -1,0 +1,60 @@
+module CC
+  module CLI
+    class ConfigGenerator
+      CODECLIMATE_YAML = Command::CODECLIMATE_YAML
+      AUTO_EXCLUDE_PATHS = %w(config/ db/ features/ node_modules/ script/ spec/ test/ vendor/).freeze
+
+      def self.for(filesystem, engine_registry, upgrade_requested)
+        if upgrade_requested && upgrade_needed?(filesystem)
+          UpgradeConfigGenerator.new(filesystem, engine_registry)
+        else
+          ConfigGenerator.new(filesystem, engine_registry)
+        end
+      end
+
+      def initialize(filesystem, engine_registry)
+        @filesystem = filesystem
+        @engine_registry = engine_registry
+      end
+
+      def eligible_engines
+        return @eligible_engines if @eligible_engines
+
+        engines = engine_registry.list
+        @eligible_engines = engines.each_with_object({}) do |(name, config), result|
+          if engine_eligible?(config)
+            result[name] = config
+          end
+        end
+      end
+
+      def exclude_paths
+        AUTO_EXCLUDE_PATHS.select { |path| filesystem.exist?(path) }
+      end
+
+      def post_generation_verb
+        "generated"
+      end
+
+      private
+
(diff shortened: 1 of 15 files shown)


spec/cc/cli/init_spec.rb
```diff
@@ -94,12 +82,12 @@ module CC::CLI
...
       end
 
-      describe "when a .codeclimate.yml file is already present in working directory" do
+      describe "when a platform .codeclimate.yml file is already present in working directory" do
         it "does not create a new file or overwrite the old" do
           within_temp_dir do
             filesystem.exist?(".codeclimate.yml").must_equal(false)
 
-            yaml_content_before = "This is a test yaml!"
+            yaml_content_before = "---\languages:\n  Ruby: true\n"
```
pbrisbin: Is this supposed to be `---\nlanguages...`?

fhwang:
`init.rb` has a lot of calls to `upgrade?` At the risk of inciting OO wankery, I wonder if it would clean things up to use a little delegation or even inheritance.

spec/cc/cli/init_spec.rb
wfleming: …Yes. I thought I had caught all the instances. Thanks.

wfleming:
@fhwang Yep, I'm in progress on that now.

wfleming:
Ok, this is feeling better. Another look @fhwang @codeclimate/review ?

lib/cc/cli/config_generator.rb
```diff
@@ -0,0 +1,61 @@
...
+        AUTO_EXCLUDE_PATHS.select { |path| filesystem.exist?(path) }
+      end
+
+      def post_generation_verb
+        "generated"
+      end
+
+      private
+
+      attr_reader :filesystem
```
GordonDiggs: :lipstick: combine these 2 lines

lib/cc/cli/init.rb
```diff
@@ -51,29 +57,17 @@ def available_configs
         all_paths.reject { |path| ['.', '..'].include?(File.basename(path)) }
       end
 
-      def engine_eligible?(engine)
-        !engine["community"] && engine["enable_regexps"].present? && has_files?(engine)
-      end
+      def engines_enabled?
+        return @engines_enabled if @engines_enabled
```
GordonDiggs: The one issue with this memoization is that if `@engines_enabled` is false, it won't use the memoized value. I think you can get around that by returning `unless @engines_enabled.nil?`

wfleming: Yep, good point.

GordonDiggs:
LGTM

PR #138, opened 2015-10-09
Skip required yaml on engines flag

mrb:
Kind of dirty I admit


lib/cc/analyzer/formatters/formatter.rb
```diff
@@ -7,7 +7,7 @@ def initialize(filesystem, output = $stdout)
           @output = output
         end
 
-        def write(data)
+        def write(_)
```
pbrisbin: Nice, can you `_data` so we don't lose the information about what it would be? Same with `_output`.

lib/cc/cli/command.rb
```diff
@@ -5,6 +5,7 @@
 
 module CC
   module CLI
+
```
pbrisbin: :lipstick: extra whitespace

lib/cc/cli/analyze.rb
```diff
@@ -61,7 +63,13 @@ def source_dir
       end
 
       def config
-        @config ||= CC::Yaml.parse(filesystem.read_path(CODECLIMATE_YAML))
+        if !@engine_options.empty?
```
pbrisbin: :lipstick: `if !` => `unless`, then `unless` / `else` => flip the conditional

```rb
if @engines_options.any?
  # ...
else
  # ...
end
```

pbrisbin:
Can you describe a little bit (preferably ammending it into the commit too) in what cases we can/should run an `#analyze` without the cc-yaml present? That is what this change is allowing for right?

PR #139, opened 2015-10-09
Add requiresafe

mrb:
Yay!

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -155,3 +155,9 @@ scss-lint:
   enable_regexps:
   default_ratings_paths:
     - "**.scss"
+requiresafe:
+  image: codeclimate/codeclimate-requiresafe
+  description: Security tool for Node.js dependencies
+  community: true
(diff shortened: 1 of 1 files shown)


GordonDiggs:
:+1:

config/engines.yml
```diff
@@ -155,3 +155,9 @@ scss-lint:
   enable_regexps:
   default_ratings_paths:
     - "**.scss"
+requiresafe:
+  image: codeclimate/codeclimate-requiresafe
+  description: Security tool for Node.js dependencies
+  community: true
+  enable_regexps:
+  default_ratings_paths:
```
pbrisbin: Should these be filled in with something?

mrb: @pbrisbin This tool only looks at one file, and doesn't impact grades yet, so I think we're fine for now.

pbrisbin: I'd expect that one file in the `enable_regexps` then, but OK.

PR #140, opened 2015-10-09
Add codeclimate-init.

fhwang:
To eliminate collisions between `bin/codeclimate` here, and the `codeclimate` binstub we distribute with out Docker image, we:

* Create a new `bin/codeclimate-init` binary, which can be used in Platform and is included as an executable in the gemspec
* Remove `bin/codeclimate` from the gemspec. It's still in the repo and directly findable, just not installed anywhere by `gem install` or `bundle install`.


pbrisbin:
LGTM

PR #141, opened 2015-10-12
Container raises BlankImage exception if passed a blank string.

fhwang:
One more step in trying to make our stack easier to develop and debug.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -10,7 +10,7 @@ module CC
         :status,        # status, for a finished event
         :stderr,        # stderr, for a finished event
       )
-      BlankImage = Class.new(StandardError)
+      ImageRequired = Class.new(StandardError)
 
       DEFAULT_TIMEOUT = 15 * 60 # 15m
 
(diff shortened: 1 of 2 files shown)


pbrisbin:
"Blank" feels a little implementation-ey -- what is the source of this error?

fhwang:
I was trying to run a builder image on the command line. I forgot to set BUILDER_CLONE_IMAGE in my call to `bin/codeclimate-builder`, which ended up eventually calling `CC::Analyzer::Container` with an image that was an empty string (because it's trying to get them from the env). Figuring out the issue took me maybe fifteen minutes of writing layers of debugging code.

There are probably lots of other places this issue could've been prevented and/or made visible, but at the very least trying to run `docker run` with no image name seems incorrect. I'm calling it "blank" in this case to go along with the definition of blankness in ActiveSupport.

pbrisbin:
Cool. I was wondering if it was because of a bad lookup into engine metadata.

Given that it's more generic, how about `NoImageGiven` or `ImageNameRequired`?

I think tying the exception name directly to the fact that you're testing it with `#blank?` is exactly what I'm against doing (though I'm doing a poor job of articulating why).

fhwang:
Yeah, `ImageNameRequired` is a good one, I'll change to that and fix the CC issues.

pbrisbin:
Cool, LGTM after that.

PR #142, opened 2015-10-12
Add a Container#stop method to exit early and send a finished event

dblandin:
This PR allows the running container to stop analysis immediately in case of an output processing error.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -42,13 +42,13 @@ module CC
         started = Time.now
         @listener.started(container_data)
 
-        pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command(options))
+        @pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command(options))
 
         t_out = read_stdout(out)
         t_err = read_stderr(err)
-        t_timeout = timeout_thread(pid)
+        t_timeout = timeout_thread(@pid)
 
-        _, status = Process.waitpid2(pid)
+        _, status = Process.waitpid2(@pid)
 
         duration = ((Time.now - started) * 1000).round
         @listener.finished(container_data(duration: duration, status: status))
@@ -65,6 +65,15 @@ module CC
           t_out.join if t_out
           t_err.join if t_err
         end
+      end
+
+      def stop
+        # Prevents the processing of more output after first error
+        @on_output = ->(*) { }
+
+        Process.kill("TERM", @pid) if @pid
+      rescue Errno::ESRCH
+        Analyzer.statsd.increment("container.kill_process_rescue")
       end
 
       private
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -33,7 +33,7 @@ module CC
 
         container.on_output("\0") do |output|
           unless output_filter.filter?(output)
-            stdout_io.write(output)
+            stdout_io.write(output) || container.stop
           end
         end
 


lib/cc/analyzer/container.rb
```diff
@@ -67,6 +67,12 @@ def run(options = [])
         end
       end
 
+      def stop
+        @on_output = ->(*) { }
+        Process.kill("TERM", @pid)
```
GordonDiggs: What happens if `@pid` is nil here?

dblandin: Looks like if `@pid` were `nil` we would get a `TypeError: no implicit conversion from nil to integer` but I don't think that's possible. `Container#stop` is only called in response to output emitted from the running engine which would have been assigned a PID. If the process terminates before we send a `TERM` command, we'll get an `Errno::ESRCH` exception, which we're already handling with a no-op rescue.

lib/cc/analyzer/container.rb
```diff
@@ -67,6 +67,13 @@ def run(options = [])
         end
       end
 
+      def stop
+        @on_output = ->(*) { }
```
pbrisbin: Why is this necessary?

lib/cc/analyzer/container.rb
```diff
@@ -67,6 +67,13 @@ def run(options = [])
...
+        Process.kill("TERM", @pid)
+      rescue Errno::ESRCH
+        Analyzer.logger.info("Tried to send TERM to process with PID #{@pid}, but not found.")
```
pbrisbin: Probably `warn`? I'm also OK completely ignoring, maybe with a statsd increment.

lib/cc/analyzer/container.rb
dblandin: It seems like the process (and output) still runs for a bit even after sending `TERM` or `KILL` so this prevents a bunch of extra errored events from reaching Kafka and triggering unprocessable state changes on the snapshot.

lib/cc/analyzer/container.rb
dblandin: I was getting a Code Climate issue for leaving this empty. I think a statsd increment sounds good.

lib/cc/analyzer/container.rb
GordonDiggs: I think this should be resilient to being called prematurely. Would a simple `if @pid` achieve that?

lib/cc/analyzer/engine.rb
```diff
@@ -33,7 +33,7 @@ def run(stdout_io, container_listener)
 
         container.on_output("\0") do |output|
           unless output_filter.filter?(output)
-            stdout_io.write(output)
+            container.stop unless stdout_io.write(output)
```
GordonDiggs: this trailing conditional seems odd to me because `sdtout_io.write` is the primary action here. What do you think of

```ruby
stdout_io.write(output) or container.stop
```

I don't usually advocate using `or` but given that it's for control flow and not logic, I think it's a good use case.

lib/cc/analyzer/container.rb
dblandin: Sure, that works for me :+1:

lib/cc/analyzer/engine.rb
dblandin: Yep, I think that gives more weight to the primary action. :+1:

lib/cc/analyzer/container.rb
```diff
@@ -67,6 +67,13 @@ def run(options = [])
         end
       end
 
+      def stop
+        @on_output = ->(*) { }
+        Process.kill("TERM", @pid)
+      rescue Errno::ESRCH
+        Analyzer.statsd.increment("container.kill_process_rescue")
```
dblandin: @pbrisbin Is this fine as a metric key or would you prefer something else?

GordonDiggs:
LGTM

lib/cc/analyzer/engine.rb
pbrisbin: > I don't usually advocate using or but given that it's for control flow and not logic, I think it's a good use case.

FWIW, I love `and` / `or` for control flow. That's what they're there for; it's why the lower precedence.

```rb
foo_bar or raise "x"

x = get_it and go(x)
```

We seem to avoid them here, so I've been abiding but I'd be happy to have a discussion about bringing them back into our toolbox.

pbrisbin:
LGTM too

PR #144, opened 2015-10-14
Add empty output to engine output filter rules

dblandin:
This prevents processing errors when the engine output is parsed into another structure by adding empty output to the `EngineOutputFilter` filter rules.

### DIFF
--- a/lib/cc/analyzer/engine_output_filter.rb
+++ b/lib/cc/analyzer/engine_output_filter.rb
@@ -8,6 +8,8 @@ module CC
       end
 
       def filter?(output)
+        return true unless output.present?
+
         if (json = parse_as_json(output))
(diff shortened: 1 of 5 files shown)


wfleming:
This looks good. I think you can remove the lines in the formatters made redundant by this change.

There may be some highger-level specs here that might be good to add to as well, if convenient.

fhwang:
Thanks, this is a better place to fix this issue. For example, now other formatters get the benefit of this filtering.

dblandin:
Updated the remove the redundant empty output checks within formatters.

wfleming:
Looks ok to me. Might want to build as a docker image locally and use the CLI to run analysis a couple times if you haven't already.

dblandin:
Seems to be running fine locally :+1:

PR #145, opened 2015-10-14
Add functionality to support snapshot events.

fhwang:


lib/cc/analyzer/container.rb
```diff
@@ -121,6 +124,7 @@ def container_data(duration: nil, status: nil)
           @name,
           duration,
           status,
+          @stdout_io.string,
```
GordonDiggs: Should there be a change to the constructor for `ContainerData`

GordonDiggs:
I vote we disable the long class check - we never fix it and now we require CC to be green to merge this

lib/cc/analyzer/container.rb
fhwang: Yeah, it's on line 11 above :)

lib/cc/analyzer/container.rb
```diff
@@ -94,6 +96,7 @@ def docker_run_command(options)
       def read_stdout(out)
         Thread.new do
           out.each_line(@output_delimeter) do |chunk|
+            @stdout_io.write(chunk)
```
pbrisbin: I think we need to actually avoid always storing stdout for all containers. The Engines' stdout is completely useless and can be _crazy_ big, already causing memory issues across our infrastructure.

My original thought is that we could default `@on_output` to write to a `StringIO`, like you have here (instead of the current just-ignore-it lamba).

Then engines would continue to override it to read their results (and not bloat up a big `@stdout_io` object) but non-engines would have access to that output in the container data passed to the lifecycle methods.

lib/cc/analyzer/engines_runner.rb
```diff
@@ -13,27 +13,36 @@ def initialize(registry, formatter, source_dir, config, requested_paths = [], co
...
+        @container_listeners[engine] = listener
+      end
+
       private
 
       attr_reader :requested_paths
 
-      def run_engine(engine, container_listener)
+      def run_engine(engine)
+        listener = @container_listeners[engine] || ContainerListener.new
```
pbrisbin: I can see how engines-specific listeners may be valuable (or required) soon, but is it needed to do the stdout stuff you have in this PR? Feels orthogonal.

fhwang:
I ended up supporting this differently and with minimal changes to this repo, so I'm closing this.

PR #147, opened 2015-10-15
Extract Analyzer::Engines from Analyzer::EnginesRunner.

fhwang:
This reduces the responsibility of Analyzer::EnginesRunner so it only
runs analyses on an array of pre-instantiated Engines. Choosing and
instantiating engines with the correct config is now the responsibility
of Analyzer::Engines.

This is motivated in part by desired changes in stream handling in
platform -- separating these responsibilities here will give us more
flexibility elsewhere. In addition, the hope is that this might lead to
phasing out EnginesRunner entirely.

### DIFF
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -5,6 +5,7 @@ module CC
     autoload :Container, "cc/analyzer/container"
     autoload :ContainerListener, "cc/analyzer/container_listener"
     autoload :Engine, "cc/analyzer/engine"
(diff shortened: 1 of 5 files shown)


wfleming:
It's nice to see the existing tests pass with this, but `Engines` should probably get a unit spec for itself?

fhwang:
Good point @wfleming . I've moved over the relevant specs to a new `engines_spec.rb`.

wfleming:
:+1:

PR #149, opened 2015-10-16
Only enable bundler-audit if Gemfile.lock in root

jpignata:
The previous regular expression pattern would match if there was a file named Gemfile.lock in any directory of the project but bundler-audit crashes if a Gemfile.lock isn't present in the project's root directory.


GordonDiggs:
:+1:

PR #153, opened 2015-10-20
Container timeout calls only listener#timed_out, not listener#finished.

fhwang:


pbrisbin:
Seems reasonable to me /cc @dblandin though

fhwang:
@dblandin checks are passing, can you take a look?

lib/cc/analyzer/container.rb
```diff
@@ -27,16 +27,14 @@ def initialize(
...
         @on_output = ->(*) { }
-
         @timed_out = false
         @stderr_io = StringIO.new
       end
 
-      def on_output(delimeter = "\n", &block)
-        @output_delimeter = delimeter
+      def on_output(delimiter = "\n", &block)
+        @output_delimiter = delimiter
```
pbrisbin: Nice catch, but what do you think about pulling this to its own PR? This logic is so critical and so futzy, I want to be super careful with any changes to it.

dblandin:
Looks good to me! :+1:

lib/cc/analyzer/container.rb
fhwang: Yeah, that's fair, I'll back it out.

PR #154, opened 2015-10-20
Use docker kill, not Process.kill

pbrisbin:
It's generally believed that sending kill -9 to the docker-run client process in
an attempt to kill the running container is unreliable and incorrect. Really,
how could it work? The client process must exit immediately and can do nothing
to handle such a signal (like, for example, tell the docker server to stop the
container).

That said, this approach *seems* to work reliably in the timeout case. The spec
added here around the timeout behavior, which I believe is an accurate
reproduction of what should not work, works 100% reliably with either version of
the code.

What did not work was Container#stop. When a spec was added around this
behavior, the following was seen:

- The Process.kill("TERM") was ignored, but
- The Process.kill("KILL") occurring on eventual timeout was also ignored
- The container was left running

This should explain the zombie container we see. Container#stop is used on
invalid output, which phpcodesniffer (a common offender) is known to produce.

It's worth noting that replacing TERM with KILL in the #stop method (something
I'd expect to work, given how timing out via this signal works reliably) did not
address the issue -- the same behavior was seen.

Understanding all this, and aided by the two new integration specs (which
actually run and kill real containers), both uses of Process.kill were replaced
with the assumed-better docker-kill approach.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -42,13 +42,13 @@ module CC
         started = Time.now
         @listener.started(container_data)
 
-        @pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command(options))
+        pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command(options))
 
         t_out = read_stdout(out)
         t_err = read_stderr(err)
-        t_timeout = timeout_thread(@pid)
+        t_timeout = timeout_thread
 
-        _, status = Process.waitpid2(@pid)
+        _, status = Process.waitpid2(pid)
         if @timed_out
           @listener.timed_out(container_data(duration: @timeout))
         else
@@ -70,9 +70,7 @@ module CC
         # Prevents the processing of more output after first error
         @on_output = ->(*) { }
 
-        Process.kill("TERM", @pid) if @pid
-      rescue Errno::ESRCH
-        Analyzer.statsd.increment("container.kill_process_rescue")
+        reap_running_container
       end
 
       private
@@ -104,16 +102,21 @@ module CC
         end
       end
 
-      def timeout_thread(pid)
+      def timeout_thread
         Thread.new do
           sleep @timeout
           @timed_out = true
-          Process.kill("KILL", pid)
+          reap_running_container
         end
       end
 
       def container_data(duration: nil, status: nil)
         ContainerData.new(@image, @name, duration, status, @stderr_io.string)
+      end
+
+      def reap_running_container
+        Analyzer.logger.warn("killing container name=#{@name}")
+        POSIX::Spawn::Child.new("docker", "kill", @name)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/cc/analyzer/container.rb
```diff
@@ -124,6 +124,15 @@ def container_data(duration: nil, status: nil)
           @stderr_io.string
         )
       end
+
+      def reap_running_container
+        container_id = `docker ps --quiet --filter name=#{@name}`.chomp
```
brynary: `docker kill` (and all docker commands0 accept a container name as a substitute for the container ID. Therefore, I think the name lookup is unnessary.

pbrisbin: Unfortunately not true for me. Can you test, maybe it was a version
difference?

brynary: Doubtful. The behavior I'm referring to has been in Docker for a long time now. Example:

```
$ docker run -d --name=foo alpine sleep 120
219b77e8f943b41c4ca61461c123af62f2e7533a5ac23fa23734c75c739e67e1

☺ ~ 
 [osx 18:41:36] $ docker kill foo
foo
```

(That's off 1.8.2)

pbrisbin: You're right! That's much better.

I had been getting errors in the specs which I had to fix by letting the container start before trying to reap it (hence the `sleep 1`, etc) and mistook that as meaning I had to use the ID -- it's been a confusing evening on this track....

pbrisbin:
Commit message updated. Ready for re-review

lib/cc/analyzer/container.rb
```diff
@@ -104,17 +102,22 @@ def read_stderr(err)
...
         end
       end
 
       def container_data(duration: nil, status: nil)
         ContainerData.new(@image, @name, duration, status, @stderr_io.string)
       end
+
+      def reap_running_container
+        Analyzer.logger.warn("killing container name=#{@name}")
+        POSIX::Spawn::Child.new("docker", "kill", @name)
```
GordonDiggs: What happens if this command fails?

pbrisbin: Nothing :)

I'm not sure what would be worthwhile to do in such a case, any suggestions?

GordonDiggs: I guess the line above would get logged and we could see if orphaned containers were attempted to be killed. It's probably fine. The only suggestion I would make is to execute this command in some way that raises if it fails (ie. backticks)

pbrisbin: I would actually be against it raising if it fails because we're already in an error-handler of sorts and you'd end up masking what really happened with that failure -- maybe that'd be OK, I don't know.

GordonDiggs: Makes sense. I'm fine with trying this and seeing what happens. Might be a non-issue

GordonDiggs:
LGTM

PR #157, opened 2015-10-21
Split out bin/prep-release

pbrisbin:
Now that master is a protected branch, we can't script the entire release
process because we're required to open the version bump as a PR and merge in
the web UI.

This patch splits out a bin/pre-release for the tasks associated with opening
that PR. bin/release is then reduced to only the steps to perform after that PR
is merged to master.

/cc @codeclimate/review

These scripts were used to release v0.9.4


fhwang:
LGTM

PR #158, opened 2015-10-23
upgrade config: force existing exclude_paths to be an array

wfleming:
This is in response to a crash we just saw. I don't think this was
really valid config on classic either, but I guess we handled it there?

cc @codeclimate/review @brynary


brynary:
Going in a different direction.

PR #159, opened 2015-10-24
Include default tool configuration files in gem

jpignata:
Since we've moved to using the `codeclimate-init` script within Builder
we need to ensure that the default configurations are available within
the gem package so the initialization process can properly seed
repositories.


pbrisbin:
Nice catch. LGTM

PR #160, opened 2015-10-25
Analyzer::Engine no longer raises exceptions on container error or timeout

fhwang:
This PR is a possible first step in streamlining how Analyzer::Engine and Analyzer::Container communicate with collaborators. Container results are returned as an object, which can be used as desired by collaborators. Specifically, in doing recent debugging I've found cases where the exit status of the container process doesn't map automatically to one result or another, so it's best to simply return it and let others deal with the complexity.

Note that this PR is non-backwards compatible; once we merge this in we'll want to bump a minor version.


fhwang:
Closing this PR in favor of one that changes less on the codeclimate side.

PR #162, opened 2015-10-26
Make some Analyzer classes more flexible.

fhwang:


fhwang:
Hold off on review for now, there's a little more work necessary on this branch.

fhwang:
K, this is now ready for review. In the interest of cleaning up work in platform, I ended up changing two classes here:

# Analyzer::Container#run

Now returns a result object

# Analyzer::Engines

* Renamed to `Analyzer::EnginesBuilder`
* is-a Array behavior has been removed in favor of a more vanilla `run` method.
* `run` accepts an optional engine class. This way collaborators can take advantage of its config parsing functionality but still use a different engine class.

All of these changes are forward compatible.

@codeclimate/review

lib/cc/analyzer/engines_builder.rb
```diff
@@ -2,7 +2,7 @@
 
 module CC
   module Analyzer
-    class Engines < Array
+    class EnginesBuilder
       def initialize(registry:, config:, container_label:, source_dir:, requested_paths:)
         super()
```
pbrisbin: Probably don't need this now that you aren't subclassing `Array`

pbrisbin:
LGTM

PR #163, opened 2015-10-26
Fix unknown formatter error

dblandin:
This fixes an error I encountered when referencing an unknown
formatter. There's already logic for outputting a helpful error message,
but it looks like we broke that behavior at some point after moving
some references around.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,4 +8,5 @@ group :test do
   gem "minitest-around"
   gem "minitest-reporters"
   gem "rack-test"
+  gem "rake"
 end
--- a/lib/cc/analyzer/formatters.rb
+++ b/lib/cc/analyzer/formatters.rb
@@ -1,5 +1,7 @@
 module CC
   module Analyzer
+    autoload :InvalidFormatterError, "cc/analyzer/invalid_formatter_error"
+
     module Formatters
       autoload :Formatter,  "cc/analyzer/formatters/formatter"
       autoload :JSONFormatter, "cc/analyzer/formatters/json_formatter"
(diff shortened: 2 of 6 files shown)


pbrisbin:
Here's some context, if you're curious: https://github.com/codeclimate/codeclimate/pull/125

pbrisbin:
LGTM

dblandin:
@pbrisbin Do you have any insight into these CircleCI failures?

```
Error response from daemon: no such id: codeclimate-container-test
FATA[0000] Error: failed to kill one or more containers 
Error response from daemon: no such id: codeclimate-container-test
FATA[0000] Error: failed to remove one or more containers 
.Error response from daemon: no such id: codeclimate-container-test
FATA[0000] Error: failed to remove one or more containers 
Error response from daemon: no such id: codeclimate-container-test
FATA[0000] Error: failed to kill one or more containers 
```

Seems to be coming from the "stopping containers" spec?

pbrisbin:
Hmm. I can't see how what you're doing would impact that. The spec is a little shaky by the nature of what it's doing, but it's been robust enough so far...

pbrisbin:
I think those lines are a red herring. I can reproduce the abort locally and the messages that show differ each run (depending on spec order).

It looks like you're calling `InvalidFormatterError.new(name), "..."` which is probably a problem, maybe *the* problem?

dblandin:
This has to be one of the most unhelpful error messages:

```
❯ rake test TEST=spec/cc/cli/analyze_spec.rb                                                                                                                                                                                                      codeclimate/git/devon/fix-unknown-formatter-error !

# Running tests with run options --seed 27771:

.rake aborted!
Command failed with status (1): [ruby -I"lib:spec" -I"/usr/local/opt/rbenv/versions/2.2.2/lib/ruby/2.2.0" "/usr/local/opt/rbenv/versions/2.2.2/lib/ruby/2.2.0/rake/rake_test_loader.rb" "spec/cc/cli/analyze_spec.rb" ]
```

brynary:
This looks like it's getting stale. Should we close it?

dblandin:
Unfortunately, this is still an issue on master.

```bash
$ codeclimate analyze -f badformatter
error: (NameError) uninitialized constant CC::Analyzer::Formatters::InvalidFormatterError
```

I can revive this work and see if I can get the specs to pass

brynary:
Let's punt on this for the moment.

PR #167, opened 2015-10-27
Allow optional memory limit to be set in ENV

GordonDiggs:
Certain applications will want this to be a higher value than 512MB, so
allow that to be set in the environment.

@codeclimate/review ping @jpignata @noahd1

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -8,6 +8,8 @@ module CC
 
       attr_reader :name
 
+      DEFAULT_MEMORY_LIMIT = 512_000_000.freeze
+
       def initialize(name, metadata, code_path, config, label)
         @name = name
         @metadata = metadata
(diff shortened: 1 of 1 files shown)


lib/cc/analyzer/engine.rb
```diff
@@ -68,6 +68,11 @@ def config_file
       def output_filter
         @output_filter ||= EngineOutputFilter.new(@config)
       end
+
+      # Memory limit for a running engine in bytes
+      def memory_limit
+        (ENV["ENGINE_MEMORY_LIMIT"] || 512_000_000).to_s
```
noahd1: Do you want to suffix it with `_BYTES` so the unit is understood?
Also, maybe a constant for the default?

GordonDiggs: Agreed on both - will make that change

GordonDiggs: Fixed

jpignata:
:+1:

PR #169, opened 2015-10-27
Always enable fixme engine when generating engines config

dblandin:
The behavior added in this PR adds the fixme engine after matching against any file within the include paths.

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -134,7 +134,8 @@ fixme:
   description: Find FIXME, TODO, HACK, etc. comments
   community: false
   enable_regexps:
-  default_ratings_paths:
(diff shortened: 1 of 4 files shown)


wfleming:
Can we please do the rubocop config separately? Changing/enforcing our style guide should be separated from functional changes. And the PRs are a heck of a lot easier to review when they're separate. The functional one is naturally smaller, and the rubocop-style one almost doesn't need human review by definition if everything's green.

dblandin:
@wfleming Yep, just pruned it out. Will introduce the updated rubocop configuration in a separate PR.

wfleming:
LGTM

PR #170, opened 2015-10-27
Update rubocop configuration with CC default

dblandin:
This PR updates the rubocop configuration to the CC default and updates most of the low-hanging fruit to conform to the enforced rules.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -7,3 +7,4 @@ ratings:
   - "**.rb"
 exclude_paths:
   - .bundle/**/*
+  - spec/**/*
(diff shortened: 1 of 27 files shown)


pbrisbin:
Nice. I'd request a merge around the long-method for now.

wfleming:
Quick suggestion for refactoring that [long method](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/analyzer/source_buffer.rb#L31-L45):

```
def line_begins
  @line_begins ||= begin
    @line_begins, index = [ [ 0, 0 ] ], 1

    @source.each_char do |char|
      @line_begins.unshift [ @line_begins.length, index ] if char == "\n"

      index += 1
    end
  end
end
```

I think that will get you under the limit.

I'd also be up for boosting that limit a bit. 10 is quite restrictive.

PR #172, opened 2015-10-28
Specify remote and branch name when pushing to avoid git warning

dblandin:
I ran into this git warning when running the pre-release script:

```
warning: push.default is unset; its implicit value has changed in
Git 2.0 from 'matching' to 'simple'.
```

This PR specifies the origin and branch for a more explicit push which silences
the warning.


pbrisbin:
LGTM

PR #173, opened 2015-10-28
Include engine's default configuration when adding to `codeclimate.yml`

dblandin:
Some engines require a default configuration before they can work
properly. One such engine is the duplication engine which requires
languages to be specified before analysis. This pulls in any default
configuration specified in the engines registry when enabling an engine
within a `.codeclimate.yml` configuration file.

Follows codeclimate/codeclimate-duplication#24

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -25,6 +25,12 @@ duplication:
   community: false
   enable_regexps:
   default_ratings_paths:
+  default_config:
+    languages:
+      - ruby
+      - javascript
+      - python
+      - php
 gofmt:
   image: codeclimate/codeclimate-gofmt
   description: gofmt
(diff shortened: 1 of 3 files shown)


config/engines.yml
```diff
@@ -25,6 +25,12 @@ duplication:
   community: false
   enable_regexps:
   default_ratings_paths:
+  default_config:
```
fhwang: Not a blocker for this, but this file's getting kind of hairy. At some point somebody should probably break this up into individual YAML files in a `config/engines` directory or something.

fhwang:
LGTM

config/engines.yml
pbrisbin: It might be good enough to extract out a `&defaults` reference, so the engine keys can specify less if they care about less.

```yaml
defaults: &defaults
  beta: false
  community: false
  enable_regexps:
  default_ratings_paths:
  default_config:

gofmt:
  <<: *defaults
  image: ...
```

dblandin: That sounds like a good approach! I'll extract out some defaults in a separate PR.

PR #174, opened 2015-10-28
Cleanup engines registry by extracting defaults and alphabetizing

dblandin:
This cleans up the engines registry file by extracting some common defaults and
alphabetizing by engine name.

This change also has an effect on the order of engines added to a `.codeclimate.yml` file. The engines added will now appear in alphabetical order -> the same order they appear in within the engines registry.

PR #178, opened 2015-11-02
Alphabetize engines for easier addition going forward

ABaldwinHunter:
@codeclimate/review @BlakeWilliams @jpignata In anticipation of adding new engines to manifest, this change orders engines already present in `engines.yml` manifest so that it is easier to read and modify going forward.

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -9,16 +9,32 @@
 # the `default_ratings_paths` globs, which are used during analysis to determine
 # which files should be rated.
 #
-rubocop:
-  image: codeclimate/codeclimate-rubocop
-  description: A Ruby static code analyzer, based on the community Ruby style guide.
+bundler-audit:
+  image: codeclimate/codeclimate-bundler-audit
+  description: Patch-level verification for Bundler
   community: false
   upgrade_languages:
-   - Ruby
+    - Ruby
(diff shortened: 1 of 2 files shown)


ABaldwinHunter:
Interesting. The order in `engines.yml` affects the order in which engines are written to `.codeclimate.yml` during init. Is that intentional behavior? And/or do we want to decouple those things.

GordonDiggs:
Probably consequential behavior but IMO not undesirable

dblandin:
@ABaldwinHunter Yep. Ran into that in #174.

ABaldwinHunter:
@dblandin Ah got it.

@GordonDiggs Hm I don't see a ton of benefit in maintaining an order in the generated config as the list of engines grows.

But I do see some value.

If we want that behavior, I like keeping the hierarchy knowledge somewhere else (or perhaps giving an ID or pecking order property to the engine).

Benefit: `engines.yml` is easier to navigate and only acts as a simple source of information

ABaldwinHunter:
Wdyt? (I can also side step for now)

ABaldwinHunter:
@GordonDiggs @dblandin Fixed tests. Ready for re-review!

GordonDiggs:
LGTM

PR #181, opened 2015-11-03
Fixes to circle.yml

pbrisbin:
- Use dependency.override, not pre
- Use conventional PRIVATE_REGISTRY variable
- Build the CLI image as a dependency
- Run tests in a docker container

This is an attempt to address intermittent failures in the Container#stop specs
by running tests in a more consistent environment.

/cc @codeclimate/review

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,17 +3,20 @@ machine:
     - docker
   environment:
     CLOUDSDK_CORE_DISABLE_PROMPTS: 1
-    registry_root: us.gcr.io/code_climate
+    PRIVATE_REGISTRY: us.gcr.io/code_climate
 
 dependencies:
-  pre:
-    # Run in container_spec
+  override:
+    # Used by container_spec
     - docker pull alpine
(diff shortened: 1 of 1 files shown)


GordonDiggs:
LGTM

ABaldwinHunter:
Late: but why don't we make this look more like https://github.com/codeclimate/codeclimate-rubocop/blob/master/circle.yml

ABaldwinHunter:
Also, thanks for addressing this issue!

pbrisbin:
@ABaldwinHunter short answer: I didn't think about it.

Looking at it now though, I think the only difference is that that one runs `docker build` in the test step instead of dependencies -- that's actually a bug since not overriding the dependency step will cause an inferred `bundle install` step to run, which (as we've seen) can cause problems by creating config files in the source tree that end up in the docker image.

So I'd argue that rubocop's should actually be made to look like this.

ABaldwinHunter:
Okay cool. Thanks for the explanation.

PR #183, opened 2015-11-03
Improve container spec

pbrisbin:
Locally, if I run the spec in a loop I did see the assertion at line 196 fail
occasionally. This assertion is unnecessary IMO, so I removed it. On Circle, I
have seen the one at line 128 fail, so I've replaced it (and one other) with
better assertions that should provide more information next time they fail.

See commit messages for more details.

/cc @codeclimate/review


wfleming:
:+1:

PR #184, opened 2015-11-03
Increase upper limit in duration expectation

pbrisbin:
This assertion is designed to ensure the command was stopped and did not run to
completion. It seems the mechanics of stopping the container can take right
around 5 seconds and so a 5 second assertion fails intermittently on CI.

Example: https://circleci.com/gh/codeclimate/codeclimate/545

Given that the command, if not stopped, will run for 10 seconds, using that as
the assertion should reduce false-positives without risk of false-negatives.

/cc @codeclimate/review


wfleming:
:+1:

PR #186, opened 2015-11-04
better config upgrade behavior

wfleming:
* include newer cc/yaml for better error messages
* validate existing config before upgrading
* if config is not valid, emit errors & exit with status 1

Next step is `builder`.

Thoughts, @codeclimate/review?

### DIFF
--- a/lib/cc/analyzer/engines_builder.rb
+++ b/lib/cc/analyzer/engines_builder.rb
@@ -44,7 +44,7 @@ module CC
 
       def names_and_raw_engine_configs
         {}.tap do |ret|
-          @config.engines.each do |name, raw_engine_config|
+          (@config.engines || {}).each do |name, raw_engine_config|
             if raw_engine_config.enabled? && @registry.key?(name)
               ret[name] = raw_engine_config
             end
--- a/lib/cc/cli/config_generator.rb
+++ b/lib/cc/cli/config_generator.rb
@@ -17,6 +17,10 @@ module CC
         @engine_registry = engine_registry
       end
 
+      def can_generate?
+        true
+      end
+
       def eligible_engines
         return @eligible_engines if @eligible_engines
 
@@ -26,6 +30,10 @@ module CC
             result[name] = config
           end
         end
+      end
+
+      def errors
+        []
       end
 
       def exclude_paths
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -9,13 +9,11 @@ module CC
 
       def run
         if !upgrade? && filesystem.exist?(CODECLIMATE_YAML)
-          say "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
+          fatal "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
         elsif upgrade? && engines_enabled?
-          say "--upgrade should not be used on a .codeclimate.yml configured for the Platform.\nTry running 'validate-config' to check configuration."
+          fatal "--upgrade should not be used on a .codeclimate.yml configured for the Platform.\nTry running 'validate-config' to check configuration."
         else
-          create_codeclimate_yaml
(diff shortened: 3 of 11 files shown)


lib/cc/analyzer/engines_builder.rb
```diff
@@ -44,7 +44,7 @@ def engine_config(raw_engine_config)
 
       def names_and_raw_engine_configs
         {}.tap do |ret|
-          @config.engines.each do |name, raw_engine_config|
+          (@config.engines || []).each do |name, raw_engine_config|
```
fhwang: Out of curiosity, is this for the case where the `engines` array is entirely missing from the YAML?

wfleming: Kind of. Because of how CC::YAML works, if the engines value is invalid, it won't actually be present in the fully parsed node. An empty engines node is considered invalid now.

I was going to suggest that CLI commands that use config values should probably run `validate-config` before they do anything else. But I think that should be a separate PR. In this particular case this is just a change to maintain existing behavior due to changing dependency behavior.

GordonDiggs: We have activesupport here, right? Could this use `Array.wrap(@config.engines)`?

lib/cc/cli/config_generator.rb
```diff
@@ -17,6 +17,10 @@ def initialize(filesystem, engine_registry)
         @engine_registry = engine_registry
       end
 
+      def can_generate?
+        true
```
GordonDiggs: When would this be false?

lib/cc/cli/init.rb
```diff
@@ -33,6 +44,7 @@ def create_codeclimate_yaml
         end
 
         config.add_exclude_paths(config_generator.exclude_paths)
+        puts "config is now: #{config.inspect}"
```
GordonDiggs: Should this use `say`?

lib/cc/cli/upgrade_config_generator.rb
```diff
@@ -3,8 +3,16 @@
 module CC
   module CLI
     class UpgradeConfigGenerator < ConfigGenerator
+      def can_generate?
+        existing_yaml.errors.blank?
```
GordonDiggs: Should this call the `errors` method on this object?

lib/cc/cli/config_generator.rb
wfleming: For the `UpgradeConfigGenerator`: the stub is here for API consistency between the two.

GordonDiggs: EDIT: nvm - figured out what is happening here

lib/cc/cli/init.rb
wfleming: That...was debugging code :disappointed: . Removed.

lib/cc/cli/upgrade_config_generator.rb
wfleming: Sure. Done.

lib/cc/analyzer/engines_builder.rb
wfleming: I guess we could. Is it worth it, though? It's longer to write, and `wrap` has some surprising semantics, and the returned object is a YAML node that already responds to `each`, so why wrap it if it's already present?

wfleming: Just tried this locally, and it broke… a lot of things. `Array.wrap` & YAML `Node`s both have some surprising behavior, and they do not play well together.

@GordonDiggs do you have other concerns around this PR?

GordonDiggs:
LGTM

lib/cc/analyzer/engines_builder.rb
ABaldwinHunter: I think `{}` is slightly better than `[]` for the default `engines`, because it's usually a hash not array

wfleming: Good point. I'll change that.

PR #187, opened 2015-11-04
Add radon engine to manifest

ABaldwinHunter:
The radon engine is currently generated from a [CC fork of the
radon project](https://github.com/mrb/radon), with hopes to be
integrated natively into the orginal project quickly.

Although the radon tool provides many code metrics, including
raw metrics, maintainability, halstead metrics and cyclomatic
complexity, the present engine wrapper will only report on
complexity issues.

More information about radon can be found at:
https://github.com/rubik/radon

Users can configure the minimum level of complexity they'd like to have
reported in their .codeclimate.yml file (A-F):

```
engines:
  radon:
    enabled: true
    config:
      threshold: 'd'
```

following, radon's conventions.

For more information about how radon grade levels translate to
cyclomatic complexity, please see the [radon docs](http://radon.readthedocs.org/en/latest/commandline.html).

@codeclimate/review @mrb @jpignata @BlakeWilliams

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -141,6 +141,14 @@ phpmd:
     - "**.php"
     - "**.module"
     - "**.inc"
+radon:
+  image: codeclimate/codeclimate-radon
+  description: Python tool used to compute Cyclomatic Complexity.
+  community: true
+  enable_regexps:
+    - \.py$
+  default_ratings_paths:
+    - "**.py"
 requiresafe:
   image: codeclimate/codeclimate-requiresafe
   description: Security tool for Node.js dependencies


config/engines.yml
```diff
@@ -141,6 +141,14 @@ phpmd:
     - "**.php"
     - "**.module"
     - "**.inc"
+radon:
+  image: codeclimate/codeclimate-radon
+  description: Python tool used here to compute Cyclomatic Complexity from the source code.
```
GordonDiggs: I would word this more simply as

> Python tool used to compute Cyclomatic Complexity.

wfleming: "here" seems a bit unnecessary.

wfleming: As does "from the source code."

ABaldwinHunter: Okay sounds good. Making change.

(I mostly cargo-culted from radon's [self description](https://github.com/rubik/radon/blob/master/README.rst))

ABaldwinHunter:
Thanks @GordonDiggs @wfleming for feedback.

Ready for re-review!

wfleming:
LGTM.

Unrelated nit: I notice we're inconsistent about whether descriptions get periods at the end or not. I'm making a note to do that in a separate PR.

GordonDiggs:
LGTM2

config/engines.yml
```diff
@@ -141,6 +141,14 @@ phpmd:
     - "**.php"
     - "**.module"
     - "**.inc"
+radon:
+  image: codeclimate/codeclimate-radon
+  description: Python tool used to compute Cyclomatic Complexity.
+  community: true
+  enable_regexps:
```
wfleming: Related to earlier discussion around `upgrade_languages`, it feels weird to me that we include `enable_regexps` for `community` engines that we'll never auto-enable. Is the thinking that eventually we'll remove the community restriction for auto-enable?

ABaldwinHunter:
@wfleming thanks! haha I noticed that too. Figured it would be better fixed in a separate pr

config/engines.yml
ABaldwinHunter: @wfleming Good question. I included the `enable_regexps` because that seemed to be the pattern present for the other community engines, like gofmt.

However, I personally think we'll auto-enable certain community engines eventually and I'd advocate that.

That seems like the way we'd offer the same easy first-time-analysis-without-config experience to users with projects in languages like `python`, and `go`

wfleming: I agree, but I think then it also makes sense to include `upgrade_languages`, doesn't it?

ABaldwinHunter: @wfleming Ah sorry saw this too late!

I'm happy with either way.

I could open a new PR making that change for all of them if we wanted. WDYT?

wfleming: It's not urgent, and I saw your earlier change for `brakeman` too late to discuss it there. I just think we should figure out what the correct way to be consistent with these engines is: if we plan to auto-enable community engines at some point, we should probably include both `enable_regexps` & `upgrade_languages`. If we don't plan to auto-enable `community` engines, we should probably include neither.

ABaldwinHunter: Cool. I agree

PR #190, opened 2015-11-04
pre-release: use hub, not gh

wfleming:
gh is deprecated: prefer hub if installed, but still support gh for
those who have it

cc @codeclimate/review


mxie:
LGTM

dblandin:
:+1:

PR #191, opened 2015-11-04
Replace EnginesBuilder with EnginesConfigBuilder.

fhwang:
Giving builder more flexibility in how Engine instances are instantiated will make it easier to make certain classes more cohesive. This refactor will be followed up by one on the builder side.

Note that this PR breaks backwards compatibility, so I'll follow it up soon with a small refactor in builder.

### DIFF
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -29,14 +29,6 @@ module CC
 
       private
 
-      def build_engine(engine_class, name, raw_engine_config)
-        label = @container_label || SecureRandom.uuid
-        engine_config = engine_config(raw_engine_config)
-        engine_class.new(
-          name, @registry[name], @source_dir, engine_config, label
-        )
-      end
-
       def engine_config(raw_engine_config)
         config = raw_engine_config.merge(
           exclude_paths: exclude_paths,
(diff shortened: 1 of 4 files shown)


fhwang:
@codeclimate/review this is ready for a look, thanks!

lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -11,9 +19,11 @@ def initialize(registry:, config:, container_label:, source_dir:, requested_path
         @source_dir = source_dir
       end
 
-      def run(engine_class = Analyzer::Engine)
+      def run
         names_and_raw_engine_configs.map do |name, raw_engine_config|
-          build_engine(engine_class, name, raw_engine_config)
+          label = @container_label || SecureRandom.uuid
```
GordonDiggs: What do you think of doing the `|| SecureRandom.uuid` in the constructor?

fhwang: The desired behavior, based at least on the previous code, was to call `SecureRandom.uuid` once for each engine, so I'm just preserving that here.

GordonDiggs: Ah got it

GordonDiggs:
LGTM

PR #192, opened 2015-11-05
Rename requiresafe to nodesecurity

jpignata:
^lift has merged two products and settled on this name. We'll support
both the old and new names for the time being for backwards
compatibility.

@codeclimate/review


GordonDiggs:
LGTM

PR #195, opened 2015-11-06
Test subcommand

brynary:
This PR creates the ability to invoke `codeclimate test engine_name` and have the CLI run a suite of regression tests for the engine.

The engine declares it's regression suite via the `test_paths` key in the new `engine.json` file (which must be in the root of the engine's filesystem) like this:

```json
{
  "name": "fixme",
  "description": "Discover lurking FIXMEs, BUGs, TODOs, etc. in your code",
  "maintainer": {
    "name": "Michael R. Bernstein",
    "email": "mrb@codeclimate.com"
  },
  "test_paths": [
    "/usr/src/app/test/integration/"
  ]
}
```

The test paths are unpacked, and we run every file in the directory as a test. (By invoking `codeclimate analyze` with the `--engine` option and also a single file path.) Tests are implemented as comments and look like the following:

```ruby
# [issue] check_name=BUG description="BUG found" category="Bug Risk"
# BUG: This should use double equals
if (foo = "foo")
  puts "You said foo"
end
```

The present of an `[issue]` tag indicates that the next line should report an issue. Each issue "marker" can specify as many or as few Issue attributes as it cares to.

Results:

screenshot 2015-11-06 03 46 49

Known issues:

* Only works for issues that are emitted as [line-based locations](https://github.com/codeclimate/spec/blob/master/SPEC.md#locations) which are single line. Relatively easy to fix, as we'll just allow you to specify `location.positions.begin.line=X location.positions.begin.end=Y` attributes to be included inline.
* Won't work for analysis that requires an entire directory (e.g. Brakeman) yet. Deferred for now.
* It's not possible to specify more than one Issue category as array syntax breaks the regexp parser. Not an issue for now.

/c @noahd1 @mrb @codeclimate/review

### DIFF
--- a/bin/codeclimate
+++ b/bin/codeclimate
@@ -4,5 +4,6 @@ $LOAD_PATH.unshift(File.expand_path(File.join(File.dirname(__FILE__), "../lib"))
 require "cc/cli"
 
 ENV["FILESYSTEM_DIR"] ||= "."
+ENV["CODE_PATH"] ||= ENV["PWD"]
 
 CC::CLI::Runner.run(ARGV)
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -63,6 +63,7 @@ module CC
 
       def config_file
         path = File.join("/tmp/cc", SecureRandom.uuid)
+        FileUtils.mkdir_p("/tmp/cc")
         File.write(path, @config.to_json)
         path
       end
--- a/lib/cc/cli/command.rb
+++ b/lib/cc/cli/command.rb
@@ -24,7 +24,7 @@ module CC
         run
       end
 
-      def say(message)
+      def say(message = "")
         terminal.say message
       end
 
--- a/lib/cc/cli/command.rb
+++ b/lib/cc/cli/command.rb
@@ -24,7 +24,7 @@ module CC
         run
       end
 
-      def say(message = "")
+      def say(message)
         terminal.say message
       end
 
--- a/lib/cc/cli/test.rb
+++ b/lib/cc/cli/test.rb
@@ -86,7 +86,7 @@ module CC
       end
 
       def issues_in(test_file)
-        issue_docs = capture(:stdout) do
+        issue_docs = capture_stdout do
           codeclimate_analyze(test_file)
         end
 
@@ -210,6 +210,22 @@ module CC
 
       def engine_image
         engine_registry[@engine_name]["image"]
+      end
+
+      # Stolen from ActiveSupport (where it was deprecated)
+      def capture_stdout
+        captured_stream = Tempfile.new("stdout")
+        origin_stream = $stdout.dup
+        $stdout.reopen(captured_stream)
+
+        yield
+
+        $stdout.rewind
+        return captured_stream.read
+      ensure
+        captured_stream.close
+        captured_stream.unlink
+        $stdout.reopen(origin_stream)
       end
 
     end
--- a/lib/cc/cli/test.rb
+++ b/lib/cc/cli/test.rb
@@ -1,36 +1,118 @@
+require "shellwords"
 require "cc/yaml"
 
 module CC
   module CLI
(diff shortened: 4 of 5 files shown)


bin/codeclimate
```diff
@@ -4,5 +4,6 @@ $LOAD_PATH.unshift(File.expand_path(File.join(File.dirname(__FILE__), "../lib"))
 require "cc/cli"
 
 ENV["FILESYSTEM_DIR"] ||= "."
+ENV["CODE_PATH"] ||= ENV["PWD"]
```
brynary: This was a change I had to make to get things working in my local dev environment (where I run everything within a Docker container. Since it's an `||=` it seemed safe enough to commit)

lib/cc/analyzer/engine.rb
```diff
@@ -63,6 +63,7 @@ def container_name
 
       def config_file
         path = File.join("/tmp/cc", SecureRandom.uuid)
+        FileUtils.mkdir_p("/tmp/cc")
```
brynary: This was a change I had to make to get things working in my local dev environment (where I run everything within a Docker container). This directory seems to need to exist to write the config, and I don't see where we are creating it. This seemed safe enough, and should no-op if it is already there.

brynary:
Also see https://github.com/codeclimate/codeclimate-fixme/pull/12

lib/cc/cli/test.rb
```diff
@@ -0,0 +1,233 @@
...
+        codeclimate_path = File.expand_path(File.join(File.dirname(__FILE__), "../../../bin/codeclimate"))
+
+        system([
+          "unset CODE_PATH &&",
+          "unset FILESYSTEM_DIR &&",
+          codeclimate_path,
+          "analyze",
+          "--engine", @engine_name,
+          "-f", "json",
+          relative_path
```
wfleming: I think you probably want this quoted in case of spaces. (Possibly ditto for the `codeclimate_path`)

lib/cc/cli/test.rb
```diff
@@ -0,0 +1,233 @@
...
+      def remove_null_container
+        `docker rm -f #{null_container_id}` if null_container_id
+      end
+
+      def test_paths
+        Array.wrap(engine_spec["test_paths"])
+      end
+
+      def engine_spec
+        @engine_spec ||= JSON.parse(`docker run --rm #{engine_image} cat /engine.json`)
```
wfleming: I don't see any error handling for if this is missing right now: seems worth adding since not all engines are going to have this file immediately.

lib/cc/cli/command.rb
```diff
@@ -24,7 +24,7 @@ def execute
         run
       end
 
-      def say(message)
+      def say(message = "")
```
wfleming: You're only using this in one place, and I personally find the idea of calling `say` without params a bit surprising when reading the calling code. Is this really better than being explicit at the call site?

lib/cc/cli/test.rb
```diff
@@ -0,0 +1,233 @@
...
+          say "PASS %3d: %s" % [marker.line, marker.line_text]
+          actual_issues.delete_at(found_index)
+        else
+          say colorize("MISS %3d: %s" % [marker.line, marker.line_text], :red)
+          say colorize("Searched:", :yellow)
+          say colorize(JSON.pretty_generate(marker.issue), :yellow)
+          say "\n"
+          say colorize("Actual:", :yellow)
+          say colorize(JSON.pretty_generate(actual_issues), :yellow)
+          fatal "Expected issue not found."
```
wfleming: Aborting the entire program as soon as you hit one missed issue seems harsh. I think it would be more useful to act a bit more like a test suite & emit all issues (hits or misses) before exiting.

lib/cc/cli/test.rb
brynary: The error that comes back is basically "File not found /engine.json" so it seems OK for now.

PR #196, opened 2015-11-06
Add default coffeelint.json file

GordonDiggs:
Finding these defaults is non-obvious, so I generated the default
config. More info on that: http://www.coffeelint.org/#usage

Having this here will make it more transparent what checks we're running
and make it easier for people to configure their analysis.

@codeclimate/review


wfleming:
:+1:

PR #197, opened 2015-11-08
Enforce a maximum output byte count on containers

jpignata:
Currently engines can generate an unbounded amount of output on STDOUT
and STDERR while running. This is bad for a variety of reasons, but the
symptom most pressing is that Docker has a memory leak around dealing
with large STDOUT streams especially when there are "slow" readers
attached.

This implements a concept of maximum output bytes that when tripped will
stop the container. The maximum output bytes is defaulted to 150
megabytes.
    
* Consolidate Result and ContainerData into ContainerStatus struct
* Suppress STDERR on git ls-files calls

cc: @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -1,4 +1,5 @@
 require "posix/spawn"
+require "thread"
 
 module CC
   module Analyzer
@@ -11,9 +12,17 @@ module CC
         :stderr,        # stderr, for a finished event
       )
       ImageRequired = Class.new(StandardError)
-      Result = Struct.new(:exit_status, :timed_out?, :duration, :stderr)
+      Result = Struct.new(
+        :exit_status,
+        :timed_out?,
+        :duration,
+        :maximum_output_exceeded?,
+        :output_byte_count,
+        :stderr,
+      )
 
       DEFAULT_TIMEOUT = 15 * 60 # 15m
+      DEFAULT_MAXIMUM_OUTPUT_BYTES = 500_000_000
 
       def initialize(image:, name:, command: nil, listener: ContainerListener.new)
         raise ImageRequired if image.blank?
@@ -24,7 +33,10 @@ module CC
         @output_delimeter = "\n"
         @on_output = ->(*) {}
         @timed_out = false
+        @maximum_output_exceeded = false
         @stderr_io = StringIO.new
+        @output_byte_count = 0
+        @counter_mutex = Mutex.new
       end
 
       def on_output(delimeter = "\n", &block)
@@ -43,14 +55,23 @@ module CC
         t_timeout = timeout_thread
 
         _, status = Process.waitpid2(pid)
+
         if @timed_out
-          @listener.timed_out(container_data(duration: timeout))
-          Result.new(status.exitstatus, true, timeout, @stderr_io.string)
+          duration = timeout
+          @listener.timed_out(container_data(duration: duration))
         else
           duration = ((Time.now - started) * 1000).round
           @listener.finished(container_data(duration: duration, status: status))
-          Result.new(status.exitstatus, false, duration, @stderr_io.string)
         end
+
+        Result.new(
+          status.exitstatus,
+          @timed_out,
+          duration,
+          @maximum_output_exceeded,
+          output_byte_count,
+          @stderr_io.string,
+        )
       ensure
         t_timeout.kill if t_timeout
         if @timed_out
@@ -71,6 +92,8 @@ module CC
 
       private
 
+      attr_reader :output_byte_count, :counter_mutex
+
       def docker_run_command(options)
         [
           "docker", "run",
@@ -88,13 +111,17 @@ module CC
             output = chunk.chomp(@output_delimeter)
 
             @on_output.call(output)
+            check_output_bytes(output.bytesize)
           end
         end
       end
 
       def read_stderr(err)
         Thread.new do
-          err.each_line { |line| @stderr_io.write(line) }
+          err.each_line do |line|
+            @stderr_io.write(line)
+            check_output_bytes(line.bytesize)
+          end
         end
       end
 
@@ -103,6 +130,17 @@ module CC
           sleep timeout
           @timed_out = true
           reap_running_container
+        end
+      end
+
+      def check_output_bytes(last_read_byte_count)
+        counter_mutex.synchronize do
+          @output_byte_count += last_read_byte_count
+        end
+
+        if output_byte_count > maximum_output_bytes
+          @maximum_output_exceeded = true
+          stop
         end
       end
 
@@ -116,7 +154,11 @@ module CC
       end
 
       def timeout
-        (ENV["CONTAINER_TIMEOUT_SECONDS"] || DEFAULT_TIMEOUT).to_i
+        ENV.fetch("CONTAINER_TIMEOUT_SECONDS", DEFAULT_TIMEOUT).to_i
+      end
+
+      def maximum_output_bytes
+        ENV.fetch("CONTAINER_MAXIMUM_OUTPUT_BYTES", DEFAULT_MAXIMUM_OUTPUT_BYTES).to_i
       end
     end
   end
--- a/lib/cc/analyzer/include_paths_builder.rb
+++ b/lib/cc/analyzer/include_paths_builder.rb
@@ -49,8 +49,8 @@ module CC
           tmp.write(File.read(".gitignore")) if File.file?(".gitignore")
           tmp << @cc_exclude_paths.join("\n")
           tmp.close
-          tracked_and_ignored = `git ls-files -zi -X #{tmp.path}`.split("\0")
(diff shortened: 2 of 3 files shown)


jpignata:
Opening early for thoughts since this area of the code seems to be under some churn recently. Also it passes locally on OS X but seems to fail in CI.

fhwang:
A few things:

# ContainerData and Result

It's not an easy change, conceptually, to merge these two. We've been sending `ContainerData` to listeners at container-specific events (start / finish / timeout) and returning `Result` to whoever owns the `Container` instance when the Docker instance terminates. The listener and the owner of the `Container` instance are usually not the same object.

# Container responsibilities

It's my opinion that we got ourselves into a lot of trouble by trying to make `Container` smart. This class is used for many different things, from running a simple `git clone` to invoking a very complex 3rd-party engine that could take many minutes and push out a lot of output. All those different use-cases have different ranges of what's appropriate when it comes to expected stdout, stdout delimiters, expected stderr, quantity of streamed output, exit status, etc. I think we'd be better served by making `Container` focused and simple, and giving `Container` owners lots of flexibility in how to determine behavior.

For example, in `builder` we currently listen to engine output, and if we detect unparseable output, we stop the container immediately. Although I'm not clear on specifically what problem you're trying to solve, I think that we could so something similar here as well.

jpignata:
Let's focus on Container responsibilities as the struct merge was just a target of opportunity since they had mostly overlapping attributes and might be moot depending on how this shakes out.

The problem I'm trying to solve is that containers can generate an unbounded amount of output. This is bad for a variety of reasons, but the symptom I'm most interested in solving is that Docker has a memory leak around dealing with large STDOUT streams especially when there are "slow" readers attached. Just as we enforce a maximum runtime for containers, I'd like to enforce a maximum output byte number. I picked this spot to add it because I'm concerned enforcing it upstream will still force Docker to flush bytes generated and buffered within it and I'd prefer not to generate those bytes to begin with if possible. Also, conceptually, I see something like output parsing to be a higher level concern than a gate on output. This concept feels in the same family as the timeout of the running process.

So:

1) Can I generalize this in such a way where it is applicable to all payload processes of container and not seemingly too targeted at engines? i.e., can this concept exist in `Container`?

2) If there's no way this concept should exist in `Container`, where can it exist and achieve the goal of eliminating our irritation of the Docker daemon under high output scenarios?

Thanks, @fhwang!

fhwang:
That's helpful, thanks. I agree that stopping itself if it emits a massive amount of data is properly the responsibility of `Container`. And on thinking about it more I think that a 150MB limit is so high that it makes a pretty safe cutoff.

I do wonder about forcing all listener classes to add a `maximum_output_exceeded` method. (I'd actually like to move away from our use of listeners objects overall, precisely because of this sort of inflexibility.) We have quite a few of them now.

Another option might to have `Container` check if its listener implements `maximum_output_exceeded`, and if not, to simply call `finished` with a slightly different `Result` object. That would minimize code churn at some cost in conceptual muddiness. But it is a _kind_ of finish, so ...

jpignata:
@fhwang I wonder if we could go further and collapse the listener interface to `started` and `finished` and put the onus on the listener itself to figure out the result by inspecting the passed `ContainerData` object. Kind of 💩s on tell don't ask, but it might be worth it as you indicated. Being killed for running time, too much output, memory consumption (which we need to capture and pass along state and turn it into an error code), does qualified as `finished` as you say.

So on the struct merge: it does seem as though these have mostly overlapping attributes. The path I was going down was trying to figure out if they were both going into the direction of the same concept – some kind of snapshot of the status of the `Container` whether running or complete. Despite the recipient of each struct being different, it seems like if they are interested in the same information then perhaps they can use the same delivery mechanism. WDYT? Too fraught to pursue?

fhwang:
I guess, since I've spent literally a week refactoring this code, I don't know how much more we want to do in this area for now. But in terms of priorities, it's really not my call :)

I personally would support reducing listeners to only having `started` and `finished` methods. I think that change wouldn't be too difficult.

I'd also support conceptually merging `finished` and the returned `Result` object. But I think that change could be more difficult, because it'll be tricky to make sure everything is rewired correctly. The tests are good in that area, though, so we might do okay.

But it would be the right direction, I think. For engines, having too many objects receiving signals from `Container` has introduced a lot of unnecessary complexity. This definitely added to the time it took to fix E12 and E13 reporting.

(Eventually I'd love to change `Container` so that it doesn't even take a listener object, and `Container#run` doesn't return an end result value. It just has three callbacks: `on_stdout`, `on_stderr`, and `on_event`. I don't think anything here goes against that, though.)

jpignata:
@fhwang Got it. Your desired end state seems right to me. Thanks for thinking this through with me. 

I should take the shortest path of work that accomplishes the goal of terminating containers that generate more than a defined limit of output while ensuring that we drive that error through to all collaborates of `Container` so that users will get correct error feedback. I think I can still accomplish that if I unwind my consolidation, not tinker with the callback methods right now and just extend `Result`.

jpignata:
@fhwang Updated the pull with in accordance with the above thought.

lib/cc/analyzer/container.rb
```diff
@@ -116,7 +153,15 @@ def reap_running_container
...
-        (ENV["CONTAINER_TIMEOUT_SECONDS"] || DEFAULT_TIMEOUT).to_i
+        ENV.fetch("CONTAINER_TIMEOUT_SECONDS", DEFAULT_TIMEOUT).to_i
+      end
+
+      def maximum_output_bytes
+        ENV.fetch("CONTAINER_MAXIMUM_OUTPUT_BYTES", DEFAULT_MAXIMUM_OUTPUT_BYTES).to_i
+      end
+
+      def counter_mutex
+        @counter_mutex ||= Mutex.new
```
fhwang: I can't find a solid reference for this, but `||=` is actually two operations and thus not thread-safe, right? It's hard to imagine what sort of bug this could cause, but it might just be best to initialize this before spawning secondary threads.

fhwang:
LGTM, other than my question about `||=` and the Code Climate issue.

lib/cc/analyzer/container.rb
jpignata: Yep, that's right. It could lead to a slightly inaccurate byte count if both streams fire at once. Moving into the initializer.

lib/cc/analyzer/container.rb
```diff
@@ -116,7 +154,11 @@ def reap_running_container
       end
 
       def timeout
-        (ENV["CONTAINER_TIMEOUT_SECONDS"] || DEFAULT_TIMEOUT).to_i
+        ENV.fetch("CONTAINER_TIMEOUT_SECONDS", DEFAULT_TIMEOUT).to_i
+      end
+
+      def maximum_output_bytes
+        ENV.fetch("CONTAINER_MAXIMUM_OUTPUT_BYTES", DEFAULT_MAXIMUM_OUTPUT_BYTES).to_i
```
GordonDiggs: nice thanks for putting this in the ENV

lib/cc/analyzer/container.rb
```diff
@@ -116,7 +154,11 @@ def reap_running_container
       end
 
       def timeout
-        (ENV["CONTAINER_TIMEOUT_SECONDS"] || DEFAULT_TIMEOUT).to_i
+        ENV.fetch("CONTAINER_TIMEOUT_SECONDS", DEFAULT_TIMEOUT).to_i
```
GordonDiggs: One important behavioral difference here is that if `ENV["CONTAINER_TIMEOUT_SECONDS"]` exists but is nil, this will be 0

jpignata: Hmm, could that happen? I'd think it could either be empty string (which would be 0) or a value, but I'm not sure how a `nil` value could be present in `ENV`.

GordonDiggs: I'm not sure on the behavior of this snippet:

```
CONTAINER_TIMEOUT_SECONDS=
```

But the behavior would be the same either way. What does a `0` timeout or max bytes mean behavior-wise?

wfleming: That snippet results in `""` in `irb`.

I'm pretty sure the only way you can get a `nil` in `ENV` is if the Ruby process explicitly sets it.

jpignata: @GordonDiggs a `0` timeout would result in immediate failure as would `0` max bytes. I think these inputs would result in the same behavior with the `fetch` or the `||`.

GordonDiggs: Ah yeah good point

GordonDiggs:
LGTM

jpignata:
Thoughts on the CC issue? I added some lines for clarity, and I'd rather not prune them out. I don't see any good places for a quick extraction or compaction.

fhwang:
Nothing quick comes to mind, and personally I wouldn't mind a mercy merge here.

One possibility for extraction: Since the max output logic is primarily listening to stderr and stdout, you could extract a separate class for enforcing that limit if `Container` were to take multiple `on_stdout` and `on_stderr` callbacks. I don't think that would have to be done now, though.

PR #200, opened 2015-11-10
Automatically enable brakeman, pep8 and radon

ABaldwinHunter:
This change makes the above engines become automatically turned on when a user runs `codeclimate init` or an analysis on .com without a `.codeclimate.yml` (using an inferred config).

@codeclimate/review @jpignata

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -13,6 +13,8 @@ brakeman:
   image: codeclimate/codeclimate-brakeman
   description: Static analysis tool which checks Ruby on Rails applications for security vulnerabilities.
   community: false
+  upgrade_languages:
+    - Ruby
   enable_regexps:
     - ^script\/rails$
   default_ratings_paths:
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -12,9 +12,9 @@
 brakeman:
   image: codeclimate/codeclimate-brakeman
   description: Static analysis tool which checks Ruby on Rails applications for security vulnerabilities.
-  community: true
+  community: false
   enable_regexps:
-    - \.rb$
+    - ^script\/rails$
   default_ratings_paths:
     - "app/**"
     - "**.rb"
@@ -114,7 +114,9 @@ nodesecurity:
 pep8:
   image: codeclimate/codeclimate-pep8
   description: Static analysis tool to check Python code against the style conventions outlined in PEP-8.
-  community: true
+  community: false
+  upgrade_languages:
+    - Python
   enable_regexps:
     - \.py$
   default_ratings_paths:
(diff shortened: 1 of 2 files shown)


config/engines.yml
```diff
@@ -12,9 +12,9 @@
 brakeman:
   image: codeclimate/codeclimate-brakeman
   description: Static analysis tool which checks Ruby on Rails applications for security vulnerabilities.
-  community: true
+  community: false
   enable_regexps:
-    - \.rb$
+    - ^script\/rails$
```
wfleming: We probably want to add back `upgrade_languages` here: the upgrade languages are an and condition with the `enable_regexps`: classic repos need to trigger both to get the engine.

ABaldwinHunter: Thanks @wfleming! Good catch

ABaldwinHunter: @wfleming Actually, taking a look at the code, it seems like `upgrade_languages`, if present, serve as an [additional filtering check](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/cli/upgrade_config_generator.rb#L24) on top of `regex` and `files _present`, but that the default will be whatever the [latter](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/cli/config_generator.rb#L57) comes up as.

If that's the case, it seems to me that the additional languages check doesn't add much value, and might cause harm in exceptional cases (e.g. rails repos that get added with `javascript` as default language bc of settings in Github).

wfleming:
> If that's the case, it seems to me that the additional languages check doesn't add much value here, and might cause harm in exceptional cases (e.g. rails repos that get added with javascript as default language bc of settings in Github).

These exceptional cases are actually why I think we *should* add `upgrade_languages`. The logic in `UpgradeConfigGenerator` is what it is because if you did have a classic configuration, and it has (for example), *.rb files in it, and you explicitly did not have `Ruby` in your yaml, then we should not auto-enable any Ruby engines. So *without* upgrade languages here, we would end up enabling brakeman on a repo that had a `script/rails` file but had been explicitly configured to *not* have Ruby analysis on classic, which would not be correct.

Languages used during upgrade aren't really connect to GitHub repo language recognition, by the way: they come directly from the user's checked-in `.codeclimate.yml` when it's configured for classic.

ABaldwinHunter:
@wfleming thanks for the detailed explanation and feedback!

Okay, I see what you're saying about the config. (Yeah, I was thinking those languages were coming from `repo.languages` not a traditional `.codeclimate.yml`).

I still think it's an unusual edge case (and hope we move away from the upgrade logic once the crossover happens), but adding `upgrade_languages` sounds good to me.

Added.

wfleming:
:+1:

PR #203, opened 2015-11-10
Ensure proper sorting of commingled location types

jpignata:
The previous sorting function didn't work properly with a mixed set of
locations that had positions with columns and lines. This changes the
sorting mechanism to use an array so we could indicate a secondary sort
criterion.


ABaldwinHunter:
LGTM

GordonDiggs:
:+1:

PR #204, opened 2015-11-10
Build include_paths once

BlakeWilliams:
Currently `include_paths` is called in `EnginesConfigBuilder` which is
initialized once for each engine which means we may have to build
`include_paths` several times when analyzing a repo.

Since building `include_paths` can be slow this hoists `include_paths`
into `EnginesRunner` and passes `include_paths` to
`EnginesConfigBuilder`.

### DIFF
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -11,11 +11,11 @@ module CC
         :container_label,
       )
 
-      def initialize(registry:, config:, container_label:, source_dir:, requested_paths:)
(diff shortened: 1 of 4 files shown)


wfleming:
:+1:

wfleming:
Seems like a great quick win. Good catch!

BlakeWilliams:
Thanks! Thanks for taking a look!

ABaldwinHunter:
I'm late but second that. Nice catch!

GordonDiggs:
Very nice. Would be interested to see some benchmarks on this

PR #205, opened 2015-11-11
Add default environments to default eslintrc

jpignata:
Right now it doesn't specify default environments which results in
variables like $ and jQuery being reported as issues. This adds
browser, node, jquery, amd, and commonjs as default environments as a
starting point for repos running with an inferred configuration.


fhwang:
LGTM

PR #206, opened 2015-11-11
Filter ignored issues

BlakeWilliams:
Engines now have a config option named `ignored_issues` which takes a
sequence of fingerprint values to check against reported issues. If an
issue matches one of the values of `ignored_issues` we filter it out.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     codeclimate (0.14.2)
       activesupport (~> 4.2, >= 4.2.1)
-      codeclimate-yaml (~> 0.5.0)
+      codeclimate-yaml (~> 0.6.0)
       faraday (~> 0.9.1)
       faraday_middleware (~> 0.9.1)
       highline (~> 1.7, >= 1.7.2)
@@ -24,7 +24,7 @@ GEM
       tzinfo (~> 1.1)
     ansi (1.5.0)
     builder (3.2.2)
-    codeclimate-yaml (0.5.0)
+    codeclimate-yaml (0.6.0)
       activesupport
       secure_string
     coderay (1.1.0)
@@ -78,6 +78,3 @@ DEPENDENCIES
   mocha
   rack-test
   rake
-
-BUNDLED WITH
-   1.10.6
--- a/lib/cc/analyzer/engine_output_filter.rb
+++ b/lib/cc/analyzer/engine_output_filter.rb
@@ -30,7 +30,19 @@ module CC
       end
 
       def ignore_issue?(json)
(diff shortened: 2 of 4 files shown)


Gemfile.lock
```diff
@@ -3,7 +3,7 @@ PATH
   specs:
     codeclimate (0.14.2)
       activesupport (~> 4.2, >= 4.2.1)
-      codeclimate-yaml (~> 0.5.0)
+      codeclimate-yaml (~> 0.6.0)
```
GordonDiggs: Is this dependency true? The [gemspec still specifies 0.5](https://github.com/codeclimate/codeclimate/blob/master/codeclimate.gemspec#L21)

BlakeWilliams: Yes, it's updated in this branch.

lib/cc/analyzer/engine_output_filter.rb
```diff
@@ -30,9 +30,21 @@ def issue?(json)
       end
 
       def ignore_issue?(json)
+        check_enabled?(json) || ignore_fingerprint?(json)
+      end
+
+      def check_enabled?(json)
```
GordonDiggs: Should this be named `check_disabled?`?

Gemfile.lock
GordonDiggs: Oops sorry - I thought this was builder.

.rubocop.yml
```diff
@@ -16,6 +16,9 @@ Lint/Debugger:
   Exclude:
   - lib/cc/cli/console.rb
 
+Lint/AssignmentInCondition:
```
GordonDiggs: Why disable this? [Our styleguide](https://github.com/codeclimate/styleguide/blob/master/rubocop.yml) does not

BlakeWilliams: I can remove it if we prefer that. I only disabled it because CC reported it and was failing the build.

Do we prefer not to have assignment in condition? If there's no consensus I wouldn't mind making a PR to the style guide.

Gemfile.lock
BlakeWilliams: No worries! :+1:

.rubocop.yml
GordonDiggs: We're ok with it. IIRC the rule specifies that they have to be surrounded by parentheses, so the line below would be 

```ruby
if (fingerprint = json["fingerprint"])
```

That should not set off the cop.

BlakeWilliams: Weird! I'll try that, thanks.

GordonDiggs:
:+1:

BlakeWilliams:
Thanks for the review!

PR #209, opened 2015-11-16
Replace recursive directory class with minimizer

BlakeWilliams:
This replaces the existing Directory class with a minimizer class which
takes the paths to check against and diffs it against all the files in
project to get directories where possible instead of a list of files.

This change is necessary to be able to use engine specific excludes
without having to build up the include_paths over and over for each
engine.

### DIFF
--- a/lib/cc/analyzer/include_paths_builder.rb
+++ b/lib/cc/analyzer/include_paths_builder.rb
@@ -1,13 +1,10 @@
 require "file_utils_ext"
+require "cc/analyzer/path_minimizer"
 
 module CC
   module Analyzer
     class IncludePathsBuilder
-      def self.relevant_entries(path)
-        Dir.entries(path).reject do |e|
-          %w(. .. .git).include?(e) || File.symlink?(File.join(path, e))
-        end
-      end
+      IGNORE_PATHS = [".", "..", ".git"].freeze
 
       attr_reader :cc_include_paths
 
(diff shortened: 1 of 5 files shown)


spec/cc/analyzer/include_paths_builder_spec.rb
```diff
@@ -42,7 +42,7 @@ module CC::Analyzer
       end
 
       it "keeps sibling files and directories in include_paths" do
-        %w[subdir/ trackable.rb].each do |trackable_file_or_directory|
+        %w[subdir trackable.rb].each do |trackable_file_or_directory|
```
fhwang: Trailing slashes for directories is part of the spec. It's meant to make it a little easier for engine implementors to find directories:

https://github.com/codeclimate/spec/blob/master/SPEC.md (under "Which files to analyze")

So they should probably be kept in unless there's a strong enough reason to force all the engines to update.

BlakeWilliams: Gotcha, wasn't sure if there was a reason for that or not. I'll change it back.

The engine should probably check for itself too though.

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -17,152 +14,76 @@ def initialize(cc_exclude_paths, cc_include_paths = [])
...
+        @_build ||= PathMinimizer.new(paths).minimize
+      end
 
-        paths.each do |path|
-          raise_on_unreadable_files(path)
-        end
+      def build_with_excludes(excludes)
+        new_paths = paths.reject { |path| matches_globs?(path, excludes) }
+        PathMinimizer.new(new_paths).minimize
       end
```
fhwang: I think having both a `build` and `build_with_excludes` sort of muddies the intention of this class. And, in fact, you could argue that `build` here is just `build_with_excludes` with an empty excludes array. What do you think about merging the two?

spec/cc/analyzer/engines_runner_spec.rb
```diff
@@ -19,7 +19,7 @@ module CC::Analyzer
       formatter = null_formatter
 
       expect_engine_run("an_engine", "/code", formatter)
-      FileUtils.stubs(:readable_by_all?).at_least_once.returns(true)
+      # FileUtils.stubs(:readable_by_all?).at_least_once.returns(true)
```
fhwang: Looks like you forgot to delete this line.

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -17,152 +14,76 @@ def initialize(cc_exclude_paths, cc_include_paths = [])
...
 
-      def raise_on_unreadable_files(path)
-        if File.directory?(path)
-          raise_on_unreadable_files_in_directory(path)
-        elsif !FileUtils.readable_by_all?(path)
-          raise CC::Analyzer::UnreadableFileError, "Can't read #{path}"
+      def include_paths
+        if @cc_include_paths.empty?
+          Dir.glob("**/*", File::FNM_DOTMATCH)
```
joshuaclayton: Does this coincide with the logic in `glob_or_include_path`? It'd be neat if we could do something like:

```ruby
def include_paths
  cc_include_paths.map do |path|
    glob_or_include_path(path)
  end.flatten
end

def cc_include_paths
  if @cc_include_paths.empty?
    @cc_include_paths += Dir.glob("**/*", File::FNM_DOTMATCH)
  end
  @cc_include_paths
end
```

lib/cc/analyzer/path_minimizer.rb
```diff
@@ -0,0 +1,78 @@
...
+      def diff
+        @_diff ||=
+          (all_files - paths).
+          reject { |path| File.symlink?(path) }.
+          map { |path| build_entry_combinations(path) }.
+          flatten
+      end
+
+      def filtered_paths
+        @_filtered_pathss ||=
```
joshuaclayton: `@_filtered_paths`?

spec/cc/analyzer/engines_runner_spec.rb
joshuaclayton: should this be removed?

lib/cc/analyzer/include_paths_builder.rb
BlakeWilliams: I started looking into it but I don't think we'd want to call `glob_or_include_path` if `@cc_include_paths` is empty because `Dir.glob("**/*")` already includes all files / folders.

The case where we need `glob_or_include_path` is when the class is given `@cc_include_paths` that have directories since we need to work on the files in the directory at this stage.

lib/cc/analyzer/include_paths_builder.rb
BlakeWilliams: Fair enough. One reason I avoided it was so that it wouldn't have to loop over the paths another time.

I'll change it to just `build` with an optional `extra_excludes` param.

spec/cc/analyzer/engines_runner_spec.rb
BlakeWilliams: Yep!

BlakeWilliams: :+1:

BlakeWilliams:
@fhwang Changed the public API to be just `build`

lib/cc/analyzer/include_paths_builder.rb
```diff
@@ -16,153 +13,85 @@ def initialize(cc_exclude_paths, cc_include_paths = [])
...
-        Tempfile.open(".cc_gitignore") do |tmp|
-          tmp.write(File.read(".gitignore")) if File.file?(".gitignore")
-          tmp << @cc_exclude_paths.join("\n")
-          tmp.close
-          tracked_and_ignored = `git ls-files -zi -X #{tmp.path} 2>/dev/null`.split("\0")
-          untracked_and_ignored = `git ls-files -zio -X #{tmp.path} 2>/dev/null`.split("\0")
-          tracked_and_ignored + untracked_and_ignored
-        end
+      def build_without_extra_excludes
+        @_build_without_extra_excludes ||= PathMinimizer.new(paths).minimize
```
fhwang: I think our house style is not to put leading underscores on ivars like this. i.e. just `@build_without_extra_excludes`

fhwang:
Other than one style nit, LGTM!

lib/cc/analyzer/include_paths_builder.rb
BlakeWilliams: Even if they're not meant to be used directly?

BlakeWilliams: Talked about this in Slack, keeping for now.

BlakeWilliams:
Thanks for the reviews!

PR #210, opened 2015-11-16
green output

ABaldwinHunter:
This PR has two changes to the `codeclimate init` command config generation:

*minor*
1) make successful output messaging green

*major*
2) generate default configs if possible, even if `.codeclimate.yml` already exists. Keep same error / exit code 1 behavior. 

Since it might still be useful for users with a `.codeclimate.yml` to get default configs by running
`codeclimate init`, this change generates applicable configs even when
.codeclimate.yml present, before finishing with the same warning message and
exit code 1:

```
$ codeclimate init
Config file .csslintrc successfully generated.
Skipping generating coffeelint.json file (already exists).
Config file .eslintrc successfully generated.
Skipping generating .rubocop.yml file (already exists).
Config file .codeclimate.yml already present.
Try running 'validate-config' to check configuration.
```

cc @codeclimate/review

### DIFF
--- a/lib/cc/cli/command.rb
+++ b/lib/cc/cli/command.rb
@@ -24,7 +24,7 @@ module CC
         run
       end
 
-      def congrats(message)
+      def success(message)
         terminal.say colorize(message, :green)
       end
 
--- a/lib/cc/cli/command.rb
+++ b/lib/cc/cli/command.rb
@@ -24,6 +24,10 @@ module CC
         run
       end
 
+      def congrats(message)
+        terminal.say colorize(message, :green)
+      end
+
       def say(message)
         terminal.say message
       end
--- a/lib/cc/cli/help.rb
+++ b/lib/cc/cli/help.rb
@@ -14,7 +14,7 @@ module CC
 
       def commands
         [
-          "analyze [-f format]",
+          "analyze [-f format] [-e engine] <path>",
           "console",
           "engines:disable #{underline('engine_name')}",
           "engines:enable #{underline('engine_name')}",
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -9,8 +9,8 @@ module CC
 
       def run
         if !upgrade? && filesystem.exist?(CODECLIMATE_YAML)
+          warn "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
           create_default_configs
-          fatal "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
         elsif upgrade? && engines_enabled?
           fatal "--upgrade should not be used on a .codeclimate.yml configured for the Platform.\nTry running 'validate-config' to check configuration."
         else
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -33,7 +33,7 @@ module CC
         end
 
         create_codeclimate_yaml
-        say "Config file .codeclimate.yml successfully #{config_generator.post_generation_verb}.\nEdit and then try running 'validate-config' to check configuration."
+        say colorize("Config file .codeclimate.yml successfully #{config_generator.post_generation_verb}.\nEdit and then try running 'validate-config' to check configuration.", :green)
         create_default_configs
       end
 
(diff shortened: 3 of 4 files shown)


lib/cc/cli/init.rb
```diff
@@ -9,6 +9,7 @@ class Init < Command
 
       def run
         if !upgrade? && filesystem.exist?(CODECLIMATE_YAML)
+          create_default_configs
           fatal "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
```
GordonDiggs: It seems odd to have a fatal output if we successfully created default configs. Maybe the message should mention that we generated them?

GordonDiggs: EDIT: nvm - I see the output of that now

GordonDiggs:
LGTM

lib/cc/cli/init.rb
ABaldwinHunter: Yeah, I had qualms about the final exitcode as well.

wfleming: Yeah, if this default configs is the desired behavior, I think it should be a successful output exit code & message. I would suggest outputting the informational message about not changing `.codeclimate.yml` first, and then generate the default configs.

This can & probably should actually be refactored to be part of `#create_codeclimate_yaml` behavior. Putting it here make behavior kind of strange in that `codeclimate init` will not touch `.codeclimate.yml`, but will touch default engine configs, whereas `codeclimate init --upgrade` will refuse to generate default engine configs if `.codeclimate.yml` already has engines.

ABaldwinHunter: @wfleming cool. Thanks for the explanation. I'd like consistently generating a successful exit code.

What was our original reason in favor of going with the `exitcode 1` when there's no `.codeclimate.yml`? What would you think about something more neutral like `.codeclimate.yml already present, skipping generation`, like we do for each engine config that's already present?

wfleming: > What was our original reason in favor of going with the exitcode 1 when there's no .codeclimate.yml

The logic predated me, but it was failing with exit code 1 if `.codeclimate.yml` **already existed** because `init`'s primary purpose was to create that file.

> What would you think about something more neutral like .codeclimate.yml already present, skipping generation

I think that sounds fine (`#warn` seems appropriate for the output). The verb "generation" would not be appropriate for the `--upgrade` case, though: it seems like you'll need another `*_verb` method on the generator class to distinguish between the generated/upgraded.

ABaldwinHunter:
@codeclimate/review @wfleming Ready for re-review.

I updated the PR to return a warning that `.codeclimate.yml` is not found, instead of error.

I left the behavior of [init --upgrade](https://github.com/codeclimate/codeclimate/pull/210/files#diff-ebd3ae72c4e4074aa627b08e8ed743bbL13) the same because it seems separate enough, and perhaps desirable as is: at the moment, `codeclimate init` doesn't get called on .com unless the `.codeclimate.yml` file is missing or contains engines, so this change shouldn't have an impact there.

Let me know what you think!

lib/cc/cli/help.rb
```diff
@@ -14,7 +14,7 @@ def run
 
       def commands
         [
-          "analyze [-f format]",
+          "analyze [-f format] [-e engine] <path>",
```
GordonDiggs: I don't think it needs to be removed, but want to mention that this seems like a drive-by fix

lib/cc/cli/command.rb
```diff
@@ -24,6 +24,10 @@ def execute
         run
       end
 
+      def congrats(message)
```
GordonDiggs: WDYT of `success`?

ABaldwinHunter: k. I like that

lib/cc/cli/help.rb
ABaldwinHunter: yeah, agreed.

At first I was curious about whether we have `--upgrade` in the help commands, and noticed this absence as well.

ABaldwinHunter:
@GordonDiggs @wfleming fixed!

GordonDiggs:
LGTM

PR #211, opened 2015-11-16
Add engine specific exclude_paths

BlakeWilliams:
This change allows engines to have their own configured `exclude_paths`
which are merged with the global exclude paths.

For example, you can skip all spec files for just Rubocop:

```
rubocop:
  enabled: true
  exclude_paths:
  - **/*spec.rb
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     codeclimate (0.14.2)
       activesupport (~> 4.2, >= 4.2.1)
-      codeclimate-yaml (~> 0.6.1)
+      codeclimate-yaml (~> 0.7.0)
       faraday (~> 0.9.1)
       faraday_middleware (~> 0.9.1)
       highline (~> 1.7, >= 1.7.2)
@@ -16,7 +16,7 @@ PATH
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (4.2.4)
+    activesupport (4.2.5)
       i18n (~> 0.7)
       json (~> 1.7, >= 1.7.7)
       minitest (~> 5.1)
@@ -26,7 +26,7 @@ GEM
     builder (3.2.2)
     codeclimate-test-reporter (0.4.8)
       simplecov (>= 0.7.1, < 1.0.0)
-    codeclimate-yaml (0.6.1)
+    codeclimate-yaml (0.7.0)
       activesupport
       secure_string
     coderay (1.1.0)
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -29,12 +29,13 @@ module CC
 
       private
 
-      attr_reader :include_paths
-
       def engine_config(raw_engine_config)
+        engine_excludes = Array(raw_engine_config.delete("exclude_paths"))
+        excludes = exclude_paths(engine_excludes)
+
         config = raw_engine_config.merge(
-          exclude_paths: exclude_paths,
-          include_paths: include_paths,
+          exclude_paths: excludes,
+          include_paths: include_paths(excludes),
         )
         # The yaml gem turns a config file string into a hash, but engines
         # expect the string. So we (for now) need to turn it into a string in
@@ -56,13 +57,19 @@ module CC
         end
       end
 
-      def include_paths
-        IncludePathsBuilder.new(exclude_paths, Array(@requested_paths)).build
+      def include_paths(excludes)
+        IncludePathsBuilder.new(
+          excludes,
+          Array(@requested_paths)
+        ).build
       end
 
-      def exclude_paths
-        PathPatterns.new(@config.exclude_paths || []).expanded +
-          gitignore_paths
+      def exclude_paths(engine_excludes)
+        global_excludes + engine_excludes
+      end
+
+      def global_excludes
+        PathPatterns.new(@config.exclude_paths || []).expanded + gitignore_paths
       end
 
       def gitignore_paths
--- a/lib/cc/analyzer/path_patterns.rb
+++ b/lib/cc/analyzer/path_patterns.rb
@@ -32,7 +32,7 @@ module CC
         # are translated to **/*-style globs within cc-yaml's custom
         # Glob#value method. It was thought that that would work correctly
         # with Dir.glob but it turns out we have to actually invoke #value
-        # directrly for this to work. We need to guard this on class (not
+        # directly for this to work. We need to guard this on class (not
         # respond_to?) because our mocking framework adds a #value method to
         # all objects, apparently.
         if pattern.is_a?(CC::Yaml::Nodes::Glob)
--- a/spec/cc/analyzer/include_paths_builder_spec.rb
+++ b/spec/cc/analyzer/include_paths_builder_spec.rb
@@ -15,7 +15,7 @@ module CC::Analyzer
 
     before do
       system("git init > /dev/null")
-      FileUtils.stubs(:readable_by_all?).at_least_once.returns(true)
+      FileUtils.stubs(:readable_by_all?).returns(true)
     end
 
     describe "when the source directory contains only files that are tracked or trackable in Git" do
@@ -248,6 +248,17 @@ module CC::Analyzer
       it "skips it entirely" do
         FileUtils.readable_by_all?(".")
         result.include?("./").must_equal(true)
+      end
+    end
+
+    describe "when analyzing a single file" do
+      let(:cc_includes) { ["subdir/subdir_file.rb"] }
+
+      it "returns the file" do
+        make_file("root_file.rb")
+        make_file("subdir/subdir_file.rb")
+
+        builder.build.must_equal(["subdir/subdir_file.rb"])
       end
     end
 
--- a/codeclimate.gemspec
+++ b/codeclimate.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency "activesupport", "~> 4.2", ">= 4.2.1"
   s.add_dependency "tty-spinner", "~> 0.1.0"
-  s.add_dependency "codeclimate-yaml", "~> 0.6.1"
+  s.add_dependency "codeclimate-yaml", "~> 0.7.0"
   s.add_dependency "faraday", "~> 0.9.1"
   s.add_dependency "faraday_middleware", "~> 0.9.1"
   s.add_dependency "highline", "~> 1.7", ">= 1.7.2"
--- a/spec/cc/analyzer/engines_config_builder_spec.rb
+++ b/spec/cc/analyzer/engines_config_builder_spec.rb
@@ -139,6 +139,45 @@ module CC::Analyzer
       end
     end
 
+    describe "when the engine specifies its own excludes" do
+      let(:config) do
+        CC::Yaml.parse <<-EOYAML
+          engines:
+            rubocop:
+              enabled: true
+              exclude_paths:
+              - subdir/*
+          exclude_paths:
+          - "ignored.rb"
+        EOYAML
+      end
+      let(:registry) { registry_with_engine("rubocop") }
+
+      before do
+        make_file("ignored.rb")
+        make_file("root_file.rb")
+        make_file("subdir/subdir_file.rb")
+      end
+
+      it "excludes engine specific paths and passes combined exclude paths to engine" do
+        IncludePathsBuilder.
+          stubs(:new).with(["ignored.rb", "subdir/*"], []).
+          returns(mock(build: ["."]))
+        expected_config = {
+          "enabled" => true,
+          exclude_paths: ["ignored.rb", "subdir/*"],
+          include_paths: ['.']
+        }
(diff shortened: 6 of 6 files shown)


lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -11,11 +11,11 @@ class EnginesConfigBuilder
         :container_label,
       )
 
-      def initialize(registry:, config:, container_label:, source_dir:, include_paths:)
+      def initialize(registry:, config:, container_label:, source_dir:, include_paths_builder:)
```
fhwang: Yeah, my vote would be that we keep this responsibility inside `EnginesConfigBuilder`. If we're genuinely seeing performance issues from this, there are ways to have `IncludePathsBuilder` cache its own results -- class variables, singletons, etc.

For context: `builder` doesn't even use `EnginesRunner` any more, it uses a `builder`-specific class. So pushing this responsibility our of `EnginesConfigBuilder` is going to lead to some cross-repo duplication.

BlakeWilliams: Gotcha, I wasn't aware that it wasn't using EnginesRunner. I'll work on moving this out.

BlakeWilliams: Also, IncludePathsBuilder does cache its own results, which is why we pass down the instance to EnginesConfigBuilder due to having the global excludes and then engine specific excludes.

fhwang: Right, I meant something on the class-level instead of the instance-level.

BlakeWilliams: Ah, gotcha, missed the singleton part.

Would it be hard to do this in builder? I think all you need to do is pass in:

```ruby
IncludePathsBuilder.new(
  @config.exclude_paths || [], @requested_paths
)
```

If not this might be worth moving to a singleton if that works for builder.

fhwang: It wouldn't be hard to do as a one-off. It just increases coupling between codeclimate and builder if builder has to be aware of one more class, and how to instantiate it. We've already had a lot of slowness forced on us by how entangled these repos are, so we should be looking to narrow touchpoints as much as possible.

BlakeWilliams: Right, I'm 100% unaware of builder and how it works so I want to make sure whatever change I make here is going to be compatible / what we want to do while also keeping the startup time of each engine down in the CLI.

BlakeWilliams:
@fhwang Made `include_paths_builder` optional now.

fhwang:
Can you also support the `requested_paths` key that Builder uses?

BlakeWilliams:
Yep, turns out some other requirement need it too. So adding that in. :smiley:

ABaldwinHunter:
@BlakeWilliams can you fix the failing CI tests and also run `codeclimate analyze` under a few scenarios with the new docker image locally if you haven't already?

Gemfile.lock
```diff
@@ -87,6 +87,3 @@ DEPENDENCIES
   mocha
   rack-test
   rake
-
-BUNDLED WITH
-   1.10.6
```
ABaldwinHunter: What's our stance on keeping these in diffs?

BlakeWilliams: No idea, I didn't remove this manually.

BlakeWilliams:
Oddly enough it's passing locally. I'll look into that.

I tried it out locally giving it several files, single file, and directories and it seemed to work as intended.

Gemfile.lock
pbrisbin: Relatively recent policy on this:

- Everyone should be using latest bundler on all projects
- That is currently 1.10.6, so doing so will prevent this diff from cropping up
- If/when bundler does update, we'll see some noise here once per project

So... please `git reset` this against master, then `gem install bundler`, then you shouldn't see it vanish again on your next `bundle` invocation.

lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -29,12 +29,23 @@ def run
...
 
-      attr_reader :include_paths
+      attr_reader :include_paths_builder
+
+      def include_paths(engine_excludes)
+        if @requested_paths.empty?
+          IncludePathsBuilder.new(global_excludes).build(engine_excludes)
+        else
+          paths = IncludePathsBuilder.new([], @requested_paths).build
+          paths
```
pbrisbin: I'd probably ditch this local

BlakeWilliams: Good catch, was left over from debugging.

lib/cc/analyzer/path_filter.rb
```diff
@@ -0,0 +1,65 @@
...
+        self
+      end
+
+      def reject_unreadable_paths
+        @paths = paths.reject { |path| unreadable_path_entries.include?(path) }
+        self
+      end
+
+      def reject_paths(ignore_paths)
+        @paths = paths.reject { |path| ignore_paths.include?(path) }
```
pbrisbin: Is this equivalent to `paths - ignore_paths`?

lib/cc/analyzer/path_filter.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      def unreadable_paths
+        paths.select do |path|
+          File.directory?(path) && !FileUtils.readable_by_all?(path)
+        end
+      end
+
+      def matches_globs?(path, globs)
+        globs.any? do |exclude_path|
+          File.fnmatch(exclude_path, path)
```
pbrisbin: I just want to highlight that we've had a number of subtle bugs around some confusing behavior in Ruby.

Imagine the following structure:

```
foo.rb
foo/
  bar.rb
```

`Dir.glob("**/*.rb")` will find `foo.rb` (as expected). However, `File.fnmatch?("**/*.rb", "foo.rb")` is `false`.

This is why we (for a time) had to tell users they needed `**/*.rb` in excludes to "exclude all ruby files", but `**.rb` in ratings-paths to "rate all ruby file" -- confusing, and dumb.

So, BE CAREFUL. Where do `globs` come from? User input? Do you expect to see `**.rb` or `**/*.rb`? Is there a way to use the `PathPatterns` class, which we created in an effort to standardize on one kind of behavior in all our glob handling?

lib/cc/analyzer/path_minimizer.rb
```diff
@@ -34,13 +36,10 @@ def filtered_paths
...
       def filter_path?(path)
-        dir = File.dirname(path)
-
-        in_diff = in_diff?(path) || in_diff?(dir)
-        is_nested = nested?(path) || nested?(dir)
+        in_diff = in_diff?(path)
         is_ignored = IncludePathsBuilder::IGNORE_PATHS.include?(path)
 
-        in_diff || is_nested || is_ignored
+        in_diff || is_ignored
```
pbrisbin: Now that these stanzas are simpler, can we ditch the locals?

```rb
in_diff?(path) || IncludePathsBuilder::IGNORE_PATHS.include?(path)
```

BlakeWilliams: :+1:

spec/cc/analyzer/include_paths_builder_spec.rb
```diff
@@ -261,5 +261,18 @@ module CC::Analyzer
...
       end
     end
+
+    describe "when analyzing a single file" do
+      let(:cc_includes) { ["subdir/subdir_file.rb"] }
+
+      before do
+        make_file("root_file.rb")
+        make_file("subdir/subdir_file.rb")
+      end
```
pbrisbin: Thoughts on co-locating the `let` and `before` into the (only) example using them?

BlakeWilliams: I can extract the before block, but I think the `let` has to stay since it's used waaaaay above.

lib/cc/analyzer/path_filter.rb
BlakeWilliams: I think these are the `exclude_paths` that users give us.

I'll look into PathPatterns and see if that works here.

spec/cc/analyzer/include_paths_builder_spec.rb
pbrisbin: Oh, this is worse than I thought ;) -- `cc_includes` is somehow magic and used by something not visible within this `describe` or `it`?

Oh well, backing away...

BlakeWilliams: Yeah.... Just gonna say that it wasn't me. :stuck_out_tongue:

lib/cc/analyzer/path_patterns.rb
```diff
@@ -10,30 +10,40 @@ def expanded
         @expanded ||= expand
       end
 
+      def match?(path)
+        @patterns.any? do |pattern|
+          value = glob_value(pattern)
+          File.fnmatch(pattern, path)
```
BlakeWilliams: @pbrisbin I replaced my method for matching globs with this.

spec/cc/analyzer/path_entries_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    describe "#entries" do
+      it "filters out trailing slashes" do
+        make_file("lib/foo.rb")
+
+        paths = PathEntries.new("lib/").entries
+
+        paths.sort.must_equal(["lib", "lib/foo.rb"])
+      end
```
ABaldwinHunter: Nice! I like this test. Can we also have a test that checks what happens when passed dir with no slash (e.g. `lib`), and multiple paths `lib test.rb app/models`?

lib/cc/analyzer/path_patterns.rb
pbrisbin: Sorry, I'm not sure if that answers my concerns.

This implementation will require `pattern` to be (.e.g) `**.rb` (not `**/*.rb`) to match a top-level files like `foo.rb` -- is that the behavior you want / need? It's usually not, but it may be; just want to be sure.

lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -29,12 +29,22 @@ def run
...
       private
 
-      attr_reader :include_paths
+      attr_reader :include_paths_builder
+
+      def include_paths(engine_excludes)
+        if @requested_paths.empty?
+          IncludePathsBuilder.new(global_excludes).build(engine_excludes)
+        else
+          IncludePathsBuilder.new([], @requested_paths).build
```
ABaldwinHunter: I might be missing the complete flow of things, but aren't we going to honor `excludes` from `.codeclimate.yml` unless `--ignore-excludes` gets set? and if so, would we do that here?

BlakeWilliams: Oops! This was left over from when we were discussing it and then I got sidetracked. Adding it back in.

spec/cc/analyzer/path_entries_spec.rb
BlakeWilliams: The class doesn't accept multiple paths, it just generates entries for a given path. I'll add a test for no slash though.

lib/cc/analyzer/path_patterns.rb
pbrisbin: OK, I think I'm tracking better now...

Assuming `@patterns` are coming in as `cc/yaml` `Glob` instances, then you're right in that my point number 1 above is being followed (thanks to `#glob_value`). You should only ever be seeing `**/*.rb` here.

However, point number 2 is not. By using `File.fnmatch?` you're going to return `false` here for `path = "foo.rb"`. I suggest you use `PathPatterns`.

BlakeWilliams: Gotcha! That's kind of a tricky one. I'll use `expand` instead.

BlakeWilliams:
cc @codeclimate/review

pbrisbin:
This LGTM.

The CC issue is annoying. I want trailing commas in arrays and hashes, and hash arguments to method calls, but I dislike the trailing comma in *all* method calls. I don't think rubocop can be tuned that finely. Maybe I/we should just get used to the trailing comma everywhere?

ABaldwinHunter:
@BlakeWilliams as chatted about in slack, my concern is performance issues and I think it would be smart to cache some results of  one walk:
```
safe_dirs = IncludePathsBuilder.new([codeclimate.yml_excludes + all_engine_excludes]).build

IncludePathsBuilder::SAFE_DIRS = _

```

then for each engine

IncludePathsBuilder.build(includes, excludes)

will have an extra step

```
if SAFE_DIRS.include?(path)
  engine_include_paths << path
else
   drill down
end

```

ABaldwinHunter:
Also as discussed, not sure if we'll abandon drilling down into each directory, in which case this performance difference would be negligible.

But see what other peeps think.

jpignata:
@codeclimate/review This a month old and our priorities shifted away from it. My preference is to close this and reference it later when we move back to it. Any objections?

jpignata:
Closing for now.

PR #212, opened 2015-11-17
add init f

ABaldwinHunter:
This change allows users to overwrite a `.codeclimate.yml` file if present.

For the time being, it will not overwrite default engine configs if they are already
present -e.g. .rubocop.yml, with the thought that users might be more likely to have special configurations
there.

This PR also adds options processing to the `Command` class (overwritten by `Analyze`), and throws an error when unrecognized parameters get passed:
```
Ashleys-MacBook-Pro-2:nillson ashleybaldwin-hunter$ codeclimate init --x
error: (CC::CLI::Command::InvalidOption) option --x not found. Try running codeclimate --help for a list of valid commands.
```

### DIFF
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -34,12 +34,12 @@ module CC
       def process_args
         while (arg = @args.shift)
           case arg
-          when "-f"
-            @formatter = Formatters.resolve(@args.shift).new(filesystem)
-          when "-e", "--engine"
-            @engine_options << @args.shift
           when "--dev"
             @dev_mode = true
+          when "-e", "--engine"
+            @engine_options << @args.shift
+          when "-f"
+            @formatter = Formatters.resolve(@args.shift).new(filesystem)
           else
             @path_options << arg
           end
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -8,7 +8,7 @@ module CC
       include CC::Analyzer
 
       def run
-        if !upgrade? && filesystem.exist?(CODECLIMATE_YAML)
+        if !upgrade? && filesystem.exist?(CODECLIMATE_YAML) && !force
           warn "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
           create_default_configs
         elsif upgrade? && engines_enabled?
--- a/lib/cc/cli/command.rb
+++ b/lib/cc/cli/command.rb
@@ -7,9 +7,11 @@ module CC
   module CLI
     class Command
       CODECLIMATE_YAML = ".codeclimate.yml".freeze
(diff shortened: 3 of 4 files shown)


lib/cc/cli/analyze.rb
```diff
@@ -34,12 +34,12 @@ def run
...
-            @formatter = Formatters.resolve(@args.shift).new(filesystem)
-          when "-e", "--engine"
-            @engine_options << @args.shift
           when "--dev"
             @dev_mode = true
+          when "-e", "--engine"
+            @engine_options << @args.shift
+          when "-f"
+            @formatter = Formatters.resolve(@args.shift).new(filesystem)
           else
```
ABaldwinHunter: Alphabetize these to pave the way for additional options

fhwang:
There's probably some context I'm missing: What's the use-case here? Why not just tell users to edit `.codeclimate.yml` directly or delete it?

lib/cc/cli/init.rb
```diff
@@ -8,7 +8,7 @@ class Init < Command
       include CC::Analyzer
 
       def run
-        if !upgrade? && filesystem.exist?(CODECLIMATE_YAML)
+        if !upgrade? && filesystem.exist?(CODECLIMATE_YAML) && !force
```
fhwang: This is some fairly confusing branching behavior. And I notice that `create_default_configs` is called in two places, here in `run` and lower at the end of `generate_config`. You might be able to make this clearer if:

* `create_default_config` were pulled out of `generate_config` and put somewhere inside of `run`, to more clearly communicate the fact that it's going to get called whether or not `.codeclimate.yml` already exists
* `create_default_config` were renamed to indicate we're talking about engine-specific configs

wfleming:
Having `--force` behavior apply specifically to one file this command touches but not other is surprising behavior, I think, from an end user perspective. I'm also wondering what use case drives this, but if it's needed I think it should behave the way `--force` does on pretty much every other UNIX utility and force *everything*. As Francis said, deleting the .codeclimate.yml is easy if that's the only file you want overwritten.

ABaldwinHunter:
@fhwang @wfleming thanks for the feedback! I'm halting this branch in favor of engines QA at the moment, but can make the suggested changes later.

> what's driving this change?

I actually just found a [card](https://trello.com/c/406lyKnf/38-codeclimate-init-force) for it in trello backlog of Engines board and - since I just worked on the `init` command, figured I would take a crack at it.

wfleming:
That card is really old. My guess is that the use-case @brynary had in mind for this was what we ended up addressing with `--upgrade`, and that this card should be dropped. Bryan, can you chime in if you had other cases in mind?

brynary:
Fine to close this 



Typed with thumbs.

ABaldwinHunter:
Okay sounds fine

PR #213, opened 2015-11-17
Fix brakeman ratings_paths

wfleming:
The value of `app/**` results in things like images in `assets` being rated as "A"s, which isn't appropriate, since nothing is really scanning those files.

This restricts the pattern more to files & extensions we know brakeman looks at.

cc @codeclimate/review

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -18,8 +18,12 @@ brakeman:
   enable_regexps:
     - ^script\/rails$
   default_ratings_paths:
-    - "app/**"
+    - "Gemfile.lock"
+    - "**.erb"
+    - "**.haml"
     - "**.rb"
+    - "**.rhtml"
+    - "**.slim"
 bundler-audit:
   image: codeclimate/codeclimate-bundler-audit
   description: Patch-level verification for Bundler.


config/engines.yml
```diff
@@ -18,7 +18,10 @@ brakeman:
   enable_regexps:
     - ^script\/rails$
   default_ratings_paths:
-    - "app/**"
+    - "config.ru"
+    - "Gemfile.lock"
+    - "**.erb"
+    - "**.haml"
```
pbrisbin: I think we support slim too.

wfleming: Ah, neat, added.

wfleming:
I can't find any solid references in Brakeman docs about what they process, and the source doesn't lend itself to an easy check either. The list now added is the best reflection of brakeman's support I can determine, so I'm comfortable shipping it for now.

@ABaldwinHunter I removed `config.ru` since I can't actually find any mention of the file or the extension in the brakeman source. Can you point me at something indicating it is processed if you want it added back?

ABaldwinHunter:
@wfleming fine by me to remove it. 
:+1:  

I don't feel particularly confident that it's checked.

PR #216, opened 2015-11-19
Container: fix race with IO processing

wfleming:
There's a race between the main thread getting to this point after the
process finishes, and the two threads that process output from stdout &
stderr. When main thread "wins" that race, the result can be an
incorrect status for the events data on the engine step down the line:
e.g. the engine emits invalid output, but by the time the background
thread has tried to parse it & emitted an error, the main thread has
already decided the container is finished, so the error is never
attached to the relevant step.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -60,6 +60,7 @@ module CC
           duration = timeout
           @listener.timed_out(container_data(duration: duration))
         else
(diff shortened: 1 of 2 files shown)


GordonDiggs:
Nice find. LGTM

ABaldwinHunter:
wow. so subtle. nice

pbrisbin:
Would moving the calls to `#join` up from the ensure block have solved this
too?

PR #218, opened 2015-11-19
Only add default config if one exists for engine

ABaldwinHunter:
It is confusing to add a blank config key and results in errors
for certain engines, like coffeelint

https://github.com/codeclimate/codeclimate-coffeelint/pull/12

cc @codeclimate/review @fhwang

### DIFF
--- a/lib/cc/analyzer/config.rb
+++ b/lib/cc/analyzer/config.rb
@@ -35,11 +35,13 @@ module CC
         if engine_present?(engine_name)
           @config["engines"][engine_name]["enabled"] = true
         else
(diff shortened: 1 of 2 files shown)


GordonDiggs:
LGTM

fhwang:
Code-wise I think this looks good. Curious if others can think of a reason not to do this but if not let's do it

ABaldwinHunter:
Sounds good. I like it better than the original coffeelint change https://github.com/codeclimate/codeclimate-coffeelint/pull/12/files

PR #219, opened 2015-11-20
Don't raise on unreadable files when building paths: ignore them.

wfleming:
This is to address a regression
(https://trello.com/c/yQr7T7nH/1211-unreadable-files-lead-to-u10-without-error-visible)
that made it into builder from some recent changes here: these raises
were not caught anywhere, and the result on .com was a U10 without any
informative error text (because of where the exception ocurred).

Based on a recent conversation including Bryan, we'll probably expect
handling unreadable files to be an engine concern (if we want to avoid
walking the entire tree, we effectively *must* make it an engine
concern). This backs out this specific bit of logic to intentionally
*not* raise instead.

In my testing, this didn't actually change CLI behavior for the affected
case at all. In .com, analysis ran successfully with this fix, with some
engines emitting warnings on stderr about the file not being readable,
which seems like correct behavior to me. (Crashing the engine & emitting
an E10 would also have been acceptable, but warnings are even better.)

Long story short: this behavior broke on .com in an ugly way. Removing
it improves .com experience and doesn't seem to change CLI behavior in
any important way.

cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/include_paths_builder.rb
+++ b/lib/cc/analyzer/include_paths_builder.rb
@@ -24,7 +24,6 @@ module CC
         @_paths =
           PathFilter.new(include_paths).
           reject_paths(ignored_files).
-          raise_if_any_unreadable_files.
           reject_unreadable_paths.
           select_readable_files.
           reject_symlinks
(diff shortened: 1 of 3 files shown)


dblandin:
:+1:

ABaldwinHunter:
@wfleming Nice.

So the idea is to let engines raise or warn when they encounter unreadable files, instead of CLI, to get clearer error reporting?

That sounds good to me.
 
Perhaps we can drop the directory walk in the future, but I'm happy with little change first. LGTM.

PR #221, opened 2015-11-20
prep-release: generate a compare link

wfleming:
since we like to tag release PRs with compare links, I figured we may as
well automate that process where feasible.

Somebody with better git/terminal-fu than me can probably suggest a more effective way of getting the correct SHA for the release branch.

Thoughts, @codeclimate/review ?

### DIFF
--- a/bin/prep-release
+++ b/bin/prep-release
@@ -33,10 +33,11 @@ git add VERSION Gemfile.lock
 git commit -m "Release v$version"
 git push origin $branch
 
+compare_link="https://github.com/codeclimate/codeclimate/compare/${old_version}...$(git rev-parse --short $branch)"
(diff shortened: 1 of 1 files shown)


bin/prep-release
```diff
@@ -33,10 +33,11 @@ git add VERSION Gemfile.lock
 git commit -m "Release v$version"
 git push origin $branch
 
+compare_link="https://github.com/codeclimate/codeclimate/compare/${old_version}...$(git show-ref --hash=10 $branch | head -1)"
```
jpignata: Hmm, would `git rev-parse [tree-ish]` achieve the same thing?

wfleming: Good call. That is much more effective.

PR #222, opened 2015-11-20
bin-release: don't require version arg

wfleming:
since release is now a 2-step process, it seems silly to have to pass
the version to the release script since it'll be the value currently in
the VERSION file. If anything, there was a typo risk of typing the wrong
version & ending up with a mismatched release.

:eyes: @codeclimate/review

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -4,20 +4,15 @@
 #
 # Assumes bin/prep-release was run and the PR merged.
 #
-# Usage: bin/release VERSION
+# Usage: bin/release
 #
 ###
 set -e
 
-if [ -z "$1" ]; then
-  echo "usage: bin/release VERSION" >&2
-  exit 64
-fi
-
-version=$1
-
 git checkout master
 git pull
+
+version=$(< VERSION)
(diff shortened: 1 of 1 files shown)


bin/release
```diff
@@ -4,17 +4,12 @@
...
 
-if [ -z "$1" ]; then
-  echo "usage: bin/release VERSION" >&2
-  exit 64
-fi
-
-version=$1
+version=$(< VERSION)
 
 git checkout master
```
pbrisbin: The reason it worked this way, is that this script doesn't assume you've pulled master since the PR was merged -- that's done here.

Not saying what you're picturing won't or can't work, but it might need a few more changes to ensure running this script doesn't release the existing version.

pbrisbin: I guess the simplest thing is to just move the `version=` line to after the `checkout`/`pull`

wfleming: Oh, good point. So `co master; pull; version=...`

pbrisbin:
LGTM

PR #223, opened 2015-11-20
Pass test directories instead of test files to engines

dblandin:
Certain engines in the Code Climate Platform require analysis in take place within the context of a directory. One good example is the brakeman engine which expects the context of a Ruby on Rails application.

This commit updates the CLI test command to iterate over the engine specification's `test_paths`, extract all assertion markers within the directory, and pass the entire test directory to the engine.

Another side effect of this change is fewer shell outs to `codeclimate analyze` which should speed things up overall.

### DIFF
--- a/lib/cc/cli/test.rb
+++ b/lib/cc/cli/test.rb
@@ -106,16 +106,16 @@ module CC
       end
 
       def run_tests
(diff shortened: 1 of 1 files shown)


GordonDiggs:
LGTM

PR #225, opened 2015-11-20
improve io waiting, ensure timeout thread runs immediately

wfleming:
1. Yesterday's patch for IO wait worked, but this technique is more performant (and I'm deleting redundant code I missed yesterday, and adding a spec.)
2. Due to how the timeout thread was instantiated, it might not be scheduled soon, and so actual timeout could be well after what was desired. By running the thread immediately, this should improve this behavior.

This is working for me locally, and I have managed to make things timeout correctly locally. It should also be noted that timeout behavior was already specced as well, and those specs were running fine. So I think the missed timeout I saw in production this morning was something different, possibly an isolated problem, but almost certainly not a consequence of the other recent changes. This should tighten things up overall. I'll continue to dig into potential ways timeouts could be delayed or missed as well.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -57,7 +57,7 @@ module CC
         _, status = Process.waitpid2(pid)
 
         if @timed_out
-          duration = timeout
+          duration = timeout * 1000
           @listener.timed_out(container_data(duration: duration))
         else
           sleep 1 until !t_out.alive? && !t_err.alive?
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -74,9 +74,7 @@ module CC
           @stderr_io.string,
         )
       ensure
-        t_timeout.kill if t_timeout
-        t_out.kill if t_out
-        t_err.kill if t_err
+        [t_timeout, t_out, t_err].each { |t| t.kill if t }
       end
 
       def stop
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -60,7 +60,7 @@ module CC
           duration = timeout
           @listener.timed_out(container_data(duration: duration))
         else
-          sleep 0.01 until !t_out.alive? && !t_err.alive?
+          t_out.join || t_err.join
           duration = ((Time.now - started) * 1000).round
           @listener.finished(container_data(duration: duration, status: status))
         end
@@ -78,9 +78,6 @@ module CC
         if @timed_out
           t_out.kill if t_out
           t_err.kill if t_err
-        else
-          t_out.join if t_out
-          t_err.join if t_err
         end
       end
 
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -60,7 +60,7 @@ module CC
           duration = timeout
           @listener.timed_out(container_data(duration: duration))
         else
-          t_out.join || t_err.join
+          t_out.join && t_err.join
           duration = ((Time.now - started) * 1000).round
           @listener.finished(container_data(duration: duration, status: status))
         end
@@ -75,10 +75,8 @@ module CC
         )
       ensure
         t_timeout.kill if t_timeout
-        if @timed_out
-          t_out.kill if t_out
-          t_err.kill if t_err
-        end
+        t_out.kill if t_out
+        t_err.kill if t_err
       end
 
       def stop
@@ -128,7 +126,7 @@ module CC
           sleep timeout
           @timed_out = true
           reap_running_container
-        end
+        end.run
       end
 
       def check_output_bytes(last_read_byte_count)
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -60,7 +60,7 @@ module CC
           duration = timeout
           @listener.timed_out(container_data(duration: duration))
         else
-          t_out.join && t_err.join
+          sleep 1 until !t_out.alive? && !t_err.alive?
           duration = ((Time.now - started) * 1000).round
           @listener.finished(container_data(duration: duration, status: status))
         end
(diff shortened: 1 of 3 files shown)


lib/cc/analyzer/container.rb
```diff
@@ -60,7 +60,7 @@ def run(options = [])
           duration = timeout
           @listener.timed_out(container_data(duration: duration))
         else
-          sleep 0.01 until !t_out.alive? && !t_err.alive?
+          t_out.join || t_err.join
```
GordonDiggs: Is it possible for one of these to finish before the other? Imagining if stderr closes but stdout is still flushing? Is that possible?

pbrisbin: I think it'd be safest / simplest to just join both

lib/cc/analyzer/container.rb
```diff
@@ -131,7 +128,7 @@ def timeout_thread
           sleep timeout
           @timed_out = true
           reap_running_container
-        end
+        end.run
```
pbrisbin: What's this little guy?

pbrisbin: I see, should've read the commit

lib/cc/analyzer/container.rb
```diff
@@ -78,9 +78,6 @@ def run(options = [])
         if @timed_out
           t_out.kill if t_out
           t_err.kill if t_err
```
pbrisbin: WDYT about bring this out from under the timed-out conditional? Presumably, if we're exiting this method (successfully or otherwise) we're done with these threads. If we were successful we'll have joined them already and the kill should be a no-op -- if not, better kill-than-sorry...

lib/cc/analyzer/container.rb
wfleming: Yep, you're right. That was my intent but I messed up my boolean logic.

lib/cc/analyzer/container.rb
wfleming: Seems reasonable to me.

pbrisbin:
This appears to now just remove the `Thread#join` calls -- what waits on the threads without them?

wfleming:
@pbrisbin Gordon just reverted to the `sleep` usage since it appeared to be more reliable, if slower. The `join` was causing problems for us. At this point the only changes here are really cleanup & a spec, which I'm fine with. I do think we need to look more closely at the threading behavior still, but the current state of the PR should be behavior neutral as far as that goes.

lib/cc/analyzer/container.rb
```diff
@@ -75,13 +75,8 @@ def run(options = [])
         )
       ensure
         t_timeout.kill if t_timeout
```
wfleming: :lipstick:, but looking at this now I think it would be nicer to read as `[t_timeout, t_out, t_err].each { |t| t.kill if t }`

GordonDiggs: I think doing that would be an issue if something raised an error before one of them was initialized (which is the purpose of the `if t_out` right?)

lib/cc/analyzer/container.rb
```diff
@@ -128,10 +123,18 @@ def read_stderr(err)
...
+          loop do
+            sleep 1
+            duration = Time.now - start_time
+            break if duration >= timeout
+          end
+
           @timed_out = true
           reap_running_container
-        end
+        end.run
```
ABaldwinHunter: Curious why this call was absent before but present now?

lib/cc/analyzer/container.rb
```diff
@@ -60,7 +60,7 @@ def run(options = [])
           duration = timeout
           @listener.timed_out(container_data(duration: duration))
         else
-          sleep 0.01 until !t_out.alive? && !t_err.alive?
+          sleep 0.2 until !t_out.alive? && !t_err.alive?
```
pbrisbin: I wondered this last time it came though: doesn't this basically just re-implement `Thread#join`?

lib/cc/analyzer/container.rb
```diff
@@ -128,10 +123,18 @@ def read_stderr(err)
...
+          # we do a series of short timeouts before exiting
+          start_time = Time.now
+          loop do
+            sleep 1
+            duration = Time.now - start_time
+            break if duration >= timeout
+          end
+
           @timed_out = true
           reap_running_container
```
dblandin: Should we keep the behavior between a timeout and too much memory by both calling `#stop` which stops the `on_output` behavior? I'm wondering if the timeout + `on_output` still handing output might be causing something to hang.

lib/cc/analyzer/container.rb
GordonDiggs: This is from @wfleming's original change - it's explained in the PR description

lib/cc/analyzer/container.rb
GordonDiggs: Interesting idea - I would like to ship this and revisit that if the problem isn't solved, but keeping this as small as possible seems like a good idea

GordonDiggs:
Would like support to mercy merge this as it could help fix the waiting status

lib/cc/analyzer/container.rb
```diff
@@ -57,10 +57,10 @@ def run(options = [])
         _, status = Process.waitpid2(pid)
 
         if @timed_out
-          duration = timeout
+          duration = timeout * 1000
           @listener.timed_out(container_data(duration: duration))
         else
-          sleep 0.01 until !t_out.alive? && !t_err.alive?
+          sleep 1 until !t_out.alive? && !t_err.alive?
```
pbrisbin: Comment lost to the diff:

I wondered this last time it came though: doesn't this basically just re-implement Thread#join?

GordonDiggs: @wfleming had a reason to back that change out - will, what was that?

lib/cc/analyzer/container.rb
wfleming: Ah, good point, but, that being the case, I'm not sure the existing code is correct either: `t_out.kill if t_out` will also raise an exception if `t_out` never got defined. We'd want to either use `defined?(t_tout)`, or switch to instance vars that have the nil-if-undefined behavior.

GordonDiggs: Hmm that might be right. I'll use the array and see what happens

lib/cc/analyzer/container.rb
ABaldwinHunter: :+1:  thanks

lib/cc/analyzer/container.rb
ABaldwinHunter: :+1:  thanks

PR #227, opened 2015-11-24
fix issue reported by codeclimate

mgracer48:
about Use || instead of or.

### DIFF
--- a/lib/cc/analyzer/formatters.rb
+++ b/lib/cc/analyzer/formatters.rb
@@ -12,7 +12,7 @@ module CC
       }.freeze
 
       def self.resolve(name)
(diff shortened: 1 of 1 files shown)


andyw8:
You'll need to wrap the arguments to `raise` in parentheses for this to be valid syntax.

(However, RuboCop's rule for this is somewhat controversial – see [this post](http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/) by Avdi Grimm for arguments in favour of using `and` and `or`).

mgracer48:
I don't know  this coding I'm just trying to fix issues that are like u don't need to know the code can u fix it up for me pls

GordonDiggs:
@mgracer48 Using `or` here is actually preferred, as we are using it for control flow and not logic operations (the post that @andyw8 shared explains that concept well).

PR #235, opened 2015-11-29
Disable Rails/Output check in inferred config

noahd1:
Default may make sense but the description of the issue is
confusing if your app isn't a rails app: "Do not write to stdout.
Use Rails' logger if you want to log"

/c @jpignata changed my mind after seeing the description


jpignata:
+1.

PR #237, opened 2015-12-01
close io pipes when threads exit

wfleming:
This seems like a possible cause of hanging builders after timeouts: we
aren't manually closing these IO pipes, and so they may be left open
when the process they're connected to is killed on some platforms? The
`pspawn` docs even say you should be sure to close these.

I haven't been able to reproduce the hanging builders locally, so this
is a bit of a shot in the dark.

:eyes: @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -60,7 +60,7 @@ module CC
           duration = timeout * 1000
           @listener.timed_out(container_data(duration: duration))
         else
-          sleep 1 until !t_out.alive? && !t_err.alive?
+          [t_out, t_err].each(&:join)
           duration = ((Time.now - started) * 1000).round
           @listener.finished(container_data(duration: duration, status: status))
         end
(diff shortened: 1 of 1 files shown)


lib/cc/analyzer/container.rb
```diff
@@ -101,20 +101,28 @@ def docker_run_command(options)
...
-            check_output_bytes(output.bytesize)
+          begin
+            out.each_line(@output_delimeter) do |chunk|
+              output = chunk.chomp(@output_delimeter)
+
+              @on_output.call(output)
+              check_output_bytes(output.bytesize)
+            end
+          ensure
+            io.close
```
pbrisbin: What is `io`, do you mean `out`?

dblandin:
Worth a shot :+1:

lib/cc/analyzer/container.rb
wfleming: I do. The fact that tests passed with that dumb typo is worrying. Hmmm.

wfleming: Oh, I think it was because those thread's aren't set to raise on exceptions.

wfleming:
@pbrisbin updated with the thread joining we discussed.

pbrisbin:
LGTM

PR #240, opened 2015-12-01
fix constructing multi-line pull request description

wfleming:
It appears `hub` doesn't like escaped chars in strings with `-m`. This
switches around to writing the message to a file and using that instead.

:eyes: @codeclimate/review

### DIFF
--- a/bin/prep-release
+++ b/bin/prep-release
@@ -34,12 +34,15 @@ git commit -m "Release v$version"
 git push origin $branch
 
 compare_link="https://github.com/codeclimate/codeclimate/compare/${old_version}...$(git rev-parse --short $branch)"
+pr_description_file=$(mktemp -t cc_commit_message) || exit 1
+printf "Release v$version\n\n$s\n" "$compare_link" > "$pr_description_file"
 if command -v hub > /dev/null 2>&1; then
-  hub pull-request -m "Release v$version\n\n$compare_link"
+  hub pull-request -F "$pr_description_file"
 elif command -v gh > /dev/null 2>&1; then
-  gh pull-request -m "Release v$version\n\n$compare_link"
+  gh pull-request -F "$pr_description_file"
 else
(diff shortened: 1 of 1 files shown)


bin/prep-release
```diff
@@ -34,12 +34,15 @@ git commit -m "Release v$version"
 git push origin $branch
 
 compare_link="https://github.com/codeclimate/codeclimate/compare/${old_version}...$(git rev-parse --short $branch)"
+pr_description_file=`mktemp -t cc_commit_message` || exit 1
```
GordonDiggs: should this use `$(...)`?

wfleming: I borrowed this from an example online, but my understanding was that `$(…)` could hide a bad exit code? I can never remember how `$(…)` and <code>\`…\`</code> differ.

pbrisbin: Never use backtick, always `$( )`. http://mywiki.wooledge.org/BashFAQ/082

bin/prep-release
```diff
@@ -34,12 +34,15 @@ git commit -m "Release v$version"
...
+printf "Release v$version\n\n$s\n" $compare_link > $pr_description_file
 if command -v hub > /dev/null 2>&1; then
-  hub pull-request -m "Release v$version\n\n$compare_link"
+  hub pull-request -F $pr_description_file
```
pbrisbin: :lipstick: quote all the things, these commands and the `rm` below.

pbrisbin: Oh and the usage above at L38

pbrisbin:
Will `gh` / `hub` read the message on `stdin`? If so, you could ditch the tempfile.

wfleming:
@pbrisbin I played with that, but it doesn't seem to support any of the ways you would think would work for that behavior.

wfleming:
Thanks @pbrisbin @GordonDiggs, ready for re-review.

pbrisbin:
LGTM

PR #241, opened 2015-12-02
Fix output reader behavior in the face of timeouts

pbrisbin:
Previous behavior:

- Engine would run and complete in minutes
- Process.waitpid would return a successful status
- The else branch of if @timed_out would be followed
- If the engine produced a significant amount of stdout, and we're slow to read
  it, the output buffers in the docker process (bloating RAM) and we block in
  the Thread#join calls
- The timeout thread eventually reaches 15 minutes and invokes
  reap_running_container directly (not stop)
- The docker kill is a no-op because the container's already finished
- Because it didn't go via #stop, @on_output is left untouched and continues to
  churn

In such cases, the builder runs for as long as it takes to get through the
output. Things eventually do finish, but it takes forever and the memory bloat
of the buffered stdout can cause operational issues.

The fix is to ensure the stdout/stderr processing is governed by the timeout
thread. For this to function correctly, we had to move a few things around:

- When we intentionally stop things, always go through #stop
- If we call #stop, kill the output readers then and there
- Thread#join the output readers *before* checking @timed_out 

In addition, we now pass a message argument through to the log entry written
when we kill a container.

/cc @codeclimate/review

There are two lead-up commits to fix up the specs a bit, then the key failing
spec, then the fix.

I'm going to test this branch locally on my reproduction case before shipping.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -50,17 +50,24 @@ module CC
 
         pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command(options))
 
-        t_out = read_stdout(out)
-        t_err = read_stderr(err)
+        @t_out = read_stdout(out)
+        @t_err = read_stderr(err)
         t_timeout = timeout_thread
 
+        # blocks until the engine stops. there may still be stdout in flight if
+        # it was being produced more quickly than consumed.
         _, status = Process.waitpid2(pid)
+
+        # blocks until all readers are done. they're still governed by the
+        # timeout thread at this point. if we hit the timeout while processing
+        # output, the threads will be Thread#killed as part of #stop and this
+        # will unblock with the correct value in @timed_out
+        [@t_out, @t_err].each(&:join)
 
         if @timed_out
           duration = timeout * 1000
           @listener.timed_out(container_data(duration: duration))
         else
-          [t_out, t_err].each(&:join)
           duration = ((Time.now - started) * 1000).round
           @listener.finished(container_data(duration: duration, status: status))
         end
@@ -74,14 +81,13 @@ module CC
           @stderr_io.string,
         )
       ensure
-        [t_timeout, t_out, t_err].each { |t| t.kill if t }
+        kill_reader_threads
+        t_timeout.kill if t_timeout
       end
 
-      def stop
-        # Prevents the processing of more output after first error
-        @on_output = ->(*) {}
-
-        reap_running_container
+      def stop(message = nil)
+        reap_running_container(message)
+        kill_reader_threads
       end
 
       private
@@ -139,7 +145,7 @@ module CC
           end
 
           @timed_out = true
-          reap_running_container
+          stop("timed out")
         end.run
       end
 
@@ -150,7 +156,7 @@ module CC
 
         if output_byte_count > maximum_output_bytes
           @maximum_output_exceeded = true
-          stop
+          stop("maximum output exceeded")
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/cc/analyzer/container.rb
```diff
@@ -158,8 +164,13 @@ def container_data(duration: nil, status: nil)
...
-      def reap_running_container
-        Analyzer.logger.warn("killing container name=#{@name}")
+      def kill_reader_threads
+        @t_out.kill if @t_out
+        @t_err.kill if @t_err
+      end
+
+      def reap_running_container(message)
+        Analyzer.logger.warn("killing container name=#{@name} message=#{message.inspect}")
         POSIX::Spawn::Child.new("docker", "kill", @name)
```
GordonDiggs: [`docker stop`](https://docs.docker.com/engine/reference/commandline/stop/) will signal `SIGTERM` and then kill after a time. Would there be any benefit in using that here?

GordonDiggs: Maybe with a very low timeout (1)?

lib/cc/analyzer/container.rb
```diff
@@ -50,17 +50,24 @@ def run(options = [])
 
         pid, _, out, err = POSIX::Spawn.popen4(*docker_run_command(options))
 
-        t_out = read_stdout(out)
-        t_err = read_stderr(err)
+        @t_out = read_stdout(out)
+        @t_err = read_stderr(err)
         t_timeout = timeout_thread
```
ABaldwinHunter: curious about why this variable isn't also instance?

ABaldwinHunter:
Minor comment but LGTM. Nice sleuthing and explanatory commit message. So happy to have this bug unearthed.

lib/cc/analyzer/container.rb
pbrisbin: Because it doesn't need to be, it's not referenced outside this method.

lib/cc/analyzer/container.rb
pbrisbin: That makes sense to me, but I'm leaning towards reducing variables and not making that unrelated change as part of this fix.

GordonDiggs: WFM

pbrisbin:
FYI: worked locally, builder stopped with this at the 15 minute mark:

```
[565e25e4d309520001000227] emitting to kafka type=document collection=container_events document={:snapshot_id=>BSON::ObjectId('565e25e4d309520001000227'), :name=>"finished", :container_name=>"cc-engines-csslint-75c10835-8dd5-4a6f-93da-687cec0fecf4", :errored=>true, :exit_code=>nil, :step_name=>"engine:csslint", :host=>"", :image_name=>"codeclimate/codeclimate-csslint:b27", :occurred_at=>2015-12-02 16:56:34 +0000, :stderr=>"", :stdout=>"", :_schema_version=>"2"} immediate_flush=true enqueued_at=2015-12-02 16:56:34 +0000 container_id=837bb9ea5d8f
[565e25e4d309520001000227] emitting to kafka type=state value=error snapshot_id=BSON::ObjectId('565e25e4d309520001000227') details={:error=>{:code=>"E11", :details=>{:output=>"", :backtrace=>nil}}} expectation_set={"0"=>true, "1"=>false} enqueued_at=2015-12-02 16:56:34 +0000 container_id=837bb9ea5d8f
[565e25e4d309520001000227] emitting to kafka type=state value=error snapshot_id=BSON::ObjectId('565e25e4d309520001000227') details={:error=>{:code=>"E11", :details=>{:output=>"", :backtrace=>nil}}} expectation_set={"0"=>false, "1"=>true} enqueued_at=2015-12-02 16:56:34 +0000 container_id=837bb9ea5d8f
W, [2015-12-02T16:56:34.181397 #1]  WARN -- : ** [Bugsnag] Your API key () is not valid, couldn't notify
[565e25e4d309520001000227] backtrace=nil code=E11 exception_class=CC::Builder::Engine::EngineTimeout message=
```

Same build without this patch was still running after an hour.

GordonDiggs:
LGTM. Love the comments

wfleming:
Awesome. Nice spelunking. :+1:

pbrisbin:
Thanks all. Will pick up with a release and builder-side changes after lunch.

PR #248, opened 2015-12-03
Centralize on hub, use stdin

pbrisbin:
- gh was a go rewrite of hub, that eventually took its name
- Using stdin means no mktemp, which didn't work as invoked on Linux

/cc @codeclimate/review


GordonDiggs:
LGTM

PR #249, opened 2015-12-04
Correct default csslint config

pbrisbin:
First, the format was incorrect. The correct[2] JSON format should be:

    {
      "ignore": "rule,rule,..."
    }

However, this format has a known bug[2], therefore I've elected to avoid it and
used the options-passing style.

1: https://github.com/codeclimate/csslint/blob/master/src/cli/common.js#L18
2: https://github.com/CSSLint/csslint/issues/567

Second, had this "default" config worked, it would've ignored *every single*
check (except the should-be-ignored order-alphabetical, which I believe was an
oversight). I doubt very much that was the intention.

Using the corrected and relatively conservative blacklist here, twbs/bootstrap
has ~502 issues (once you exclude doc, vendor, and minified files), the top
offenders of which are:

    132  Use of !important [csslint]
     80  The property -moz-box-shadow is compatible with -webkit-box-shadow and should be included as well. [csslint]
     66  Missing vendor-prefixed CSS gradients for Firefox 3.6+. [csslint]
     20  Using height with padding can sometimes make elements larger than you expect. [csslint]
     16  The property -moz-transition is compatible with -webkit-transition and -o-transition and should be included as well. [csslint]
     12  The universal selector (*) is known to be slow. [csslint]
     12  The box-sizing property isn't supported in IE6 and IE7. [csslint]
     12  Outlines should only be modified using
     12  Missing vendor-prefixed CSS gradients for Firefox 3.6+, Old Webkit (Safari 4+, Chrome). [csslint]
     10  The property -moz-appearance is compatible with -webkit-appearance and should be included as well. [csslint]

These seem reasonable enough to me.

/cc @codeclimate/review


dblandin:
LGTM :+1:

PR #250, opened 2015-12-04
Exclude minified css via --exclude-ext options

pbrisbin:
Excluding by extension is less robust but much easier to implement and more
performant than other options that involve inspecting file content.

The tradeoff is worth it because doing it via a default configuration like this
is both discoverable and easily correctable if it causes users trouble.

/cc @codeclimate/review


ABaldwinHunter:
LGTM

PR #252, opened 2015-12-04
Release v0.15.0

pbrisbin:
/cc @codeclimate/review 

v0.15.7 was a mistake because I forgot to reset minor to 0 when coming from v0.14.7.

I've reverted all of the things and am re-doing it as v0.15.0


pbrisbin:
Looks like it misfired... trying again

PR #254, opened 2015-12-08
Ignore box-model in default csslintrc

pbrisbin:
Details: https://github.com/CSSLint/csslint/wiki/Beware-of-box-model-size

Our reasoning is that if you understand the CSS box model, you would find this
rule annoying and we should optimize for experienced CSS authors.

/cc @codeclimate/review


wfleming:
:+1:

ABaldwinHunter:
:+1:

PR #257, opened 2015-12-08
Attempt to address flaky CI

pbrisbin:
Example: https://circleci.com/gh/codeclimate/codeclimate/799

I've attempted to reproduce locally (with the same seed) and after looping for
a half an hour it's never failed. The error message is confusing, it states
that finished() was an unexpected invocation even though finished() is expected
and not yet invoked.

The ContainerData argument is probably a Red Herring, as that argument should
be present all the time, not (intermittently) missing. Therefore, I suspect
it's the boolean assertion within the with-block returning false intermittently
and materializing as the failure with a poor message (yay minitest-mocha) --
yes, this may indicate a real bug.

I believe moving the stdout_lines assertion outside of the with does not weaken
it, and I'm hoping it'll be more stable or at least provides a clearer message
when failing.

/cc @codeclimate/review

### DIFF
--- a/spec/cc/analyzer/container_spec.rb
+++ b/spec/cc/analyzer/container_spec.rb
@@ -156,9 +156,7 @@ module CC::Analyzer
         it "waits for IO parsing to finish" do
           stdout_lines = []
           listener = TestContainerListener.new
-          listener.expects(:finished).once.with do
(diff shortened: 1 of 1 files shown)


GordonDiggs:
Seems good to me

PR #259, opened 2015-12-09
Enable duplication engine automatically for supported languages

dblandin:
This PR updates our engines registry to enable the [duplication engine][] for any project with Ruby, Python, JavaScript, or PHP files.

@codeclimate/review :mag_right:

[duplication engine]: https://circleci.com/gh/codeclimate/codeclimate-duplication

PR #263, opened 2015-12-10
Enable duplication engine automatically for supported languages

dblandin:
Duplicate of #259 

This PR updates our engines registry to enable the [duplication engine][] for any project with Ruby, Python, JavaScript, or PHP files.

[duplication engine]: https://circleci.com/gh/codeclimate/codeclimate-duplication

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -55,7 +55,21 @@ duplication:
   description: Structural duplication detection for Ruby, Python, JavaScript, and PHP.
   community: false
   enable_regexps:
+    - \.inc$
+    - \.js$
+    - \.jsx$
+    - \.module$
+    - \.php$
+    - \.py$
+    - \.rb$
   default_ratings_paths:
+    - "**.inc"
+    - "**.js"
+    - "**.jsx"
+    - "**.module"
+    - "**.php"
+    - "**.py"
+    - "**.rb"
   default_config:
     languages:
       - ruby
--- a/spec/cc/cli/init_spec.rb
+++ b/spec/cc/cli/init_spec.rb
@@ -31,11 +31,12 @@ module CC::CLI
           YAML.safe_load(new_content).must_equal({
             "engines" => {
               "csslint" => { "enabled"=>true },
+              "duplication" => { "enabled" => true, "config" => { "languages" => ["ruby", "javascript", "python", "php"] } },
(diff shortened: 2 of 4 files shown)


config/engines.yml
```diff
@@ -55,7 +55,17 @@ duplication:
   description: Structural duplication detection for Ruby, Python, JavaScript, and PHP.
   community: false
   enable_regexps:
+    - \.js$
+    - \.jsx$
+    - \.php$
```
GordonDiggs: PHP [has a few extensions](https://github.com/codeclimate/codeclimate-duplication/blob/master/lib/cc/engine/analyzers/php/main.rb#L13-L15)

spec/cc/cli/init_spec.rb
```diff
@@ -170,11 +171,12 @@ module CC::CLI
           YAML.safe_load(new_content).must_equal({
             "engines" => {
               "csslint" => { "enabled"=>true },
+              "duplication" => { "enabled" => true, "config" => { "languages" => ["ruby", "javascript", "python", "php"] } },
```
GordonDiggs: Is this array format still valid? I've been doing them as mappings recently

dblandin: Looks like the array format is still [valid](https://github.com/codeclimate/codeclimate-duplication/blob/ce27d9e3b99513115968df64eafa13c8693d5a3c/lib/cc/engine/analyzers/engine_config.rb#L57-L67). The hash format is necessary when you're overriding mass threshold for a language.

config/engines.yml
dblandin: @GordonDiggs Are those extensions common enough to include by default or should we leave it as an opt in? If we add those two other extensions here, we should probably add them for ratings as well.

GordonDiggs: They are definitely common. I would support adding them in both

dblandin: Sounds good :+1:

dblandin:
Updated to include on `.inc` and `.module` files.

spec/cc/cli/init_spec.rb
GordonDiggs: good to know! thanks

GordonDiggs:
:+1:

PR #265, opened 2015-12-10
Run Circle specs in docker

pbrisbin:
- The image and test targets are generally useful
- Running specs in a container should mean more consistent behavior
- Working with a codeclimate/codeclimate image until and unless we're deploying
  to GCR, at which time we tag and push, feels simpler

/cc @codeclimate/review

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,9 +1,6 @@
 machine:
   services:
     - docker
-  environment:
-    CLOUDSDK_CORE_DISABLE_PROMPTS: 1
-    PRIVATE_REGISTRY: us.gcr.io/code_climate
 
 dependencies:
   override:
@@ -14,13 +11,3 @@ dependencies:
 test:
   override:
     - make test_only
-
-deployment:
-  registry:
-    branch: master
-    commands:
-      - echo $gcloud_json_key_base64 | sed 's/ //g' | base64 -d > /tmp/gcloud_key.json
-      - curl https://sdk.cloud.google.com | bash
-      - gcloud auth activate-service-account --key-file /tmp/gcloud_key.json
-      - gcloud docker -a
-      - docker push $PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
--- a/circle.yml
+++ b/circle.yml
@@ -9,12 +9,11 @@ dependencies:
   override:
     # Used by container_spec
     - docker pull alpine
-    - docker build -t=$PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM .
-    - bundle install
+    - make image
 
 test:
   override:
-    - bundle exec rake
+    - make test_only
 
 deployment:
   registry:
--- a/circle.yml
+++ b/circle.yml
@@ -11,3 +11,7 @@ dependencies:
 test:
   override:
     - make test_only
+
+notify:
+  webhooks:
+    - url: https://cc-slack-proxy.herokuapp.com/circle
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -173,24 +173,7 @@ module CC
         Analyzer.logger.warn("killing container name=#{@name} message=#{message.inspect}")
         POSIX::Spawn::Child.new("docker", "kill", @name)
 
-        # On Circle, the tests around this method fail when run via make test,
-        # within a container process.
-        #
-        # Known:
-        #
-        #   - The docker kill succeeds
-        #   - The container process exits with SIGPIPE and no exit code
-        #   - Therefore, the assertion on the exit code fails
-        #
-        # Theory: docker kill succeeds, but the container hasn't actually
-        # stopped when we go ahead and kill the output reading threads. The loss
-        # of the output readers causes the still-running container to SIGPIPE
-        # which causes the result to not have an exit code.
-        #
-        # This sleep is a bit of a hack, but passes the specs. It shouldn't
-        # negatively impact production as we're already in a kill scenario, so
-        # small delay between killing the engine and killing the output reading
-        # threads doesn't strike me as a huge deal
+        # Required for specs to pass on Circle. See commit message for details.
         sleep 2
       end
 
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -174,7 +174,7 @@ module CC
         POSIX::Spawn::Child.new("docker", "kill", @name)
 
         # Required for specs to pass on Circle. See commit message for details.
-        sleep 2
+        #sleep 2
       end
 
       def timeout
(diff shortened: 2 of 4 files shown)


circle.yml
```diff
@@ -22,6 +21,6 @@ deployment:
     commands:
       - echo $gcloud_json_key_base64 | sed 's/ //g' | base64 -d > /tmp/gcloud_key.json
       - curl https://sdk.cloud.google.com | bash
+      - docker tag codeclimate/codeclimate $PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
       - gcloud auth activate-service-account --key-file /tmp/gcloud_key.json
-      - gcloud docker -a
-      - docker push $PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
+      - gcloud docker push $PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
```
pbrisbin: I'm actually curious why we do this. Do we ever use the CLI images on GCR?

dblandin: I'm not sure either. We [push to docker hub](https://github.com/codeclimate/codeclimate/blob/master/bin/release#L21) as part of the `bin/release` script, which is [pulled during installation](https://github.com/codeclimate/codeclimate/blob/76981f8d17c9184948ff8344dc6d70def359ca93/Makefile#L10).

As far as I know, we just install codeclimate as a gem in builder.

circle.yml
```diff
@@ -9,12 +9,11 @@ dependencies:
...
     # Used by container_spec
     - docker pull alpine
-    - docker build -t=$PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM .
-    - bundle install
+    - make image
 
 test:
   override:
-    - bundle exec rake
+    - make test
```
wfleming: Small consideration: I believe this will run `make image` *again*. Which should be fast since the image will already exist, but is doing needless work. Maybe we should just remove the build from `dependencies` and let it happen here?

circle.yml
wfleming: There was a time when builder just ran `codeclimate init` using the cc binary. That may have used the docker image, so maybe this is vestigial to that?

lib/cc/analyzer/container.rb
```diff
@@ -72,6 +72,8 @@ def run(options = [])
           @listener.finished(container_data(duration: duration, status: status))
         end
 
+        p status
```
wfleming: This looks like debugging code?

circle.yml
dblandin: Or we could remove the image pre command and keep `make image` in dependencies. I personally like seeing the two actions in separate sections on CircleCI:

screen-shot-2015-12-10-14-12-07-rsqemsyjit

pbrisbin: Yup, you're right -- there's no reason to have it in the circle dependencies.

lib/cc/analyzer/container.rb
pbrisbin: Oh I'm sorry. I started down a debugging track on Circle with WIP commits. I should've marked the PR itself WIP too.

pbrisbin:
I'm a little baffled here. Shouldn't running the specs in a container make them *more* reliable? Not less? Everything passes locally, consistently...

circle.yml
pbrisbin: @dblandin I agree with that, and that's why I put it in there original. Unfortunately, I don't want to take it out of the Makefile because then I'm sure there will be a day, or many days, I run `make test` and can't figure out why my changes aren't having the desired effect.

I'd probably go back to what I had. Even though it's technically duplicated work, it is a 100% cached and fast build when Circle runs the test. I don't feel strongly though. @wfleming ?

circle.yml
pbrisbin: I think even then, it was using the gem executable not a docker image.

pbrisbin:
I looks like on Circle the container's not stopping as quickly as locally, so it's still running when we kill the output reading threads causing a SIGPIPE instead of an exit with an exit code like locally. Adding a call to `docker ps` after the kill (which I did to debug) slows things down enough to pass. Feels like an actual race we could be suffering from in production -- but I'm not sure how to best fix it.

circle.yml
dblandin: @pbrisbin Makes sense. Having the second 100% cached build sounds like a good move then. Generally I don't think that docker image build time should contribute to  test suite time.

wfleming: Eh. Whatever. None of the solutions seem good.

wfleming: Thought of something else about why this kind of bugs me: I agree with Devon that having the desperate step in circle is nice. And I also agree having the task be dependent in the Makefile is important. I think doing *both* is a false advantage because, yes, you have the build in a separate step, but you are also polluting the test step output with the same build output all over again (except that every later will be cached), which is bad.

So on balance I still think we should either have separate circle steps *or* Makefile dependencies, but not both. Personally I lean towards keeping the Makefile dependencies since that seems like a more important use case for local development.

dblandin: All good points. What do you think of a separate Makefile command called `citest` that does not have the `image` dependency?

wfleming: If we're going that direction that much, I would propose calling the task `test_only`, and then declaring `test` as just `test: image test_only` with no other content.



~will out.

lib/cc/analyzer/container.rb
```diff
@@ -172,6 +172,26 @@ def kill_reader_threads
...
+        # Theory: docker kill succeeds, but the container hasn't actually
+        # stopped when we go ahead and kill the output reading threads. The loss
+        # of the output readers causes the still-running container to SIGPIPE
+        # which causes the result to not have an exit code.
+        #
+        # This sleep is a bit of a hack, but passes the specs. It shouldn't
+        # negatively impact production as we're already in a kill scenario, so
+        # small delay between killing the engine and killing the output reading
+        # threads doesn't strike me as a huge deal
+        sleep 2
```
pbrisbin: Interested in folks' thoughts here. If we feel this is too hacky to commit, I think I'll have to abandon this approach. Without some sort of delay after the kill, it fails reliably on Circle.

pbrisbin:
This is ready for re-review. @codeclimate/review

dblandin:
LGTM

pbrisbin:
Ugh. I cannot get this thing to pass reliably. Back to some CI debugging.

pbrisbin:
I have to abandon this. I can't get it to pass on Circle.

I'll re-open if I get some time to continue digging.

lib/cc/analyzer/container.rb
wfleming: I'm ok with this. You've explained it thoroughly enough. I wonder if it would be more reliable to poll for whether the container pid is still alive? Probably not worth it.

pbrisbin: I considered that, but then we run the risk of a run-away poll meaning a run-away builder in production. Not something I wanted to add because of flakiness on Circle.

In the end though, this sleep seems to introduce additional flakiness that I was able to see locally as well, so it's not a good enough solution. I've abandoned this branch for now.

PR #266, opened 2015-12-11
Fix fingerprint excludes

GordonDiggs:
All @wfleming's work here, but I tested it and it is working. The issue previously was that the the fingerprint defaulting happened _after_ the output filtering. This ports a bunch of stuff over from builder to do the issue validation here. A future PR will remove this all from builder.

@codeclimate/review


wfleming:
Thanks @gordondiggs. I'm happy to have this go out as-is if it's urgent, but I wasn't really considering this quite ready to ship: my plan was to shift things so that the adapted output was emitted from the the engine class here in the CLI, and and then remove all of this code from builder so that it was only happening in one place.

The biggest piece of this that is still kind of in flux is actually error handling: right now adapter will raise for issue validation problems. If we need to ship this ASAP, the output filter should probably catch those & return false to avoid changing error handling behavior elsewhere.

GordonDiggs:
@wfleming thanks! I will work on that

wfleming:
Let's touch base later today if you don't finish it today. I was going to work on it this weekend.

PR #267, opened 2015-12-11
Generate default fingerprints for issues

GordonDiggs:
Previously, excluding issues by fingerprints did not work because the
fingerprint generation happened _after_ the filtering. This moves the
fingerprint generation to a new class and then passes the hash to
EngineOutputFilter.

@codeclimate/review


lib/cc/analyzer/engine.rb
```diff
@@ -34,8 +34,12 @@ def run(stdout_io, container_listener)
         )
 
         container.on_output("\0") do |output|
-          unless output_filter.filter?(output)
-            stdout_io.write(output) || container.stop
+          next if output.blank?
+
+          issue_hash = IssueHash.new(output)
```
pbrisbin: Mentioned IRL, but for visibility:

I think we should go with an actual `Issue` class we can pass around and filter on that responds to methods. Then we `#to_json` that for the `stdout_io.write` call.

wfleming: I'm :+1: on @pbrisbin's approach. That was the direction I was trying to move in.

GordonDiggs:
@pbrisbin @codeclimate/review ready for another look!

lib/cc/analyzer/issue.rb
```diff
@@ -0,0 +1,58 @@
...
+        )
+      end
+
+      def fingerprint
+        parsed_output.fetch("fingerprint", default_fingerprint)
+      end
+
+      # Allow access to hash keys as methods
+      SPEC_ISSUE_ATTRIBUTES.each do |key|
+        define_method(key) do
```
GordonDiggs: Open to other ideas on this. I'm torn between doing this and using `method_missing`. What do others think?

wfleming: I prefer this over method missing.

wfleming:
I'm still weirded it that we're still not validating output at any point here: unless I've missed something, it seems wrong to me that Issue validation behavior is essentially part of IOProcessor in builder. In other words, it's an implementation detail of a particular output formatter, not core behavior of how engines are run.

I know I probably missed in-office discussion, so ignore me if that's not germane. If the goal was to get just this bug fixed fast, this approach seems ok. But the responsibilities of where input "adaptation" & validation continue to feel wrongly separated to me: fingerprint generation happens here, but validation (including location validation, which fingerprint partial depends on!) happen later.

lib/cc/analyzer/issue.rb
GordonDiggs: I think I do too

GordonDiggs:
@wfleming the only in-office discussion was that it felt weird to pull all of `SmellAdapter` over because our internal database schema doesn't seem like a concern of the CLI. In that regard, I think the adaptation should still happen in builder. 

However, I agree on the validation - I'll move that stuff into here now.

wfleming:
That makes sense to me.

Maybe validation should be a next, separate PR since it will impact builder with error reporting. As long as we're tracking it, I guess.

GordonDiggs:
@wfleming that might be a good idea. As it stands now, integrating this with builder means bumping the gem and removing some code, without much other messing around. I will add a card for the validation stuff and punt on it for now, unless it gets hairy once I get over to builder land

lib/cc/analyzer/engine.rb
```diff
@@ -34,8 +34,12 @@ def run(stdout_io, container_listener)
         )
 
         container.on_output("\0") do |output|
-          unless output_filter.filter?(output)
-            stdout_io.write(output) || container.stop
+          next if output.blank?
+
+          issue = Issue.new(output)
+
+          unless output_filter.filter?(issue)
```
pbrisbin: Doh, I guess this should be `issue_filter` now. I'm fine punting on the rename for now, it'd muddy the diff quite a bit I think.

wfleming:
Makes sense to me. Even though I know the card will languish in icebox & that makes me sad.

spec/cc/analyzer/issue_spec.rb
```diff
@@ -0,0 +1,94 @@
+require "spec_helper"
```
ABaldwinHunter: nice job adding specs!

lib/cc/analyzer/engine_output_filter.rb
```diff
@@ -7,42 +7,26 @@ def initialize(config = {})
...
       end
 
       private
 
-      def parse_as_json(output)
-        JSON.parse(output)
-      rescue JSON::ParserError
-        nil
+      def issue?(issue)
+        issue.type.present? && issue.type.downcase == ISSUE_TYPE
```
pbrisbin: I'm having some naming trouble here: hey instance of `Issue` are you an `issue`? feels wrong. I think we either

- Don't construct an `Issue` unless it's the right type, or
- Rename `Issue` to something more generic, then it makes sense to ask that thing if it's an "issue" (vs, I guess, a warning?)

WDYT?

lib/cc/analyzer/engine.rb
wfleming: I don't think it'd muddy the diff that much. Might as well do it now if it's better.

lib/cc/analyzer/issue.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def default_fingerprint
+        digest = Digest::MD5.new
+        digest << path
+        digest << "|"
+        digest << check_name.to_s
+        digest.to_s
+      end
+
+      def parsed_output
```
ABaldwinHunter: what do you think about keeping the JSON parse error rescue here?

as in
https://github.com/codeclimate/codeclimate/pull/267/files#diff-f70998052f6e066aec2b9393461c245cL24

pbrisbin:
This looks really good. I think my only concern is using the name `Issue` for something that's really just generic, arbitrary engine output that we interpret as issues based on a `type` key. I think renaming it to something like `Output` would make things a lot better. You could then have `Output#issue?` and maybe even `Output#as_issue` to get an `Issue`, if it makes sense still to also have that object when you have confirmed it was an issue output by the engine.

lib/cc/analyzer/issue.rb
GordonDiggs: The other mechanisms of this class require that JSON.parse to succeed, and that one was only rescued in order to pass on invalid JSON for this very error to be raised later, so I think this is better as-is

lib/cc/analyzer/engine.rb
ABaldwinHunter: I think output filter is fine still. It's still filtering the output, no?

pbrisbin: You're right. I think it's `issue` that's actually the wrong name here, as mentioned in other comments.

lib/cc/analyzer/engine.rb
```diff
@@ -33,9 +33,11 @@ def run(stdout_io, container_listener)
           listener: composite_listener,
         )
 
-        container.on_output("\0") do |output|
+        container.on_output("\0") do |raw_output|
+          output = EngineOutput.new(raw_output)
+
           unless output_filter.filter?(output)
-            stdout_io.write(output) || container.stop
+            stdout_io.write(output.as_issue.to_json) || container.stop
```
pbrisbin: Add a `#to_json` delegate to prevent angering Demeter?

lib/cc/analyzer/engine_output.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      def issue?
+        parsed_output &&
+          parsed_output["type"].present? &&
+          parsed_output["type"].downcase == "issue"
+      end
+
+      def as_issue
+        Issue.new(raw_output)
```
pbrisbin: Would it make sense to guard this with `#issue?`?

pbrisbin: If so, I might `raise` on `false` instead of (e.g.) returning `nil`.

pbrisbin:
Two minor thoughts. LGTM.

ABaldwinHunter:
:+1:

lib/cc/analyzer/engine_output.rb
GordonDiggs: That starts to get into the validation territory IMO (if we're raising then we need to make sure it matches the errors builder expects, etc) - curious to see how this plays out as-is

PR #269, opened 2015-12-11
Autoexclude config paths

ABaldwinHunter:
This change reduces nonsensical errors -e.g. fixme engine finding `fixme` in .codeclimate.yml

Also prevents source file not found errors that come up when issues are found in config files that
are autogenerated during analysis on .com, but don't actually exist in repo.

e.g.
https://codeclimate.com/repos/566b4d8513282e4989000f3a/issues/categories/bugrisk

cc @codeclimate/review

### DIFF
--- a/lib/cc/cli/config_generator.rb
+++ b/lib/cc/cli/config_generator.rb
@@ -2,7 +2,7 @@ module CC
   module CLI
     class ConfigGenerator
       CODECLIMATE_YAML = Command::CODECLIMATE_YAML
-      AUTO_EXCLUDE_PATHS = %w(config/ db/ features/ node_modules/ script/ spec/ test/ tests/ vendor/).freeze
+      AUTO_EXCLUDE_PATHS = %w(.codeclimate.yml .csslintrc .eslintrc .rubocop.yml config/ db/ features/ node_modules/ script/ spec/ test/ tests/ vendor/).freeze
 
       def self.for(filesystem, engine_registry, upgrade_requested)
         if upgrade_requested && upgrade_needed?(filesystem)


ABaldwinHunter:
example bugsnag https://bugsnag.com/code-climate/app/errors/562a8f5622e62149658053ac?filters%5Bevent.since%5D%5B%5D=30d

pbrisbin:
I think @brynary has strong opinions against fixing it this way. In cases like this he wants the responsibility to lie with the engine, leaving the runner as dumb as possible. I'm parroting though, so I'd let him chime in before moving forward or abandoning on this.

ABaldwinHunter:
@pbrisbin Okay, I'll wait for @brynary's thoughts. I like engines being discriminate about what files they analyze as well, but in this case we're also passing them config files that don't actually belong to the repo. Figment of our autogeneration.

noahd1:
That's a good point -- but there are still cases where they aren't figments of autogeneration, right?

ABaldwinHunter:
@noahd1 yes. But v few have seemed valuable. I don't think we have any engines that analyze config files ATM except `fixme` - and its finding of its own name has been deemed irksome. [E.g.](https://github.com/codeclimate/codeclimate-fixme/issues/15) and [here](https://github.com/codeclimate/codeclimate-fixme/issues/17). (Agree that's more of a FIXME bug than CLI bug).

We may however have a `yml` linter in the future. 

For me, biggest selling point is that the `autoexclude` paths only get generated if a user runs `codeclimate init` in the CLI.

Otherwise, it gets run on prod when there's no codeclimate.yml.

brynary:
@ABaldwinHunter are there other specific issues we've identified with not having these excludes in places _besides the fixme issue_?

ABaldwinHunter:
@brynary at the moment there are no others outstanding.

Agree it feels like this fix might fit better in the `fixme` engine. 

Although not a problem at the moment, it seems potentially hazardous to analyze generated files long term.

I'll go ahead and close this PR for now in favor of a change to `fixme`.

PR #271, opened 2015-12-14
Disable PEP-8 and PHP CodeSniffer by default

jpignata:
These engines will only be ran for Python and PHP projects respectively
when explicitly enabled via a project's codeclimate.yml.

This change strips each of their `enable_regexps` values and their
`upgrade_languages` keys.

@codeclimate/review @brynary


jpignata:
@codeclimate/review Any thoughts or objections?

brynary:
LGTM

PR #276, opened 2015-12-17
eslint configuration changes

jpignata:
* Ignore minified files

> Ignore any file suffixed by .min.js to avoid analyzing minified files.

* Tweak default eslint configuration

> * Remove any rules that were set to 0 as that's the default
* Add modules to ecmaFeatures to permit import and export keywords
* Enable additional rules to highlight Bug Risk, Clarity, Security, and
  Compatibility issues
* Disable all "Stylistic Rules" by default
* Use inline arrays instead of multiline
* Group rules by documentation groupings
* Alphabetize groups of rules


wfleming:
> Remove any rules that were set to 0 as that's the default

I recall there being discussion around this & other defaults that we were intentionally including rules set to defaults for discoverability. I.e. It's easier for a customer to realize they can turn some knobs if it's already in the file we provide than if they have to go research docs for that tool.

jpignata:
That makes some good sense. I'll re-add them and re-add a comment I had
pulled before pushing that points to the documentation site.

PR #279, opened 2015-12-18
Don't enable duplication by default

wfleming:
By surfacing & no longer silently swallowing errors that could lead to non-deterministic results, we are now in a bit of a tough spot because this is surfacing errors that look like straight-up bugs in the babel parser, but which programmatically aren't very distinguishable from previously mentioned non-deterministic results bugs.

(The underlying errors are parsing timeouts: previously these were low at 20 seconds and being swallowed, which was bad because memory pressure/gc behavior could swing execution to one side or the other of that line. Now they are high at 5 minutes & fatal, but due to what look like babel issues reliably happen on some files. It might be reasonable to swallow these extreme timeouts, but it would only take 2 or 3 of them to timeout the whole engine, so it hardly seems worthwhile.)

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -54,14 +54,14 @@ duplication:
   image: codeclimate/codeclimate-duplication
   description: Structural duplication detection for Ruby, Python, JavaScript, and PHP.
   community: false
-  enable_regexps:
-    - \.inc$
-    - \.js$
-    - \.jsx$
-    - \.module$
-    - \.php$
-    - \.py$
-    - \.rb$
+  # enable_regexps:
+  #   - \.inc$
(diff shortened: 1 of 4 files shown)


config/engines.yml
```diff
@@ -156,7 +156,7 @@ phpmd:
   enable_regexps:
     - \.php$
     - \.module$
-    - \.inc$
+    - \.inc$d
```
GordonDiggs: looks like an erroneous character here

mrb: Bahhh

wfleming:
this is time sensitive, so i'm going to mercy merge to avoid rebasing again.

PR #282, opened 2015-12-22
Print issue with begin line only, if no endline is specified

ABaldwinHunter:
This way, engines can also pass simply lines: { begin: line_num }

We should do this and be resilient to a questionable but not illogical
input, or complain loudly.

**Fix**: currently the brakeman engine only returns `begin: line`

```CONSOLE
Starting analysis
Running brakeman: Done!

== Gemfile.lock (2 issues) ==
87-: Rails 4.2.0 does not encode JSON keys (CVE-2015-3226). Upgrade to Rails version 4.2.2 [brakeman]
87-: Rails 4.2.0 is vulnerable to denial of service via XML parsing (CVE-2015-3227). Upgrade to Rails version 4.2.2 [brakeman]
```

cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/location_description.rb
+++ b/lib/cc/analyzer/location_description.rb
@@ -10,7 +10,7 @@ module CC
       def to_s
         if location["lines"]
           begin_line = location["lines"]["begin"]
-          end_line = location["lines"]["end"]
+          end_line = location["lines"]["end"] || begin_line
         elsif location["positions"]
           begin_line = position_to_line(location["positions"]["begin"])
           end_line = position_to_line(location["positions"]["end"])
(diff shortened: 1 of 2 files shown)


jpignata:
Rather than assigning a value to `end_line` that wasn't intended for it, what do you think about just changing the `render_lines` methods to do what you want explicitly?

```ruby
def render_lines(begin_line, end_line)
  [begin_line, end_line].uniq.compact.join("-")
end
```

On a higher level, the spec and validations should probably mandate what we input we actually expect here. Is just a `begin_line` even valid syntax today? If not, is it desirable? cc @mrb

ABaldwinHunter:
@jpignata Interested in hearing @mrb's thoughts, but I like your idea a lot.

An intelligent render method sounds much smarter.

PR #283, opened 2015-12-23
Update default model .rubocop.yml

ABaldwinHunter:
- Order sections alphabetically, putting Style last
- -> Lint, Metrics, Performance, Rails, Style
- Turn on Metrics/MethodLength check (for complexity) by default
- increase method length max from 10 -> 30
- increase class length max from 100 -> 250

@codeclimate/review @brynary

### DIFF
--- a/config/rubocop/.rubocop.yml
+++ b/config/rubocop/.rubocop.yml
@@ -3,7 +3,388 @@
 AllCops:
   DisabledByDefault: true
 
-################## STYLE #################################
+# Headings:
+# Lint, Metrics, Performance, Rails, Style
+
+#################### Lint ################################
+### Warnings
+
+Lint/AmbiguousOperator:
+  Description: >-
+                 Checks for ambiguous operators in the first argument of a
+                 method invocation without parentheses.
+  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#parens-as-args'
+  Enabled: true
+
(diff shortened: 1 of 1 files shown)


ABaldwinHunter:
Note: Other complexity checks are present but disabled (ABCsize, cyclomatic complexity)

jpignata:
@ABaldwinHunter This is a rather large commit. Attempting to diff the two, I'm noticing differences not noted in your commit message. For example, why did we decide to disable `Metrics/AbcSize` by default? 

Can you please unwind your reordering changes from this pull and make smaller commits that we can compare to the original and discuss?

ABaldwinHunter:
@jpignata Of course. Will do.

Sorry it's not in commit message, I commented above:
```
Note: Other complexity checks are present but disabled (ABCsize, cyclomatic complexity, perceived complexity)
```

From my discussion with @brynary, I think the reasoning is that the variety of complexity checks present several evaluation options which could be used in conjunction with one another, but would create some redundancy. (Imagine an inspector measuring building height and penalizing for being too high in meters, feet, and stories).

For a first run experience, `class` and `method length` were considered intuitive and thus good introductory units of measure.

ABaldwinHunter:
Closing this PR in favor smaller increments, starting with a reordering https://github.com/codeclimate/codeclimate/pull/284

PR #286, opened 2015-12-24
More eslintignore defaults

noahd1:
[eslint docs](http://eslint.org/docs/user-guide/configuring.html#ignoring-files-and-directories) claim to support brace expansion. Tested locally, and it seems they are not lying.

/c @codeclimate/review


jpignata:
`**/*{.,-}min.js` seems reasonable to me.

On `**/*{.,-}minified.js`, have we seen this often? Seems like perhaps we should have some minimum threshold of usage in the wild before we add patterns here. I imagine the 95% case is the jQuery convention of `.min.js`.

noahd1:
I have only seen 1 case of it to date so we could wait. My initial thought was "what's the harm if there's little chance of a false positive." However, I guess that doesn't preclude us adding "*-my-mega-optimized-tiny-file.js" and whatever weird other conventions people have.

jpignata:
Up to you. LGTM either way.

noahd1:
Removing `minified` pattern for now.

PR #287, opened 2015-12-28
Increase rubocop defaults for complexity checks

ABaldwinHunter:
Increase Class and Module length Max from 100 to 250

Turn on method length check
Increase method length Max from 10 to 80

- In light of grade point tuning
- leaves ABCSize enabled, Cyclomatic Complexity and Perceived Complexity disabled

@codeclimate/review

### DIFF
--- a/config/rubocop/.rubocop.yml
+++ b/config/rubocop/.rubocop.yml
@@ -216,7 +216,7 @@ Metrics/AbcSize:
                  A calculated magnitude based on number of assignments,
                  branches, and conditions.
   Reference: 'http://c2.com/cgi/wiki?AbcMetric'
-  Enabled: true
+  Enabled: false
   Max: 20
 
 Metrics/BlockNesting:
@@ -234,7 +234,7 @@ Metrics/CyclomaticComplexity:
   Description: >-
                  A complexity metric that is strongly correlated to the number
                  of test cases needed to validate a method.
-  Enabled: false
+  Enabled: true
 
 Metrics/LineLength:
   Description: 'Limit lines to 80 characters.'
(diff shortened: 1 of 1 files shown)


pbrisbin:
These seem way too lenient to me, I think an 80 line method is
unacceptable, and 250 for a class is pretty bad.

I'm fine with it if these have already been discussed and decided in a
larger group, but if it's still up for discussion, I'd vote for limits of
150 and 15.

wfleming:
I'm agree with @pbrisbin that these new numbers seem to large. What's the motivating issue here? I feel like the farther we stray from the underlying tool's defaults, the more potential confusion we cause for customers.

ABaldwinHunter:
@wfleming @pbrisbin thanks. These were discussed and decided on. 

I'm fine with any changes though - nothing's set until set. cc @brynary 

Related point changes in codeclimate-rubocop: https://github.com/codeclimate/codeclimate-rubocop/pull/42

ABaldwinHunter:
For some context / summary of discussion:

- Main motivation is to make scoring more similar to Classic
- These changes only impact the default .rubocop.yml generated during `init`
- The remediation points for complexity violations are steeper in the PR linked to above. Effectively, default is that users don't trigger the complexity violation until a higher threshold, but when they do, it's an automatic C.
- All complexity checks provided by Rubocop are included in our default config for informational purposes, but some are disabled. They're not necessarily meant to be used in conjunction with one another. Might create some redundant remediation scoring.

The Class and Method `length` checks seemed a more intuitive measure than `ABCSize` for beginners.

@brynary correct me on any of the above. :) 

cc @jpignata

ABaldwinHunter:
@codeclimate/review Note: I had the method length incorrect. Should have been 30. Updated.  

cc @brynary

wfleming:
30 seems much more reasonable than 80. :)

brynary:
LGTM

PR #290, opened 2015-12-29
Tag docker images with version

wfleming:
In addition to git tags, this will tag docker releases with the version tag, in addition to updating "latest"

Would be nice to adjust the homebrew formula to pull the correct tag at some point as well, but that's more complicated, so I probably won't do it yet.

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -18,7 +18,9 @@ printf "RELEASE %s\n" "$version"
 rake release
 
 docker build --rm -t codeclimate/codeclimate .
-docker push codeclimate/codeclimate
+docker push codeclimate/codeclimate:latest
+docker tag codeclimate/codeclimate "codeclimate/codeclimate:$version"
+docker push "codeclimate/codeclimate:$version"
 
 (cd ../homebrew-formulae/ && bin/release "$version")
 


bin/release
```diff
@@ -19,6 +19,8 @@ rake release
 
 docker build --rm -t codeclimate/codeclimate .
 docker push codeclimate/codeclimate
+docker tag codeclimate/codeclimate "codeclimate/codeclimate:$version"
+docker push "codeclimate/codeclimate:$version"
```
dblandin: Nice :+1: 

I think that  `docker push` will push all tagged images so you can move the `docker tag` line up and drop the second `docker push`.

```bash
 docker build --rm -t codeclimate/codeclimate .
 docker tag codeclimate/codeclimate codeclimate/codeclimate:$version
 docker push codeclimate/codeclimate
```

GordonDiggs: I'm -1 on that. Way too easy to have other tagged images locally that will be pushed up

wfleming: Agreeing with @GordonDiggs on this: I think it's better to be explicit and only push the release we care about.

dblandin: Sounds good to me. I don't think that's the script's current behavior so we should probably qualify the first `docker push` as well:

```bash
docker push codeclimate/codeclimate:latest
```

wfleming: Very good point: we actually had a debug tag of mine on docker hub already because of this! Fixed.

PR #291, opened 2015-12-31
PoC: CLI v2 as extracted sub-commands

pbrisbin:
Proof-of-concept PR for codeclimate/rfc#11: CLI as a suite of sub-commands.

As a first, medium-complexity example, `validate-config` was extracted. I'd
consider this a good example of extracting something that's complicated enough
to require Ruby, but simple enough to remain confined to just the bin/run
script and a single black-box spec.

In the process of extraction, some bugs were fixed, some behaviors added or
improved, and (I believe) the code is simpler and clearer in both
implementation and specs. I believe all these improvements were a direct result
of uncoupling the logic and forcing the code to remain isolated to a separate
source tree.

Two simple commands (`version` and `help`) were also extracted. This is to show
how some functionality doesn't even need to venture into Ruby land and can
exist fine as a few lines of shell.

There is additional details in the individual commit messages.

/cc @codeclimate/review


codeclimate-validate-config/bin/run
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env ruby
+#
+# TODO: validate engine names as valid engines. this requires access to the
+# shared engine registry, so we should think carefully if that functionality is
+# worth the coupling.
```
pbrisbin: Is the intention of `codeclimate engines:list` to list all *available* engines? If so, we could use that here to perform this validation. That would be well in line with the spirit of this architecture. The `codeclimate-engines` subcommand could encapsulate access to the registry.

wfleming: There's a bit of an edge case with this idea for when users are locally developing new engines. Maybe not worth thinking about that edge case. Even ignoring that I feel this is kind of a non-feature and not very worth the coupling & implementation effort.

codeclimate-validate-config/Gemfile.lock
```diff
@@ -0,0 +1,49 @@
+GEM
```
wfleming: I'm not sure it's best for us to split commands to the point where they have entirely different dependency trees. That seems like a good way to some day introduce bugs & inconsistent behavior from updating some commands and not others.

FWIW, git doesn't structure itself this way: dependencies are "global".

codeclimate-validate-config/bin/run
```diff
@@ -0,0 +1,78 @@
...
+
+def colorize(text, *args)
+  @rainbow ||= Rainbow.new.tap do |rainbow|
+    rainbow.enabled = false unless $stdout.tty?
+  end
+
+  @rainbow.wrap(text).color(*args)
+end
+
+strict = %w[-s --strict].include?(ARGV.first)
```
wfleming: I realize this command does not currently recognize other flags, but depending on it being the first arg doesn't seem awesome.

codeclimate-validate-config/bin/run
```diff
@@ -0,0 +1,78 @@
...
+validate = CC::ValidateConfig.new(strict)
+validate.errors.each { |error| $stderr.puts colorize("ERROR: #{error}", :red) }
+validate.warnings.each { |warning| $stderr.puts colorize("WARNING: #{warning}", :red) }
+validate.nested_warnings.each do |(key, _), warning|
+  $stderr.puts colorize("WARNING in #{key}: #{warning}", :red) if key
+end
+
+if validate.valid?
+  exit 0
```
wfleming: I don't think I'm a fan of losing the `VALID_CONFIG_MESSAGE` (at least I think it's gone?): without it, the happy path has no output at all, which could be confusing to end users.

codeclimate-validate-config/bin/run
wfleming: Our option's recognition is kind of an ad-hoc mess in general. Maybe we should actually use something like [OptionParser](http://ruby-doc.org/stdlib-2.0.0/libdoc/optparse/rdoc/OptionParser.html)?

codeclimate-help/bin/run
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+underline() { printf "\e[4m%s\e[0m" "$*"; }
+
+cat <<EOM
```
wfleming: This take is probably fine for now, but right now I think one of the important missing features in the CLI is support for `subcommand --help` or `help subcommand`. I think supporting both of those is important.

Actually I guess this script could implement the latter one by just...running `subcommand --help`, so I guess it's really just the former case we should be addressing.

wfleming:
I'm a :+1: on the philosophy behind the RFC and this POC. I'm being nitpicky since if we're going to rearchitect (autocorrect suggests: "rear hotels") the early stages are the time to discuss.

Two somewhat broader concerns:

1. Is now really the appropriate time to do this? We already have other big pieces of work kind of stranded in-progress (I'm thinking of errored snapshots in particular), so I'm not sure we should open a new surgery patient before we finished sewing up a few of the current ones. This is a cool direction, but personally I don't feel like the current code is really blocking urgent work.

2. I think there's such a thing as having the code be *too* separate. My comment on the Gemfile relates to this feeling. To go into that that in a bit more detail, I'm not convinced that sub commands should share 0 code (as the RFC currently talks about). I think inevitably there will be utility functions or small classes that should be shared (I currently have a branch to work on verbose logging for debugging that would fall under this category). That code could be extracted into a separate gem, or it could live at a top-level here. Maybe it's more a packaging concern than anything else.

codeclimate-validate-config/Gemfile.lock
pbrisbin: This sub-Gemfile was more of a convenience thing: I could `cd codeclimate-validate-config` and suddenly find myself working in a much simpler, but still complete project tree.

The Ruby environment (in the container) is a shared global thing, created by the single top-level `bundle` in the Dockerfile. We could definitely figure things out one way or another, but for now I'd chalk this up to a non-interesting packaging concern.

codeclimate-validate-config/bin/run
pbrisbin: One of the biggest appeals of the isolated sub-commands is that they are free to make their own choices about dealing with whatever level of complexity they happen to be at. While `codeclimate` itself has long passed the bar for bringing in a real options parser (IMO), I don't think this command and its single flag meets it yet.

Also, this `--strict` idea might not even land (we might decide to preserve the "warnings are always fatal" behavior), so I didn't want to spend additional effort supporting it here.

codeclimate-validate-config/bin/run
pbrisbin: Just an oversight; I didn't notice that behavior when porting the code. I'll bring it back and add a spec if/when we move forward on this.

codeclimate-help/bin/run
pbrisbin: Agreed on all counts, most of all: *This take is probably fine for now* :)

pbrisbin:
> Is now really the appropriate time to do this?

The RFC and the PoC aren't meant to trigger immediate action on this. Just to get the discussion started and set a general direction for if/when anything does get actioned. There was a reason I opened these PRs over vacation :)

> I think there's such a thing as having the code be too separate. ... I'm not convinced that sub commands should share 0 code (as the RFC currently talks about).

I think it's important to start with that rule. If we come across a case where the 0-share rule is causing pain, we can discuss that in concrete terms and *only then* come up with the right solution.

I think Ruby in general, and our codebases in specific, have sufferred from over-DRYing. Whenever I've tried to pre-DRY (`CC::Analyzer::*`, `CC::Kafka::*`) it's turned out very poorly. Whenever I've taken the "duplicate, then extract" approach (`StructuredLogger`, `KafkaOffset`), I've been generally much happier.

> I think inevitably there will be utility functions or small classes that should be shared (I currently have a branch to work on verbose logging for debugging that would fall under this category).

I'd be curious to see how that shakes out. Do you really need complex logging in every, or even most sub-commands? I think you could get a long way with only that feature existing in `codeclimate-engines`.

> That code could be extracted into a separate gem, or it could live at a top-level here. Maybe it's more a packaging concern than anything else.

I do think we could look to extract *general* gems (i.e. not in any way codeclimate specific), `StructuredLogger` might be a good candidate. Having something live top-level and be loaded into the sub-commands that need it seems like a reasonable middle ground. We might end up with something shared to help testing too, for example.

In short, there's a lot of wiggle room from the "rule" I'm outlining to begin with. My take would be to start with the rule and consider if/when/how to break it, rather than start with no rule because we suspect it'll be too constrictive.

(Meta note: ^ this discussion probably should've been in the RFC PR)

pbrisbin:
Going to close this for now. There's good discussion here and it's linked back to from the RFC PR, so it's done its job. I think further discussion should take place in the RFC anyway.

If/when we're ready to move forward on this, I will re-open.

PR #295, opened 2016-01-12
tweak eslint defaults: complexity, max-statements

wfleming:
Trying to match up the defaults here to what we have for RuboCop so we
can have a bit more consistency across languages.

This changes the triggering complexity threshold to 6 (same as in our
default .rubocop.yml in this repo), and max-statements (which is subtly
different from max length, but similar enough I think we can use the
same val for now) to 30 (which is RuboCop's internal default: we don't
set it explicitly in our `.rubocop.yml`).

:eyes: @codeclimate/review


jpignata:
LGTM.

Issue #297, opened 2016-01-14
Per-engine exclude_paths options


KevinBongart:
(originally posted here: https://github.com/codeclimate/codeclimate-duplication/issues/74#issuecomment-171466949, now moving the issue to the correct repo)

Hi, is there a way to configure the engine to ignore some paths? There is a global Code Climate setting:

```yml
---
engines:
  duplication:
    enabled: true
    config:
      languages:
      - ruby
  rubocop:
    enabled: true
exclude_paths:
- spec/*
```

But I still want other engines to analyze that spec directory, so I tried both:

```yml
---
engines:
  duplication:
    enabled: true
    config:
      languages:
      - ruby
      exclude_paths:
      - spec/*
  rubocop:
    enabled: true
```

and:

```yml
---
engines:
  duplication:
    enabled: true
    exclude_paths:
    - spec/*
    config:
      languages:
      - ruby
  rubocop:
    enabled: true
```

…but no success. Let me know if I'm doing something wrong, or if this is a future feature!

ABaldwinHunter:
@KevinBongart Thanks again for your suggestion and moving the issue to this repo! 

This feature has been on our radar. 

I look forward to reporting updates as soon as there's progress.

dblandin:
Hey @KevinBongart,

#308 was merged and released a couple weeks ago. As we mentioned in #298, engine-level `exclude_paths` are now supported by the CLI.

Going to go ahead and close this issue. Thanks for your feedback and nudge towards implementing this feature!

PR #298, opened 2016-01-14
Exclude paths for a specific engine

KevinBongart:
re: https://github.com/codeclimate/codeclimate/issues/297

Attempting to add an `exclude_paths` option for specific engines like so:

```yml
---
engines:
  duplication:
    enabled: true
    config:
      languages:
      - ruby
      exclude_paths:
      - spec/*
  rubocop:
    enabled: true
```

~~Currently fails inside engine~~ update: no longer!

Where do I go from here?

### DIFF
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -33,7 +33,7 @@ module CC
 
       def engine_config(raw_engine_config)
         config = raw_engine_config.merge(
-          exclude_paths: exclude_paths(raw_engine_config.config.fetch("exclude_paths")),
+          exclude_paths: exclude_paths(raw_engine_config.fetch("config", {})["exclude_paths"]),
           include_paths: include_paths,
         )
         # The yaml gem turns a config file string into a hash, but engines
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -32,9 +32,11 @@ module CC
       attr_reader :include_paths
 
       def engine_config(raw_engine_config)
+        all_paths_to_exclude = exclude_paths(raw_engine_config.fetch("config", {})["exclude_paths"])
+
         config = raw_engine_config.merge(
-          exclude_paths: exclude_paths(raw_engine_config.fetch("config", {})["exclude_paths"]),
-          include_paths: include_paths,
+          exclude_paths: all_paths_to_exclude,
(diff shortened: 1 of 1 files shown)


KevinBongart:
I didn't read the instructions, but after running `make image`, I finally got the engines to run! However, while I can confirm the `engine_config` lists my engine-specific `exclude_paths`, the paths don't seem to be excluded from the analysis.

Any pointers would be helpful!

ABaldwinHunter:
@KevinBongart Thanks for your work!

Couple of things:

1. passing engines the `exclude_paths` config key is [deprecated](https://github.com/codeclimate/spec/blob/master/SPEC.md#input) in favor of `include_paths`, which get derived from the `exclude_paths` in a user's `.codeclimate.yml`. Passing your updates through the `include_paths` key instead should have the desired effect. 
2.  **Performance**: Building the `include_paths` has proven slow at times, and we have been thinking about a way to make that process more performant. Building them fresh for each engine - to accommodate per-engine excludes - with the current implementation could be inhibitive in terms of speed.

We're open to any suggestions! If you'd like to see the per-engine excludes locally, getting your change through the `include_paths` key should do the trick.

KevinBongart:
@ABaldwinHunter Thanks for the feedback, I understand why this wasn't working now!

I tweaked it to also update `include_paths` and it worked. Not the most elegant code or variable names, so if you have suggestions that would be quick to fix. If the direction is right, I can also write some specs.

At that point, would the feature make its way to master?

KevinBongart:
@ABaldwinHunter @GordonDiggs Any interest in getting this merged?

ABaldwinHunter:
@KevinBongart Thanks for following up!

There's actually a related Pull under review now that revamps the way we organize file paths for analysis:
https://github.com/codeclimate/codeclimate/pull/308/files

wfleming:
HI @KevinBongart,

Sorry for the lack of feedback here recently. As @ABaldwinHunter said, we've reorganized how we build the workspace for analysis in #308. We had an internal implementation of per-engine `exclude_paths` that we'd been holding off on because of the performance issues with the previous workspace-management, and we ended up including that work in #308, so I think we're going to move forward with that: sorry about that. If you have feedback on that PR, we'd be happy to discuss details of the implementation there.

KevinBongart:
@ABaldwinHunter @wfleming No problem, my implementation was mostly a workaround, I can definitely see the performance implications. Glad to see this feature making its way into the CLI!

Closing this one then :+1:

PR #300, opened 2016-01-15
Drastically simplify workspace building

pbrisbin:
/cc @codeclimate/review

This commit removes IncludePathsBuilder and all classes involved in it. In its
place, we have CC::Workspace (new terminology) with drastically simpler
semantics. This simpler approach should be faster and more reliable, at the
expense of less features.

We drop support for glob-style patterns. Instead we support a simple list of
files or directories to ignore. We stop sending an exclude_paths configuration
key in /engines.json, only include_paths now. We stop incorporating .gitignore
in workspace-building.

Reasoning:

- Paths, not patterns.

  We see little actual value in patterns support. We leave it up to the engine
  to know that it should (for example) only target .py files, therefore we
  don't have to exclude **/*.pyc from the workspace.

  We've found the vast majority of our users have just been trying to exclude
  directories. They're actually quite frustrated having to shoe-horn it into
  patterns, only to be met with slowness or frequent bugs as we needlessly
  expand and filter these patterns in complicated ways.

  Old:

      exclude_paths:
      - .node_modules/**/*
      - app/assets/vendor/**

  New:

      exclude_paths:
      - .node_modules
      - app/assets/vendor

  We minimize disruption to these users by transparently handling a pattern
  such as foo/bar/\*\* or foo/bar/\*\*/\*\* as the directory foo/bar. Other patterns,
  such as foo/\*.pyc will generate a warning and be ignored.

- Ignore .gitignore

  The motivation for this feature is unclear, so I'm testing the waters about
  dropping it as it simplifies things.

- Stop all the permissions checking

  If the current directory contents, plus the user's defined excludes causes a
  permissions error in the workspace, it should be treated as an actual error:
  the user must modify the permissions or add the unreadable file to excludes.

Additional feature:

- Per-engine excludes

  The algorithm supports repeated filtering. That is, we can build a workspace
  filtered from top-level exclude_paths once, then filter that further for each
  engine with per-engine-excludes.

Caveats:

The "explicit includes" feature (i.e. codeclimate analyze [FILE, ...]) has an
interesting interaction with this approach. Specifying files in this way
effectively means we begin building a workspace from these files rather than
the current directory's contents. This means excludes will still apply on top
of the passed files. This may not be what we want, an alternative would be to
pass along those paths as-is as include_paths and do no filtering -- I leave
the decision open, with no strong opinion of my own.

Because the passed files are compared against the exclude list, they must be
simple and relative (i.e. they cannot be "." or "./"-prefixed), as they would
come out in a * glob. This is brittle, but currently necessary. See
PathsValidator for more details.

Additionally, the way the exclude list applies on top of the given files can be
surprising. A user could exclude spec, for example, and pass spec/foo.rb. In
such a case, spec/foo.rb will be processed which may be surprising. This is
because the current exclusion check assumes we'd never even get to a file
within spec if that directory were entirely excluded, so there's no logic to
catch such a case.

I'm inclined to either simplify the feature to a simple pass-through or keep
these caveats and just document the behavior.


mrb:
@pbrisbin FWIW I think the interaction you described with single file analysis is totally fine.

ABaldwinHunter:
@pbrisbin applause for this work! I'm incredibly excited to have a revamped and simplified system for path handling.

Immediate thoughts:
The good
1) I like not worrying about permissions handling +1
2) Having `excludes` in .codelcimate.yml filter through directories passed to `codeclimate analyze` is okay and expected behavior at present

The bad
3) I'm not convinced that using paths instead of patterns is significantly more performant? and if the downside is that users can't exclude individual files, or use patterns in some cases, I don't love that.

My understanding was that the slowness came from drilling down into directories recursively. If we forget about permissions issues and forego the file-by-file tree walk, isn't that the major performance win? 

e.g. as you walk tree for exclude paths:

```
exclude_paths:
- spec/**/*
- test/**/*
```
As soon as you realize spec is a directory, add `spec` to include paths, and stop drilling down.

Only users who care enough to exclude particular files or extensions get the slower behavior.

Have you done any bench marking on different sized repos to see where the most time is spent? If the paths vs. patterns is only a minimal performance win, might not be worth it imo.

But overall, really love this simplification and happier to have the code in this state than previous one.

pbrisbin:
> I'm not convinced that using paths instead of patterns is significantly more performant

I guess no one can be convinced until I benchmark :)

Considering the old logic takes `x/**/*` and [recursively, fully expands it](https://github.com/codeclimate/codeclimate/blob/b4a06ece4235c9b7376ee6a317ec5c41a97216e8/lib/cc/analyzer/engines_config_builder.rb#L64) before doing anything else, and the new logic just looks at `x` as-is I _have to_ assume the new logic is vastly more performant. Consider that difference in behavior for something like `.node_modules`!

> As soon as you realize spec is a directory, add spec to include paths, and stop drilling down.

I'm happy to look at some code if you could implement that optimization without removing what's present, but my suspicion is that it would mean complicating already complicated bug-risky code to accomplish.

> Have you done any bench marking on different sized repos to see where the most time is spent?

I know that @wfleming has an example repo, so next week when we're both back in the office I'm going to use that for benchmarking. I may also ask the folks in the motivating Issue to try this branch.

---

I'm 50% concerned with this code because it's slow and 50% concerned with this code because it's has had numerous bugs, and the same bugs repeatedly. I want to ensure whatever solution we chose, we fix both aspects. Simplifying things greatly is the solution chosen here.

My *opinion* is that we don't need patterns: the complexity involved in supporting them is not worth it for how (not) useful they are. To back that up, I'm going to take a sampling of codeclimate.yml's on the site and try to bucket out how many would be negatively impacted by this change. Hopefully, it's very few, but we'll see!

pbrisbin:
@ABaldwinHunter FYI, I'm going to run a longer query later, but looking at repos that have analyzed the past hour, I get about ~7% using exclude paths that would be ignored by this new logic. Some examples are:

```
invalid exclude: **/vendor/*
invalid exclude: **/vendor/**/*
invalid exclude: client/tests/**/*.js
invalid exclude: *.md
invalid exclude: app/**/*.liquid
invalid exclude: module/**/config/*
invalid exclude: app/assets/javascripts/**/*bundle.js
invalid exclude: server/**/*.spec.js
invalid exclude: **.jpg
invalid exclude: **/*.scss
invalid exclude: config/*.rb
invalid exclude: components/ui/react/app/assets/javascripts/react/*_bundle.js
invalid exclude: components/ui/react/app/assets/javascripts/react/*_bundle.js
invalid exclude: */vendor/assets/**
invalid exclude: ci/*.js
invalid exclude: *.swp
invalid exclude: app/*/i18n/**/*.js
invalid exclude: engines/**/vendor/**/*
invalid exclude: public/js/*.bundle.js
invalid exclude: */__tests__/*
invalid exclude: **/*.test.js
invalid exclude: server/**/migrations/*.py
invalid exclude: public/javascripts/*.min.js
invalid exclude: app/*.php
invalid exclude: src/app/**/*.spec.js
invalid exclude: app/assets/javascripts/**/*.haml
invalid exclude: src/js/*jquery.js
invalid exclude: */tests.py
invalid exclude: lib/seed_models/**/*.rb
invalid exclude: **/spec/**/*
invalid exclude: **/*.css
invalid exclude: app/assets/javascripts/**/routes/*
invalid exclude: lib/**/*.rb
invalid exclude: app/assets/javascripts/**/*.jsx
invalid exclude: **/tools/*
invalid exclude: **.jpg
invalid exclude: app/modules/*/test/*
invalid exclude: migrations/versions/*.py
invalid exclude: .*.yml
invalid exclude: apps/*/migrations/*
invalid exclude: */vendor/**/*
```

Food for thought.

ABaldwinHunter:
@pbrisbin Thanks. Great food for thought!

> I'm happy to look at some code if you could implement that optimization without removing what's present, but my suspicion is that it would mean complicating already complicated bug-risky code to accomplish.

That could be so. Going to think about an implementation. I had in mind originally our first tree walking algorithm, which had a [good seam for checking if a path was a directory](https://github.com/codeclimate/codeclimate/blob/7dc2c391128ada61ccfb148870de89404d001aef/lib/cc/analyzer/include_paths_builder.rb#L37).

> I'm 50% concerned with this code because it's slow and 50% concerned with this code because it's has had numerous bugs, and the same bugs repeatedly. I want to ensure whatever solution we chose, we fix both aspects. Simplifying things greatly is the solution chosen here.

:+1:  I'm on board with this sentiment.  Simple code with some loss  > bedbugs.

lib/cc/workspace.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    private
+
+    attr_reader :prefix
+
+    def expand(name, exclusions)
+      path = "#{prefix}#{name}"
+
+      return [] if %w[. .. .git].include?(name)
```
GordonDiggs: This series of short-circuited returns is hard to read to me. WDYT of using `if...else`?

lib/cc/workspace.rb
```diff
@@ -0,0 +1,66 @@
...
+      CLI.debug("workspace filter recurse", prefix: prefix, path: name)
+
+      Dir.chdir(name) do
+        workspace = self.class.new(prefix: dir(path))
+        workspace.filter(exclusions.exclude_paths).paths
+      end
+    end
+
+    def dir(path)
+      "#{undir(path)}#{File::SEPARATOR}"
```
GordonDiggs: Does this just take the separator off the end and add it back?

lib/cc/workspace/exclusions.rb
```diff
@@ -0,0 +1,47 @@
...
+        if normalized.include?("*")
+          $stderr.puts(invalid_exclude_path_warning(pattern))
+        else
+          normalized
+        end
+      end
+
+      def invalid_exclude_path_warning(pattern)
+        <<-EOM
+WARNING: invalid exclude path: #{pattern.inspect}.
```
GordonDiggs: Is there a solution for this situation on `.com`?

lib/cc/workspace.rb
pbrisbin: I thought this was much more readable than that, myself. Maybe a third opinion?

lib/cc/workspace.rb
pbrisbin: Yes, in some cases.

`path` can enter this method as `foo` or `foo/`, depending on circumstances. Therefore, without the remove-and-readd, you'd get `foo/` and `foo//` as possible outputs. With the remove-and-readd, you get `foo/` and `foo/` as desired.

If you like, I could make the branches explicit, something like:

```rb
if path.end_with?(File::SEPARATOR)
  path
else
  "#{path}#{File::SEPARATOR}"
end
```

GordonDiggs: Nope - this makes sense to me as-is now!

lib/cc/workspace/exclusions.rb
pbrisbin: Can you unpack "this situation"?

GordonDiggs: A `.codeclimate.yml` with pattern-based exclusions

pbrisbin: My intent would be that the pattern is ignored and the warning is visible. That's not dot-com specific though, and assumes this `stderr` shows up on the build screen, which is may not.

But you might mean something else?

lib/cc/cli.rb
```diff
@@ -15,5 +16,16 @@ module CLI
     autoload :Test, "cc/cli/test"
     autoload :ValidateConfig, "cc/cli/validate_config"
     autoload :Version, "cc/cli/version"
+
+    def self.debug(message, values = {})
```
wfleming: Glad to see this extracted :+1:.

wfleming:
I'm excited by the re-architecture & the overall improvements! All kinds of awesome.

Similar to @ABaldwinHunter, I am concerned about completely eliminating support for globs. I think there are enough examples you found in your own research where a path equivalent would not really work, and I can think of at least one use-case that I don't see represented, which is Go projects: if you want exclude tests (a fairly common use-case, I think), it would be something like `**/*_test.go`: the alternative is effectively tracking every single file in your project in your excludes. I see similar cases in some of your examples, so I think completely dropping support for globs would be too drastic, and dropping support for something that is seen as a pretty basic feature for scanning tools in general by a wide swath of devs.

I think we can support globs while continuing to keep the architectural & performance gains you've made here. I've drafted a working implementation in #302 based off of this work, which adds back glob support but does *not* add back any of the things I considered troublesome in the old implementation.

lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -56,20 +58,16 @@ def names_and_raw_engine_configs
...
 
-      def gitignore_paths
-        if File.exist?(".gitignore")
-          `git ls-files --others -i -z --exclude-from .gitignore`.split("\0")
-        else
-          []
+      # The yaml gem turns a config file string into a hash, but engines expect
+      # the string. So we (for now) need to turn it into a string in that one
+      # scenario.
+      def normalize_config_file(config)
```
wfleming: Is this a related change or related to https://github.com/codeclimate/codeclimate-rubocop/issues/39? Feels like it should be extracted unless there's a closer connection to the workspace work I'm missing.

wfleming: Actually, thinking on it, it feels like this fix belongs in cc-yaml: if we intend to allow `config: file-name`, then `cc-yaml` should not presume the type of the `config` key.

wfleming:
Closing to re-open with the glob work & updated things.

PR #301, opened 2016-01-16
Properly namespace exception so that the constant can be found

noahd1:
Previously:

```
$ codeclimate analyze -f asdf
error: (NameError) uninitialized constant CC::Analyzer::Formatters::InvalidFormatterError
```

Now:

```
$ codeclimate analyze -f asdf
'asdf' is not a valid formatter. Valid options are: json, text
```


wfleming:
:shipit:

PR #302, opened 2016-01-18
Restore Glob Functionality to Workspace Refactor

wfleming:
I'm excited about the direction of #300, but I was concerned that completely removing glob support was overly extreme, and I'd had some thoughts while I was traveling about how it might be done performantly. This PR is my first draft at implementing those thoughts building on top of @pbrisbin's work.

The key differences with our old technique are as follows:
* Directory contents are never walked until they need to be due to explicit inclusions or exclusions. As a result:
    * The algorithm is much simpler than before: instead of a several-step process of collecting several disparate collections of paths, operating on relations between them, and then minimizing the resulting set, all operations are on a single sparse tree which is naturally minimized at all times.
   * Less overall file I/O & memory allocation for very large projects, since we're not walking every directory & building an in-memory list of every file path.
* Only things that look like globs are treated like globs. Excluding directories by name instead of pattern works just fine.
* There is no attempt to deal with unreadable files or symlinks or anything like that. (I think we'll still need to handle the case where a file we need to touch for interpreting an exclude can't be read, though).

I went out of my way to keep the existing specs from #300 mostly unchanged, except for the sole case of no longer warning on glob usage (obviously), and adding a spec for globs working. I tried to minimize changing existing bits as well to make this easier to review & keep it clearer what I was changing. So this is more of a PoC or first draft than a finished proposal of implementation. That said, the basic structure is pretty much what I had in mind.

As far as performance is concerned, this implementation works quite well on my local pathological case: it's indistinguishable from the current state of #300 as far as experienced performance goes. I'd like to write some actual code-based benchmarks before this ships so we'll have actual tests to tell us if the performance gets bad due to future changes.

:eyes: @codeclimate/review 

---------

TODOs:

- [ ] This work means `Workspace#dup` is no longer safe for multiple-engines, because the dup is using the same instance of `PathTree` as the original. I'll need to fix that & should also update the relative workspace spec to actually test the dup functionality rather than just multiple `#filter` calls.
- [ ] Handle potentially unreadable files when walking.
- [ ] Do symlinks need any special treatment?
- [ ] `.git` should be auto-excluded: that needs to be added.
- [ ] Write some benchmarks.
- [ ] make Specs test/use the results of a CC::YAML parse for both this class & Workspace.

### DIFF
--- a/lib/cc/workspace.rb
+++ b/lib/cc/workspace.rb
@@ -1,6 +1,7 @@
 module CC
   class Workspace
     autoload :Exclusions, "cc/workspace/exclusions"
+    autoload :PathTree, "cc/workspace/path_tree"
     autoload :PathsValidator, "cc/workspace/paths_validator"
 
     def initialize(paths: nil, prefix: "")
--- a/lib/cc/workspace.rb
+++ b/lib/cc/workspace.rb
@@ -4,21 +4,22 @@ module CC
     autoload :PathTree, "cc/workspace/path_tree"
     autoload :PathsValidator, "cc/workspace/paths_validator"
 
-    def initialize(paths: nil, prefix: "")
-      @prefix = prefix
+    DEFAULT_PATH = "."
 
+    def initialize(paths: nil)
+      @path_tree = PathTree.new(DEFAULT_PATH)
       if paths.present?
         validator = PathsValidator.new(paths)
         validator.run
 
-        @paths = paths
+        path_tree.include_paths(paths)
       end
 
-      CLI.debug("workspace initialize", prefix: prefix)
+      CLI.debug("workspace initialize")
     end
 
     def paths
-      @paths || ["./"]
+      path_tree.all_paths
     end
 
     def filter(exclude_paths)
@@ -28,8 +29,13 @@ module CC
 
       CLI.debug("workspace filter start", exclude_paths: exclude_paths)
 
-      @paths ||= Dir.glob("*", File::FNM_DOTMATCH)
-      @paths = @paths.flat_map { |name| expand(name, exclusions) }
+      exclusions.exclude_paths.each do |exclusion|
+        if exclusions.glob?(exclusion)
+          path_tree.exclude_paths(exclusions.expanded_glob(exclusion))
+        else
+          path_tree.exclude_paths([exclusion])
+        end
+      end
 
       CLI.debug("workspace filter end", paths: paths)
 
@@ -38,30 +44,6 @@ module CC
 
     private
 
-    attr_reader :prefix
-
-    def expand(name, exclusions)
-      path = "#{prefix}#{name}"
-
-      return [] if %w[. .. .git].include?(name)
-      return [] if exclusions.include?(undir(path))
-      return [path] unless File.directory?(name)
-      return [dir(path)] unless exclusions.apply?(undir(path))
-
-      CLI.debug("workspace filter recurse", prefix: prefix, path: name)
-
-      Dir.chdir(name) do
-        workspace = self.class.new(prefix: dir(path))
-        workspace.filter(exclusions.exclude_paths).paths
-      end
-    end
-
-    def dir(path)
-      "#{undir(path)}#{File::SEPARATOR}"
-    end
-
-    def undir(path)
-      path.sub(/#{File::SEPARATOR}$/, "")
-    end
+    attr_reader :path_tree
   end
 end
--- a/lib/cc/workspace/exclusions.rb
+++ b/lib/cc/workspace/exclusions.rb
@@ -7,40 +7,18 @@ module CC
         @exclude_paths = exclude_paths.map { |p| normalize(p) }.compact
       end
 
-      def include?(path)
-        exclude_paths.include?(path)
+      def glob?(exclusion)
+        exclusion.include?("*")
       end
 
-      def apply?(path)
-        exclude_paths.any? do |exclude_path|
-          exclude_path.start_with?(path)
-        end
+      def expanded_glob(exclusion)
+        Dir.glob(exclusion)
       end
 
       private
 
       def normalize(pattern)
-        normalized = pattern.to_s.sub(%r{/\*\*(/\*)?$}, "")
-
-        if normalized.include?("*")
-          $stderr.puts(invalid_exclude_path_warning(pattern))
-        else
-          normalized
-        end
-      end
-
-      def invalid_exclude_path_warning(pattern)
-        <<-EOM
-WARNING: invalid exclude path: #{pattern.inspect}.
-
-Unfortunately, we no longer support wildcard patterns in exclude_paths, only
-file paths or directories relative to the project root. This pattern will be
-ignored.
-
-If you feel particularly limited by this, please open an Issue at
-https://github.com/codeclimate/codeclimate and we'll try to find a
-non-exclude_paths-based solution.
-        EOM
+        pattern.to_s.sub(%r{/\*\*(/\*)?$}, "")
       end
     end
   end
--- a/lib/cc/workspace/path_tree.rb
+++ b/lib/cc/workspace/path_tree.rb
@@ -24,10 +24,10 @@ module CC
 
       protected
 
-      attr_reader :child_dirs, :root_path, :child_files
+      attr_reader :root_path
 
       def populated?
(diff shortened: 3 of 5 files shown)


wfleming:
TODO for myself: this work means `Workspace#dup` is no longer safe for multiple-engines, because the dup is using the same instance of `PathTree` as the original. I'll need to fix that & should also update the relative workspace spec to actually test the dup functionality rather than just multiple `#filter` calls.

wfleming:
Open Questions:

* @pbrisbin [removed the type check stuff for YAML's globs](https://github.com/codeclimate/codeclimate/pull/300/files#diff-44359e6b88d1752879de5f9b86729151L38) That may need to come back in some way?
* Unclear if this would need to change anything in `PathsValidator`, but I don't think it would: any glob given for include paths on the command line should be expanded by the shell before the CLI gets them.

pbrisbin:
> I think we'll still need to handle the case where a file we need to touch for interpreting an exclude can't be read, though

Can you expand on this?

> @pbrisbin removed the type check stuff for YAML's globs That may need to come back in some way?

Probably. I think we need to do two things:

- Build the `exclude_paths` argument in the specs from actual `CC::Yaml.parse(x).exclude_paths`, as it would be in production
- Add/verify spec examples for our transparent handling of `/**`-style globs, which we intend to continue to support

> Unclear if this would need to change anything in PathsValidator

Probably not. That class annoys me and I'd like to continue thinking about ways to remove it, but if it stays it can probably stay as-is across this work.

wfleming:
>> I think we'll still need to handle the case where a file we need to touch for interpreting an exclude can't be read, though
>
> Can you expand on this?

There could conceptually be a directory that is not readable by us, but is within the project & gets hit by an exclude path. Since this class works by descending through directories referenced by exclude/include paths, that would result in an exception. I'm suggesting we'll want a `catch` and a  `CLI.debug()` for that case.

> Add/verify spec examples for our transparent handling of /**-style globs, which we intend to continue to support

Can you expand on *this*? I think current spec coverage does cover this in that those style globs are shown being used?

This PR also means we would support all globs that Ruby supports, not just trailing-** globs. I tried to explain why I think that's worthwhile on #300. FWIW, in my own testing, it was not simply globbing that was performance intensive, but  the fact that we were doing far too much extraneous work on top of globbing & that our algorithms were poor. With this PR & #300, globs that can be simplified (trailing-**) are simplified & are not expanded, but globs that can't be simplified are fully expanded. The performance of this approach has been perfectly fine in my local tests, but I do want to add more strenuous benchmarks here as tests as well.

pbrisbin:
> I think current spec coverage does cover this in that those style globs are shown being used?

Yes, we (probably) do have adequate spec coverage here; thus I hedged with *Add/verify*, and didn't just say *Add* or *verify* in my suggestion.

This important point is that we shouldn't be passing in a raw Array in the specs, because that's not what the object would receive in the wild, and the difference is important to this behavior. We should pass `CC::Yaml.parse("--content--").exclude_paths`. As long as we do that, and the spec cases use both styles, we'll have our answer as to what we need to do implementation-wise (maybe nothing?)

lib/cc/workspace.rb
```diff
@@ -27,8 +29,13 @@ def filter(exclude_paths)
...
 
-      @paths ||= Dir.glob("*", File::FNM_DOTMATCH)
-      @paths = @paths.flat_map { |name| expand(name, exclusions) }
+      exclusions.exclude_paths.each do |exclusion|
+        if exclusions.glob?(exclusion)
+          path_tree.exclude_paths(exclusions.expanded_glob(exclusion))
+        else
+          path_tree.exclude_paths([exclusion])
+        end
+      end
```
pbrisbin: Now that the complicated part has shifted from the list itself to the individual patterns, I don't think `Exclusions` is what we want. We probably want `[Exclusion]` here:

```rb
def filter(patterns)
  patterns.each do |pattern|
    exclusion = Exclusion.new(pattern)

    if exclusion.glob?
      path_tree.exclude_paths(exclusion.expand)
    else
      path_tree.exclude_paths([exclusion.raw]) # or something
    end
  end
end
```

Then my next suggestion would be to resolve the TDA violation by pushing `glob?` into `expand`

```rb
class Exclusion
  def expand
    return [pattern] unless glob?

    # expansion logic...
  end
end
```

Then my *next* suggesion is... do we really need both `Workspace` and `PathTree`? At that point, `Workspace` isn't doing too much, and the indirection may not be worth it.

pbrisbin:
I didn't review super closely, I think we should merge the branches, close #300 and this, and open a new PR with updated description, centralized TODO list, and start a fresh discussion.

wfleming:
> This important point is that we shouldn't be passing in a raw Array in the specs, because that's not what the object would receive in the wild, and the difference is important to this behavior. We should pass CC::Yaml.parse("--content--").exclude_paths. As long as we do that, and the spec cases use both styles, we'll have our answer as to what we need to do implementation-wise (maybe nothing?)

Gotcha. Agreed. I'll add it to the TODO list above.

pbrisbin:
> There could conceptually be a directory that is not readable by us, but is within the project & gets hit by an exclude path. Since this class works by descending through directories referenced by exclude/include paths, that would result in an exception

Since the CLI runs as uid 0 (i.e. root), I doubt we'll see permissions exceptions here. The main source of these in the past was engines (which run as uid 9000) hitting such a scenario because they were doing their own version of workspace building.

My gut is that moving to `include_paths` pretty much solved all permissions issues and any such logic is just vestigial or was implemented because of misunderstanding. Do you believe otherwise?

wfleming:
> My gut is that moving to include_paths pretty much solved all permissions issues and any such logic is just vestigial or was implemented because of misunderstanding. Do you believe otherwise?

I do, though this scenario may be limited to `docker-machine`/OS X users. If I create a directory that my user cannot read locally, and then mount that directory as a volume in a docker container where I am root, I get a permission error trying to `ls` the directory.

wfleming:
> I didn't review super closely, I think we should merge the branches, close #300 and this, and open a new PR with updated description, centralized TODO list, and start a fresh discussion.

I'm cool with that.

pbrisbin:
> If I create a directory that my user cannot read locally

That sounds highly unrealistic to me, and is not an example of the sort of permissions issues i remember plaguing us in the past.

wfleming:
> That sounds highly unrealistic to me, and is not an example of the sort of permissions issues i remember plaguing us in the past.

I thought those were exactly the sort of permissions issues we'd seen in the past: files existed underneath the project directory that were not readable by the running user for some weird workflow reason.

pbrisbin:
> files existed underneath the project directory that were not readable by the running user

If you could find an example of that, I'd love to see it. My recollection is that `root`-owned or `0700`-moded files were the issue. The [rfc](https://github.com/codeclimate/rfc/pull/6/files#diff-edf246be8d22b04ae24171c7b92ad35cR74) is a little thin on details, but does mention the `root`-owned issue and does not mention anything about a user having a file they themselves can't read, which --even if we see a case in the wild-- I'd argue we shouldn't support.

pbrisbin:
Wait. I'm dumb -- clearly `root`-owned is a file they themselves can't read. Sorry.

wfleming:
@pbrisbin are you running `root` as your login account on your laptop or something? ;)

pbrisbin:
Still though:

- Are you sure unreadable *directory* is a common case, not just file(s)?
- Can't, and shouldn't, such a case be solved by the user excluding the directory, which under your new algorithm should actually work?

wfleming:
> Are you sure unreadable directory is a common case, not just file(s)?

No, I'm not. I don't know how common either case is, I'm just going off the memory that root-owned files/directories were a problem.

> Can't, and shouldn't, such a case be solved by the user excluding the directory, which under your new algorithm should actually work?

Yes, it should work you're right.

I was thinking of a case where a glob needing expanding ended up matching an unreadable directory/file, but I've now realized that can't happen, since `Dir.glob` handles unreadable dirs itself.So I think it would have to be a situation where a user tried to exclude something *underneath* an unreadable dir, which I agree probably qualifies as user error.

So I'm ok with not handling the unreadable thing.

pbrisbin:
Woot!

Since you're code here seems to be the bulk of it. Do you mind handling the merge-close-reopen dance?

wfleming:
Sure, I can take that.

Issue #246, opened 2015-12-03
Add a function to delete the temp folder.


envygeeks:
At this current time you create a temp folder and then hold it as root, for systems that remain on for an extended period of time, this folder can grow with the number of projects tested, since you keep the folder as root I think it would be nice if you add a function to delete the files.

pbrisbin:
Hi @envygeeks sorry for the late response here. You're right, claiming `/tmp/cc` was a poor decision we made early on and we'd like to correct it. Unfortunately, we just haven't been able to prioritize a solution just yet. Hopefully I'll be able to report back soon when we do.

dblandin:
Hey @envygeeks, did #303 (now merged) address your concerns?

envygeeks:
@dblandin I haven't check yet but I will be sure to check it out today.  However, as long as the /tmp/cc folder is gone then all complaints are address from this end so it's safe to close this!

dblandin:
@envygeeks We're still using the host's `/tmp/cc` directory so that we can write out a config file used by the next engine container. The change made in #303 will ensure that the config file we create is removed.

No existing files within the host's `/tmp/cc` directory are accessed and we remove our own temporary file, but we do mount the `/tmp/cc` directory, create it it necessary, and leave it on the system.

You can see the volume mount within our [codeclimate-wrapper](https://github.com/codeclimate/codeclimate/blob/315f67791edb530378bc313eb59be04a5e94a2d1/codeclimate-wrapper#L57) script.

I'm guessing that we're using the specific `/tmp/cc` path to avoid volume mounting the entire `/tmp` directory. Bumping the volume mount up a level to `/tmp` would allow us to create a unique temp directory and completely remove it before exiting but would also expose more of the user's files/directories to the codeclimate container, which we'd like to avoid.

dblandin:
Hey @envygeeks,

> However, as long as the /tmp/cc folder is gone then all complaints are address from this end so it's safe to close this!

As of [v0.22.1](https://github.com/codeclimate/codeclimate/releases/tag/v0.22.1), we now support a `CODECLIMATE_TMP` environment variable to specify a custom host directory for temporary files during Code Climate analyses.

Now that this enhancement is in place, I'm going to go ahead and close this issue. Thanks!

Issue #245, opened 2015-12-03
Support for a custom /tmp folder definition.


envygeeks:
`/tmp/cc` is a terrible name to use by default and ask the host to hold.  It's generic and clashes on our systems where we already have a "/tmp/cc" folder for other stuff.  It would be nice if we could get support for a custom temp folder, and if there already is, have it documented.

pbrisbin:
Hi @envygeeks , sorry about this. I agree it's a problem that we hope to address soon. I'll report back when I have a workaround or actual solution.

dblandin:
Hey @envygeeks,

We recently released [v0.22.1](https://github.com/codeclimate/codeclimate/releases/tag/v0.22.1) of the Code Climate CLI which includes @pbrisbin's updates in #369 regarding mounted code and tmp paths. You can now use a `CODECLIMATE_TMP` environment variable to specify a custom host directory for temporary files during Code Climate analyses.

Now that this enhancement is in place, I'm going to go ahead and close this issue. Thanks!

PR #303, opened 2016-01-19
Delete temporary file after the engine has ran.

lucasmazza:
A possible fix for #245 e #246.

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -41,7 +41,10 @@ module CC
           end
         end
 
+        write_config_file
         container.run(container_options)
+      ensure
+        delete_config_file
       end
 
       private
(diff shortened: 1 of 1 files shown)


GordonDiggs:
@lucasmazza here's the tricky bit about this (and why we initially went with a hardcoded path): The path within this `docker run` invocation must exist on the host machine as well (see where we mount it as a volume [here](https://github.com/codeclimate/codeclimate/blob/master/codeclimate-wrapper#L56)). So, I don't believe your current implementation will work, as `/tmp/foo` will exist within the `codeclimate` container, but not be accessible to the docker daemon, which is dropping back down to the host to mount the file.

lucasmazza:
@GordonDiggs I have the image a proper spin locally and it failed - makes sense to delete the file after running the container instead?

GordonDiggs:
@lucasmazza that would work too!

lucasmazza:
@GordonDiggs rebased and updated the PR title with the delete approach.

GordonDiggs:
LGTM. Thanks! Will merge on green

PR #307, opened 2016-01-22
Tweak CodeClimate Analysis Config

wfleming:
* Disable the Rails/Output rule -- this isn't a Rails app
* Turn on the FIXME engine

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,10 +2,14 @@
 engines:
   rubocop:
     enabled: true
+  fixme:
+    enabled: true
+    exclude_paths:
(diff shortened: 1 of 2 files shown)


pbrisbin:
Want to exclude `config/engines.yml` (or all of `config`)?

wfleming:
I was going to exclude it just for fixme (since that'll work pretty soon) and hold this until that work is merged.



~will out.

wfleming:
Going to merge around the CC failure: the 1 new issue is a TODO in a source file that we expect for now, but should fix.

PR #308, opened 2016-01-22
Improved Workspace Building

wfleming:
/cc @codeclimate/review 

This is the result of combining #300 & #302 and finishing up the remaining TODO items known from discussion on those 2 PRs.

The big takeaways:
* Calculation of the "workspace" (the set of paths an engine will operate on) is considerably more performant, and IMHO also conceptually simpler. Details of this were described in #302.
* There's less magic, which also means some things the CLI used to do it no longer does:
    * Ignore `.gitignore`: we believe this was fairly low value, and is not worth the complexity & potential performance penalty some edge cases incurred.
    * Stop doing an enormous amount of permissions/file type checking: if there are unreadable files locally that an engine tries to read, the engine should either handle it reasonably well, or the user should exclude those directories from analysis.
* **New feature**: per-engine excludes were made functionally trivial & entirely acceptable re: performance by this work, so they've been added.

Open question for the team: I have not yet done serious work auditing how existing engines will behave with unreadable files/directories. @pbrisbin seemed to say in #300 that errors being thrown in cases like this was acceptable, since the `exclude_paths` should be configured to ignore those things. Put another way: noisy failure is preferable to quietly ignoring something. Opinions? Is that acceptable, or should we do some tests & consider engines that don't handle this case cleanly as blocking shipping of this?

There are a couple CC complexity issues that popped up in the `PathTree` implementation: suggestions for improvements there are welcome. I've already improved that class from #302, and further improvements are not jumping out at me.

Caveats:

* Same as in #300, exclude paths are currently applied on top of explicit include paths (`codeclimate analyze FILE`), which can lead to unexpected behavior. In particular, due to how `PathTree` calculates its paths, you could end up analyzing *all* files if you attempted to explicitly analyze an excluded file. (This is because `all_paths` checks whether children were populated, and if they weren't it just returns its own root path -- including a file & then excluding it results in an empty set of children). Clearly that behavior is not correct, but before addressing that I think we should decide whether exclude paths should still apply when given explicit paths (i.e. in the case above the result should be no files to analyze), or should exclude paths be ignored when explicit paths have been provided?
* Explicit includes also have an interesting edge case where if an invalid path is specified, the actual path analyzed will end up being the parent of whichever part of the path wasn't valid. (i.e. if you specify `app/models/achtung`, and `achtung` does not exist, the entire contents of `app/models` will end up being analyzed). This is down to the recursive nature of the algorithm and the fact that the data structure is implemented as mutable (the latter may have been a mistake, but it does simplify some things). Since include paths are always given on the command line, I'm not sure if we should consider this blocking or just worthy of a note for future attention.


spec/cc/workspace/path_tree_benchmark_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  describe PathTree do
+    include FileSystemHelpers
+
+    MAX_FILTER_SECONDS = 10
+    NUM_FILES = 10_000
+    AVG_FILES_PER_DIR = 20
+    REPO_DEPTH = 30
+    NUM_EXCLUDES = 50
+
+    it "performs acceptably" do
```
wfleming: This spec may be better suited to a different `rake` task outside of `test` -- it's not a traditional spec in many ways. But since we got into our previous bad situation via a number of well-intentioned steps that led us to an unfortunate position, I wanted some kind of automated check to prevent this performance getting out of hand again.

brynary:
:clap:

With this model, if there is an unreadable file let's say `lib/urneadable.rb` that is not explicitly excluded... a) what the workspace paths provided to the engines look like? and b) are we still expecting the engine to potentially raise an exception if/when it tries to read it? (I think this is actually _desirable_, for the record)

pbrisbin:
> what the workspace paths provided to the engines look like? 

If there are no exclusions in `lib`, the workspace will be `["lib/"]`. If there are other exclusions in `lib` (but not `lib/unreadable.rb`) the workspace will be `["lib/unreadable.rb", "...", "..."]` with whatever was excluded, well, excluded.

(There's one caveat to  this: I think that docker-machine will prevent `lib/unreadable.rb` from even existing on the VM. I'm going to ignore this fact for now because I don't understand it well, not being an OSX user -- if you want more details, I defer to @wfleming)

> are we still expecting the engine to potentially raise an exception if/when it tries to read it? 

Yes, if the engine attempts to read a file it can't read, I'd expect/want an exception. I don't *think* we have any engines that silently rescue this case, but I'm not 100% sure.

wfleming:
Pat's description is correct, except for the following: in the docker-machine case the unreadable file will still exist, and will still be unreadable.

wfleming:
It should be noted that prior to this we explicitly excluded unreadable files in the CLI -- so if we *do* have any engines that silently rescue unreadable files, their observed behavior will not be changed by this work. I agree that's behavior we should change if it exists.

pbrisbin:
Ok. Then I'm not sure how the docker machine fact impacts things, but you
indicated previously (I think) that it may. Does it?

wfleming:
@pbrisbin the docker-machine case was part of my previous argument for why we needed to potentially handle unreadable files touched as part of the exclude/include path behavior. You were arguing (I think) that because the CLI is root in the container there shouldn't be any unreadable files, and I was providing a counter-example. Then I realized I was wrong about the needed behavior because of other details. You can safely forget everything I ever told you about docker-machine.

brynary:
OK, great. Both of those answers were what I was hoping/expecting. Do we have benchmark numbers on this implementation compared to current?

pbrisbin:
@wfleming great, I modified my comment to remove that confusion information, which, in retrospect, was just me making stuff up :)

Can the benchmark you have here be added to master (and maybe not made an assertion, but just output timings)? Then we could compare the numbers between the two branches.

lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -56,20 +62,32 @@ def names_and_raw_engine_configs
...
       end
 
-      def include_paths
-        IncludePathsBuilder.new(exclude_paths, Array(@requested_paths)).build
+      def base_workspace
+        @base_workspace ||= Workspace::PathTree.new(".").tap do |path_tree|
+          path_tree.include_paths(@requested_paths) if @requested_paths.present?
+          path_tree.exclude_paths([".git"])
+          normalize_paths(@config.exclude_paths).each do |exclude|
+            path_tree.exclude_paths(Workspace::Exclusion.new(exclude).expand)
```
pbrisbin: I think it's a little leaky to let/expect callers to reach in for `PathTree` and `Exclusion` directly. WDYT about having `Workspace.create` (or something) to encapsulate the `PathTree` instantiation, then have `PathTree#exclude_paths` accept patterns and wrap them as `Exclusion`s internally?

wfleming:
> OK, great. Both of those answers were what I was hoping/expecting. Do we have benchmark numbers on this implementation compared to current?

We do! For a known case I have on a local repo where workspace building used to take 20-30 minutes, it now takes less than 1 second. Not the most precise benchmark, I know, but I think the orders of magnitude are the important thing ;).

There's [also a spec](https://github.com/codeclimate/codeclimate/pull/308/files#diff-f51289e63709b4499881a0eb0290c346R13) here that tries to stress-test the algorithm & keep it honest & performant in the future.

pbrisbin:
@wfleming I went ahead with that refactor, bringing back `Workspace` as a proper class and it's spec -- which I think exposed a bug. Can you take a look at the `Workspace` specs marked `skip` and let me know what you think?

spec/cc/workspace/path_tree_benchmark_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      within_temp_dir do
+        make_files
+        exclude_paths = make_exclude_paths
+
+        start_time = Time.now
+        tree = PathTree.new(".")
+        tree.exclude_paths(exclude_paths)
+        elapsed = Time.now - start_time
```
pbrisbin: Should we use `Benchmark.realtime` for these measurements?

spec/cc/workspace/path_tree_benchmark_spec.rb
pbrisbin: I vote we put this and other benchmarks under `./benchmarks`, with a separate rake task. I'd also vote they output timing information (e.g. `Benchmark.bm`), rather than assert on timings being under some threshold.

I wouldn't characterize what happened as developers just not realizing, because nothing was watching their backs. Rather, I think we would've actively used benchmarks during the development of this logic, had we had them -- and we probably should've written them then. Therefore, I think benchmarks being available to run and view as needed is just as valuable as putting performance-based assertions in the normal spec suite.

Maybe some day we could incorporate it into the release process, to run the benchmarks between master and the last released tag.

spec/cc/workspace/path_tree_benchmark_spec.rb
wfleming: Yes, probably -- I also turned up some unexpected things about the size of the generated workspace while porting this to master (master did blow well past the 10 second mark at about 2 minutes, FWIW), so I think this spec needs some attention overall.

lib/cc/workspace/exclusion.rb
```diff
@@ -0,0 +1,29 @@
...
+      def glob?
+        pattern.include?("*")
+      end
+
+      private
+
+      attr_reader :pattern
+
+      def simplify(pattern)
+        pattern.to_s.sub(%r{(/\*\*)?(/\*)?$}, "")
```
pbrisbin: I believe, because of this, we no longer need that `is_a?(Glob) => .value` hack. Since this will simplify `x/**/*` and `x/**` both to `x`. After adding the workspace spec using `vendor/**`, it still passes without that logic.

lib/cc/workspace/path_tree.rb
```diff
@@ -0,0 +1,115 @@
...
+          end
+        end
+      end
+
+      def insert(path_pieces)
+        return if path_pieces.empty?
+
+        entry = Pathname.new(root_path).children.detect { |c| c.basename.to_s == path_pieces[0] }
+        if entry && entry.directory?
+          children[entry.basename.to_s] = PathTree.new(entry.to_s)
```
pbrisbin: `PathTree.new` here and `self.class.new` at line 81. Probably should be consistent, I think I prefer `self.class.new` since it's robust against renames.

lib/cc/workspace/path_tree.rb
```diff
@@ -0,0 +1,115 @@
...
+        end
+      end
+
+      protected
+
+      def populated?
+        children.present?
+      end
+
+      def delete(path_pieces)
```
pbrisbin: FWIW: If you made this method signature `delete(head, *tail)`, you'd get `path_pieces[0]` and `path_pieces.drop(1)` for free. This is a common way to operate on lists in Haskell.

lib/cc/workspace/path_tree.rb
```diff
@@ -0,0 +1,115 @@
...
+        if 1 == path_pieces.size
+          children.delete(path_pieces[0])
+        else
+          child = children[path_pieces[0]]
+          if child
+            child.delete(path_pieces.drop(1))
+            children.delete(path_pieces[0]) unless child.populated?
+          else
+            CLI.debug(
+              "Couldn't exclude because part of path doesn't exist or was already excluded.",
```
pbrisbin: Under what circumstances do you expect this? Is it possible to remove those circumstances so we don't have to have this branch here?

pbrisbin: Same question for `#insert`

brynary:
> We do! For a known case I have on a local repo where workspace building used to take 20-30 minutes, it now takes less than 1 second. 

Something something computer science something big-Oh notation blah blah :)

brynary:
OK, so with this behavioral change what's the customer impact we need to communicate? How would we instruct someone who has exclude paths to know if they need to do anything, and if so what do they need to do?

lib/cc/workspace/exclusion.rb
wfleming: Good point, thanks. I had added it back, but I just reverted it.

lib/cc/workspace/path_tree.rb
wfleming: This falls into the same category as our long slack convo. You could have an old exclude pattern that no longer applies because you deleted a directory, or maybe it only applies to local files and you're running in .com, or maybe you made a typo. I don't think any of those cases are worth being overly noisy about, but I think they are worth having a message about in debug/verbose mode.

wfleming: Also for the cases where you might give the style of paths that `PathValidator` used to cover -- if you run something like `analyze /root/path` or `analyze ../../not/in/the/project/repo`, this will provide feedback about why that didn't work (albeit only with debug visibility).

pbrisbin:
I don't think there are any behavioral changes here. Users will just see better performance. @wfleming would you agree?

There may be subtle changes to how `analyze FILE` invocations work (so called "explicit includes"), but we're still nailing that down, so I'm not sure what they'll be yet.

lib/cc/workspace/path_tree.rb
pbrisbin: I can be convinced that we should keep the logging in `#insert` (and maybe make it a WARN even). But for excludes, I think we should silently ignore.

Two reasons:

- When I create a git project, I start with a `.gitignore` full of things that don't exist (yet or never will), this is super common and I see analysis exclusions the same way
- Something not existing clearly won't be analyzed, i.e. it's been excluded -- technically speaking, we've done the Right Thing

wfleming:
The big behavioral change that probably needs broad explanation is that we don't exclude the same things `git` does using `.gitignore` anymore -- that should have 0 impact on .com, but it could dramatically change local CLI results in some cases, and customers may want to add many of their `.gitignore` paths to their `.codeclimate.yml` as well.

The non-handling of unreadable files is the other behavioral change, and should also only impact local CLI usage. It's more of an edge case, though, so probably doesn't need to be broadcast as widely.

wfleming:
@codeclimate/review I think this is in a good place now. Git history still needs cleaning up, but the code is feeling cleaner than it was this morning, and we've corrected the behavior around applying exclude paths when explicit include paths have been given (we no longer apply the excludes in this case). Other thoughts?

pbrisbin:
LG(reate)TM. I think there's still a product decision needed around customer comms, but we could consider that blocking release, not merge if we wanted.

wfleming:
@brynary You have a chance to think about how you'd like to handle shipping this & educating users of the behavior change?

My thought was, aside from putting it in the GitHub release description, posting a changelog. I could write that up today so we could release the update tomorrow along with the changelog.

robtarr:
I built this version today and ran my code through it - working great!

KevinBongart:
Just some feedback: my `.rubocop.yml` config file doesn't seem to be taken into account (https://github.com/codeclimate/codeclimate-rubocop/issues/39) when running CC from this branch. Not sure if this fits here, but it seems like this PR is also trying to fix the issue?

wfleming:
Thanks for the feedback, @KevinBongart: if you have more details about that, could you please add them to https://github.com/codeclimate/codeclimate-rubocop/issues/39? I'd be interested to know if that's only on this branch or also the case on master (I'd be surprised if the behavior changed).

Keep in mind we don't currently support the `inherit_from` directive in RuboCop config files: that's been the cause of unexpected RuboCop results in many cases.

brynary:
@wfleming I think we can probably roll this out with just a Changelog. Can discuss tomorrow.

PR #309, opened 2016-01-22
Fix docker-machine error handling in wrapper

wfleming:
Just happened to notice that this bit referenced boot2docker when it
should be using docker-machine

:eyes: @codeclimate/review


pbrisbin:
Nice. LGTM.

PR #310, opened 2016-01-24
Init: respect excludes when enabling engines

wfleming:
`ConfigGenerator` now has a slightly improved algorithm for finding
files matching engine's `enable_regexps`. This fixes a bug & noticeably
improves performance in some cases as well.

The bug fix is that because we used to check the *entire* tree for files
matching an engine's patterns, you could enable an engine whose only
matching files were in auto-exclude paths. E.g. a project under
`vendor/` might have some JavaScript files used for local scripting or
something while your own code had no JS at all.

The performance benefit comes for similar reasons as the recent
refactor of workspace calculation with exclude paths: we touch
less data & less of the filesystem.

The new algorithm is similar to the old one: we still shell out to
`find` for available files, but `find` now only searches files that are
*NOT* within the excluded directories. I tried a variety of
implementations here, including things more similar to (or using) the
new `PathTree` implementation, but this was still the fastest
implementation across a variety of cases.

I also made a behavior change here: this won't auto-enable any engine
based on files in "."-prependend directories: this is an opinion I'm happy to back
away from, but in general "."-prepended directories & files are usually stuff not
related to your actual code that you would want analyzed, so this seems
reasonable. It does introduce a slight gap in behavior in that we don't
exclude those directories from analysis, just from enabling engines.

This also removes unused functionality from `Analysis::Filesystem`.

### DIFF
--- a/lib/cc/analyzer/filesystem.rb
+++ b/lib/cc/analyzer/filesystem.rb
@@ -1,6 +1,8 @@
 module CC
   module Analyzer
     class Filesystem
+      attr_reader :root
+
       def initialize(root)
         @root = root
       end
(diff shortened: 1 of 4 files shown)


wfleming:
cc @codeclimate/review 

@noahd1 This might be worth trying on the local case you had where `init` took an unreasonably long time.

noahd1:
It was `worker` IIRC

wfleming:
@noahd1 Decent test case, though `master` wasn't *that* slow for me on `worker`.

* `master`: 10.5s
* this branch: 1.8s

noahd1:
10 seconds is prettttty slow! Nice work.

pbrisbin:
Code looks like an improvement on things. If tests pass and things are fast, LGTM.

PR #312, opened 2016-01-24
Init: don't create engine configs when inappropriate

wfleming:
Because of the behavior where we create engine configs even if a
`.codeclimate.yml` exists, there was a case where
`#create_default_engine_configs` would do the wrong thing: that method
still called `#available_engines` on `ConfigGenerator` to decide which
engines to create configs for. However, in the case of a
`.codeclimate.yml` created or edited by the user, there could be engines
that `available_engines` considers available, but that are not enabled
in the yaml. In that case, this method would create config files
for engines the user isn't using.

This changes the behavior to read the engines from the
`.codeclimate.yml`, since it must exist by the time this method is
called. As a bonus, in cases where the `.codeclimate.yml` already
existed & we didn't generate it, reading this from the yaml can be much
faster than using the `ConfigGenerator` logic.

### DIFF
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -10,7 +10,7 @@ module CC
       def run
         if !upgrade? && filesystem.exist?(CODECLIMATE_YAML)
           warn "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
-          create_default_engine_configs
+          create_default_engine_configs if engines_enabled?
         elsif upgrade? && engines_enabled?
           fatal "--upgrade should not be used on a .codeclimate.yml configured for the Platform.\nTry running 'validate-config' to check configuration."
         else
@@ -63,27 +63,32 @@ module CC
       end
 
       def available_engine_configs
-        all_paths = config_generator.eligible_engines.flat_map do |engine_name, _|
+        engine_names = existing_cc_config.engines.select do |_, config|
+          config.enabled?
+        end.keys
+
+        all_paths = engine_names.flat_map do |engine_name|
           engine_directory = File.expand_path("../../../../config/#{engine_name}", __FILE__)
           Dir.glob("#{engine_directory}/*", File::FNM_DOTMATCH)
         end
-
-        all_paths.reject { |path| [".", ".."].include?(File.basename(path)) }
+        all_paths.reject do |path|
+          %w[. ..].include?(File.basename(path))
+        end
       end
 
       def engines_enabled?
-        unless @engines_enabled.nil?
-          return @engines_enabled
-        end
-
-        if filesystem.exist?(CODECLIMATE_YAML)
-          config = CC::Analyzer::Config.new(File.read(CODECLIMATE_YAML))
-          @engines_enabled ||= config.engine_names.any?
-        end
+        cc_config = existing_cc_config
+        cc_config.present? && cc_config.engines.present?
       end
 
       def config_generator
         @config_generator ||= ConfigGenerator.for(filesystem, engine_registry, upgrade?)
+      end
+
+      def existing_cc_config
+        if filesystem.exist?(CODECLIMATE_YAML)
+          CC::Yaml.parse(filesystem.read_path(CODECLIMATE_YAML))
+        end
       end
     end
   end
--- a/spec/cc/cli/init_spec.rb
+++ b/spec/cc/cli/init_spec.rb
@@ -121,7 +121,8 @@ module CC::CLI
         it "still generates default config files" do
           filesystem.exist?(".codeclimate.yml").must_equal(false)
 
-          File.new(".codeclimate.yml", "w")
+          yaml = "---\nengines:\n  rubocop:\n    enabled: true"
+          File.write(".codeclimate.yml", yaml)
 
           filesystem.exist?(".codeclimate.yml").must_equal(true)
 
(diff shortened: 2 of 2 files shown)


wfleming:
cc @codeclimate/review

lib/cc/cli/init.rb
```diff
@@ -70,19 +70,21 @@ def available_configs
...
+        @engines_enabled ||= cc_config.present? && cc_config.engine_names.any?
       end
 
       def config_generator
         @config_generator ||= ConfigGenerator.for(filesystem, engine_registry, upgrade?)
       end
+
+      def existing_cc_config
+        if filesystem.exist?(CODECLIMATE_YAML)
+          CC::Analyzer::Config.new(filesystem.read_path(CODECLIMATE_YAML))
```
pbrisbin: `CC::Analyzer::Config` is a [dying class](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/analyzer/config.rb#L5). Can you try using `CC::Yaml` directly?

lib/cc/cli/init.rb
```diff
@@ -70,19 +70,21 @@ def available_configs
...
-          return @engines_enabled
-        end
+        return @engines_enabled unless @engines_enabled.nil?
 
-        if filesystem.exist?(CODECLIMATE_YAML)
-          config = CC::Analyzer::Config.new(File.read(CODECLIMATE_YAML))
-          @engines_enabled ||= config.engine_names.any?
-        end
+        cc_config = existing_cc_config
+        @engines_enabled ||= cc_config.present? && cc_config.engine_names.any?
```
GordonDiggs: I'm not sure how valuable this memoization is. Most of the work is probably done in `existing_cc_config` and nothing will be memoized if `present?` or `any?` return false.

lib/cc/cli/init.rb
```diff
@@ -10,7 +10,7 @@ class Init < Command
       def run
         if !upgrade? && filesystem.exist?(CODECLIMATE_YAML)
           warn "Config file .codeclimate.yml already present.\nTry running 'validate-config' to check configuration."
-          create_default_configs
+          create_default_engine_configs
```
pbrisbin: Is there a strong reason for these renames (inserting `_engine_`)? I'm finding it distracting when trying to make out the actual code change here.

Given your PR description I expected a smaller change, something along the lines of:

- extract something like `#engine_names` that uses cc.yml if present or `ConfigGenerator` if not
- Replace `config_generator.eligible_engines` with `#engine_names`

The other renames and refactors might be good, but I think it'd be easier to discern if the behavior-changing and non-behavior-changing stuff were in separate PRs.

wfleming: They're in separate commits to make review easier, if that helps. I can extract the rename commit as a separate PR if you'd like.

pbrisbin: The commits don't seem sliced that way. https://github.com/codeclimate/codeclimate/commit/668956f7ce32d996f27836d7d1aaf82649f876fb has a rename in it, for example. An, in that commit, I still can't see where you fallback to config-generator when `existing_cc_config` *doesn't* exist, but maybe I'm misunderstanding the intent (because of the diff confusion?).

wfleming: Bummer: I was doing a fair amount of rebase editing & must have fouled something. Sorry, I'll redo.

> I still can't see where you fallback to config-generator when existing_cc_config doesn't exist, but maybe I'm misunderstanding the intent (because of the diff confusion?).

I don't think that case exists, and I did not intend to implement that behavior. If the YAML already exists, it already exists, yay. If it does not exist when you run `init`, `#create_codeclimate_yaml` will write it, and again it will exist before `#create_default_engine_configs` is called.

pbrisbin: >  If it does not exist when you run init, #create_codeclimate_yaml will write it, and again it will exist before #create_default_engine_configs is called.

Sorry, yeah, I didn't get that originally. Makes sense.

lib/cc/cli/init.rb
wfleming: You're right the `||` bit is pointless: I'll remove that. We do an existence check up above, though, specifically so that we do memoize the `false` case.

pbrisbin:
LGTM. I'd support any of:

- Pull the rename out to its own PR, which I will immediately :+1:
- Fix the two commits so the right changes are in the right place, but keep as one PR
- Squash the commits to 1, if redoing them is too much of a hassle

lib/cc/cli/init.rb
wfleming: Will do.

pbrisbin: Feel free to argue for its existence though, if you find in the course of doing so that it makes sense to keep it. The decision to deprecate was made a long time ago and possibly in haste.

wfleming:
Thanks @GordonDiggs & @pbrisbin: ready for another look.

lib/cc/cli/init.rb
```diff
@@ -63,28 +63,33 @@ def create_default_engine_configs
...
+        end.keys
+
+        all_paths = engine_names.flat_map do |engine_name|
           engine_directory = File.expand_path("../../../../config/#{engine_name}", __FILE__)
           Dir.glob("#{engine_directory}/*", File::FNM_DOTMATCH)
         end
-
-        all_paths.reject { |path| [".", ".."].include?(File.basename(path)) }
+        all_paths.reject do |path|
+          [".", ".."].include?(File.basename(path))
```
GordonDiggs: :lipstick: use `%w` here

GordonDiggs:
One style thing but otherwise LGTM

PR #314, opened 2016-01-25
Refactor some names in Init

wfleming:
We use "config" all over the place here, but don't clarify if we're
talking about our own config or the config files for engines or what.
Just trying to neaten up a bit.

cc @pbrisbin


pbrisbin:
LGTM

PR #316, opened 2016-01-30
Run specs in docker, fix race condition

pbrisbin:
I went with separate test and citest make targets (vs "test: image test_only")
because running on CI is also different in that it should pass the CI-related
enviroment variables too.

d845ec6 has details about the race condition but not the fix, d0ee4b3 actually
implements what I think is a real fix.

/cc @codeclimate/review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,27 @@
-.PHONY: build install uninstall
+.PHONY: build install uninstall image test citest
 
 PREFIX ?= /usr/local
 
 image:
 	docker build -t codeclimate/codeclimate .
+
+test: image
+	docker run --rm \
+	  --entrypoint bundle \
+	  --volume /var/run/docker.sock:/var/run/docker.sock \
+	  codeclimate/codeclimate exec rake
+
+citest:
+	docker run \
+	  --env CIRCLECI=$(CIRCLECI) \
+	  --env CIRCLE_BUILD_NUM=$(CIRCLE_BUILD_NUM) \
+	  --env CIRCLE_BRANCH=$(CIRCLE_BRANCH) \
+	  --env CIRCLE_SHA1=$(CIRCLE_SHA1) \
+	  --env CODECLIMATE_REPO_TOKEN=$(CODECLIMATE_REPO_TOKEN) \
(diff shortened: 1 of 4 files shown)


pbrisbin:
FYI: this is a re-opening of #265

Makefile
```diff
@@ -1,10 +1,27 @@
...
 
 PREFIX ?= /usr/local
 
 image:
 	docker build -t codeclimate/codeclimate .
 
+test: image
+	docker run --rm \
+	  --entrypoint bundle \
+	  --volume /var/run/docker.sock:/var/run/docker.sock \
```
wfleming: Looks like we're missing the rest of the command here?

lib/cc/analyzer/container.rb
```diff
@@ -172,6 +172,7 @@ def kill_reader_threads
       def reap_running_container(message)
         Analyzer.logger.warn("killing container name=#{@name} message=#{message.inspect}")
         POSIX::Spawn::Child.new("docker", "kill", @name)
+        POSIX::Spawn::Child.new("docker", "wait", @name)
```
wfleming: In the past we've seen zombie engines that continued to run even after being killed. If that were to happen, this would result in builder effectively zombieing as well as it would wait for something that isn't actually going to die. That might make a zombie engine more visibile, which would be good, but it might also just make the problem worse. Thoughts on that?

Makefile
pbrisbin: The editor version of "I did a `mv` when I mean `cp`". Fixed.

lib/cc/analyzer/container.rb
pbrisbin: You're right. I was originally going to argue against adding the complexity of a timeout here for a number of reasons, then I realized I can just pass the `:timeout` option to `Child.new` which is super easy. I'll do that.

wfleming: That makes sense to me!

pbrisbin:
@wfleming good for a squash-merge?

wfleming:
:eggplant: :ship:

PR #317, opened 2016-02-01
Simplify exclude_paths generated by init

wfleming:
Previously we appended `**/*` to all auto-excluded directories: that's
no longer required by the new exclude behavior, so no reason to keep it.

cc @codeclimate/review

### DIFF
--- a/lib/cc/cli/config.rb
+++ b/lib/cc/cli/config.rb
@@ -21,13 +21,7 @@ module CC
 
       def add_exclude_paths(paths)
         config["exclude_paths"] ||= []
-        config["exclude_paths"] += paths.map do |path|
-          if path.ends_with?("/")
-            "#{path}**/*"
-          else
-            path
-          end
-        end
+        config["exclude_paths"] |= paths
       end
 
       private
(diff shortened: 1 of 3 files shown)


jpignata:
Did you mean to type `**/*` is what we'd append or am I missing something?

wfleming:
> Did you mean to type **/* is what we'd append or am I missing something?

Sigh. That *is* what I typed, but markdown decided that meant I wanted `\*\<em\>/\</em\>`

lib/cc/cli/config.rb
```diff
@@ -21,13 +21,7 @@ def add_engine(engine_name, engine_config)
...
       def add_exclude_paths(paths)
         config["exclude_paths"] ||= []
-        config["exclude_paths"] += paths.map do |path|
-          if path.ends_with?("/")
-            "#{path}**/*"
-          else
-            path
-          end
-        end
+        config["exclude_paths"] += paths
```
GordonDiggs: Could be a good use case for `|=` here

wfleming: Good point. Updated.

GordonDiggs:
LGTM

PR #319, opened 2016-02-01
Update Gems

wfleming:
Primarily after cc-yaml 0.8.0 here, but there's some patch releases
available we should take as well.

cc-yaml: https://github.com/codeclimate/codeclimate-yaml/commit/e141f7eaf14d4c0493149aecadce5345182504ef
rainbow: https://github.com/sickill/rainbow/compare/v2.0.0...v2.1.0

cc @codeclimate/review


pbrisbin:
LGTM

PR #323, opened 2016-02-03
Replace minitest with RSpec

pbrisbin:
Reasons:

- It's a burden to know and use two testing frameworks across our projects
- By far, rspec is what we use most frequently
- Possibly related, most of us personally prefer rspec
- We used minitest/spec, so it's *just close enough* to rspec to repeatedly
  confuse us when it's not
- We were bringing in a number of extra libraries anyway, just to give minitest
  the functionality that rspec has built-in

Changes:

This was mostly a mechanical change, but also:

- Add ProcHelpers#capture_io

  minitest, really? Really? You give me almost nothing out of the box, claim
  that as your main feature, but you have #capture_io, really?

- Add a slow tag for the container-stopping specs and don't run them by default

  I was running the suite a lot to get through this change and it's incredibly
  fast, if not for these specs. We intentionally don't touch this logic often,
  so I think it makes sense to omit them in day-to-day development but keep
  them on CI for regression protection.

There are still a few places where the specs are overly verbose or awkward
because we didn't have an rspec feature that we do now. I'm deferring fixing
those for now, this conversion was the main hurdle.

CC will also likely flag some pre-existing issues because I've touched the
line(s) -- I'll fix these when I see them.

/cc @codeclimate/review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,11 +3,8 @@ source "https://rubygems.org"
 gemspec
 
 group :test do
-  gem "mocha"
-  gem "minitest"
-  gem "minitest-around"
-  gem "minitest-reporters"
   gem "rack-test"
   gem "rake"
+  gem "rspec"
   gem "codeclimate-test-reporter", require: nil
 end
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source "https://rubygems.org"
 gemspec
 
 group :test do
-  gem "rack-test"
   gem "rake"
   gem "rspec"
   gem "codeclimate-test-reporter", require: nil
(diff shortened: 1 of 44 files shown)


Gemfile
```diff
@@ -3,11 +3,8 @@ source "https://rubygems.org"
 gemspec
 
 group :test do
-  gem "mocha"
-  gem "minitest"
-  gem "minitest-around"
-  gem "minitest-reporters"
   gem "rack-test"
```
pbrisbin: I have no idea why this would be here. I'm running the suite locally now without it and will commit its removal if it's green, or (hopefully) explain why it's here.

wfleming:
:+1:

PR #324, opened 2016-02-03
[wrapper] do not use docker-machine if there is no name

mikz:
some people have it installed, but do not use it

i could not use codeclimate wrapper with installed docker-machine
but also working local install

### DIFF
--- a/codeclimate-wrapper
+++ b/codeclimate-wrapper
@@ -28,7 +28,7 @@ invalid_docker_host() {
   invalid_setup "invalid DOCKER_HOST=$host, must be unset or unix:///var/run/docker.sock"
 }
 
-if command -v docker-machine > /dev/null 2>&1; then
+if [ -n "$DOCKER_MACHINE_NAME" ] && command -v docker-machine > /dev/null 2>&1; then
   docker-machine ssh $DOCKER_MACHINE_NAME -- \
     test -S /var/run/docker.sock > /dev/null 2>&1 || socket_missing
 


codeclimate-wrapper
```diff
@@ -28,7 +28,7 @@ invalid_docker_host() {
   invalid_setup "invalid DOCKER_HOST=$host, must be unset or unix:///var/run/docker.sock"
 }
 
-if command -v docker-machine > /dev/null 2>&1; then
+if [ "$DOCKER_MACHINE_NAME" ] && command -v docker-machine > /dev/null 2>&1; then
```
pbrisbin: I think you meant `[ -n "..." ]` (the test you have is a bashism only supported by `[[`). Also, do we think it's reasonable to go by this check alone and remove the `command` check?

pbrisbin:
Hi @mikz, thanks for this. Once the `-n` fix is in place, I'd be in favor of merging it, but I'd like for someone on our team who uses OS X and docker-machine to weigh sign-off first.

codeclimate-wrapper
mikz: Good to know. I found it there: http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html
which is the bash man. Will change it to `-n`.

wfleming:
I've tested this locally, and this is a welcome fix, so I'm in favor of merging if we can add the `-n`. @mikz would you mind rebasing against our current `master` as well when you make that change? We try to avoid merge commits when we can. Thanks!

mikz:
@wfleming @pbrisbin fixed, rebased!

wfleming:
Thanks, @mikz !

PR #325, opened 2016-02-04
Add timeout on docker kill invocation

pbrisbin:
I went with two minutes here and reduced the timeout on docker wait to match.
The five minutes originally chosen was entirely arbitrary. Allowing up to ten
minutes of additional time in a timeout scenario didn't seem like a good idea
and my gut is that if it doesn't stop within two, it probably won't.

/cc @codeclimate/review


wfleming:
LGTM

PR #330, opened 2016-02-08
ConfigGenerator: raise error if workspace is empty

wfleming:
We were just returning an empty set of paths to enable engines on in
init when everything was excluded and/or the dir was empty, but this
strikes me as incorrect. If there's really nothing for us to look at, it
seems more reasonable to call that an error condition & exit
immediately.

:eyes: @codeclimate/review

### DIFF
--- a/lib/cc/cli/config_generator.rb
+++ b/lib/cc/cli/config_generator.rb
@@ -6,6 +6,8 @@ module CC
       CODECLIMATE_YAML = Command::CODECLIMATE_YAML
       AUTO_EXCLUDE_PATHS = %w(config/ db/ dist/ features/ node_modules/ script/ spec/ test/ tests/ vendor/).freeze
 
+      EmptyWorkspaceError = Class.new(StandardError)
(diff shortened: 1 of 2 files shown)


pbrisbin:
I'm not sure I agree with this UX. At least, I'm not ready to change this UX if we don't have known cases we're dealing with where we've decided this exception is the solution. I think arguments could be made either way about what *should* happen, and so I hesitate to add a new errored-build scenario where we don't have one currently.

To me this is similar to the `NoEnginesEnabled` error we removed recently, and I agreed with the justification there. As long as it's clear in the build output (I might consider adding a warning here) that we've done a "no-op" analysis, I think it's fine and better than an errored snapshot, which can be very problematic for first-run analyses (where this scenario is most likely to be seen IMO) and merge base analyses.

wfleming:
@pbrisbin Good points. I don't care enough to do the necessary refactoring right now to make this a warning, so I'll just close for now.

PR #331, opened 2016-02-08
ConfigGenerator: Catch & report `find` failure

wfleming:
Our usage of `find` to get files to enable engines against is too
naive: there are some unusual local configurations that lead to an
invalid find invocation. The old code was effectively ignoring this,
leading to more confusing failures further down the line.

This replaces the naive execution with usage of `POSIX::Spawn` that
should 1) be more tolerant of weird file names anyway and 2) if find
does fail, will properly catch & report so at least the user gets an
error message pointing them in a useful direction.

@codeclimate/review this is in direct response to some builder errors we saw on
.com. It's only affected 2 repos I've seen, but this should result in a more
informative view on the build page, and should result in more informative
errors for us on the backend to improve this further.

### DIFF
--- a/lib/cc/cli/config_generator.rb
+++ b/lib/cc/cli/config_generator.rb
@@ -1,3 +1,4 @@
+require "posix/spawn"
 require "shellwords"
 
 module CC
@@ -5,6 +6,8 @@ module CC
     class ConfigGenerator
       CODECLIMATE_YAML = Command::CODECLIMATE_YAML
       AUTO_EXCLUDE_PATHS = %w(config/ db/ dist/ features/ node_modules/ script/ spec/ test/ tests/ vendor/).freeze
+
+      ConfigGeneratorError = Class.new(StandardError)
 
       def self.for(filesystem, engine_registry, upgrade_requested)
         if upgrade_requested && upgrade_needed?(filesystem)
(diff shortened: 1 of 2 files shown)


lib/cc/cli/config_generator.rb
```diff
@@ -80,13 +83,24 @@ def workspace_files
...
+        find_cmd = %w[find] + non_excluded_paths + %w[-type f -print0]
+        pid, _, out, err = POSIX::Spawn.popen4(*find_cmd)
+        _, status = Process.waitpid2(pid)
+
+        if status.success?
+          out.read.strip.split("\0").map do |path|
+            path.sub(%r{^\.\/}, "")
+          end
+        else
+          raise FindFailError, "We ran '#{find_cmd.shelljoin}' to find analyzeable files, but the command failed with this error: '#{err.read}'"
```
pbrisbin: typo: `analyzable`

I think we should decide if this exception is primarily for us or end-users. If it's for us, I'd argue for a more generic construct, something like `CC::Spawn` that takes a command (as Array) and returns `out` or raises a more implementation-specific error:  `raise ProcessError, "command <x, y, z> exited with status <a> and stderr <b>"`. If it's more for the end-users, then I think we shouldn't leak the abstraction so much: `raise ConfigGeneratorError, "Failed to find analyzable files.\n(additional details for us like command run, exit status, etc)"`

WDYT?

wfleming: I favor the second, at least for the time being. I think wrapping up our spawn logic somewhere might be useful, but I think getting better user-facing errors out is important, so I'll focus on that for now.

wfleming:
@pbrisbin updated

lib/cc/cli/config_generator.rb
```diff
@@ -80,13 +83,24 @@ def workspace_files
...
-            end
+            find_workspace_files
           end
         end
       end
+
+      def find_workspace_files
+        find_cmd = %w[find] + non_excluded_paths + %w[-type f -print0]
+        pid, _, out, err = POSIX::Spawn.popen4(*find_cmd)
+        _, status = Process.waitpid2(pid)
```
pbrisbin: I think `POSIX::Spawn::Child` (a pattern we use more frequently) wraps this up

```rb
child = POSIX::Spawn::Child.new(*find_cmd)

if child.success?
  # use child.out
else
  # use child.err and child.status
end
```

pbrisbin:
One thought about using the nicer `Child` interface. LGTM besides.

wfleming:
@pbrisbin Good point, updated if you want to check that.

PR #332, opened 2016-02-08
Remove characters added by docker logging daemon

GordonDiggs:
@codeclimate/review

### DIFF
--- a/lib/cc/analyzer/engine_output.rb
+++ b/lib/cc/analyzer/engine_output.rb
@@ -1,7 +1,7 @@
 module CC
   module Analyzer
     class EngineOutput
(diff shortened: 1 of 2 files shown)


pbrisbin:
How has this not caused massive problems?

GordonDiggs:
@pbrisbin >100 builds have errored in the past few days because of it. I think we might be seeing it more because of the docker upgrade and new nodes

pbrisbin:
Gotcha. I figured if it was a 1.9 thing we'd be seeing 100% errored builds since the upgrade. Interesting that it's only sometimes an issue.

Anyway, LGTM.

PR #335, opened 2016-02-08
Revert "Remove characters added by docker logging daemon"

GordonDiggs:
Reverts codeclimate/codeclimate#332

This seems to have caused more errored builds. Reverting while I investigate to unblock folks.

cc @codeclimate/review @wfleming


wfleming:
Worth a shot :+1:

PR #337, opened 2016-02-09
Allow gem callers to disable container cleanup (docker run --rm)

dblandin:
This PR moves the `--rm` docker run option up a level to allow callers of
`CC::Analyzer::Container` to specify whether docker should automatically
cleanup engine containers after they exit. The CLI will still pass the `--rm`
option during engine runs, so this will be a transparent change for CLI users.

This update benefits Code Climate debugging efforts, giving us the option to
inspect a docker container after it triggered an errored build.

### DIFF
--- a/lib/cc/analyzer/container.rb
+++ b/lib/cc/analyzer/container.rb
@@ -97,7 +97,6 @@ module CC
       def docker_run_command(options)
         [
           "docker", "run",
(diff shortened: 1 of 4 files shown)


pbrisbin:
LGTM

ABaldwinHunter:
:+!

PR #345, opened 2016-02-12
Add old RuboCop engine to manifest

jpignata:
Since there are many breaking configuration changes between 0.35.1 and
0.37.2, support the previous version of the engine also so that users
who hit blocking issues with the current engine can revert back if
they do not wish to update their RuboCop configuration file.

This change means we will have to tag b165 as
codeclimate/codeclimate-rubocop-old and push it to Docker Hub.

@codeclimate/review

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -207,6 +207,13 @@ rubocop:
     - \.rb$
   default_ratings_paths:
     - "**.rb"
+rubocop-v35:
+  image: codeclimate/codeclimate-rubocop:v35
+  description: A Ruby static code analyzer, based on the community Ruby style guide. Version 0.35.1 of RuboCop.
+  community: false
+  enable_regexps:
+  default_ratings_paths:
+    - "**.rb"
 rubymotion:
   image: codeclimate/codeclimate-rubymotion
   description: Rubymotion-specific rubocop checks.


GordonDiggs:
LGTM

config/engines.yml
```diff
@@ -207,6 +207,16 @@ rubocop:
...
   default_ratings_paths:
     - "**.rb"
+rubocop-old:
+  image: codeclimate/codeclimate-rubocop-old
+  description: A Ruby static code analyzer, based on the community Ruby style guide. Version 0.35.1 of RuboCop.
+  community: false
+  upgrade_languages:
+   - Ruby
+  enable_regexps:
+    - \.rb$
```
pbrisbin: Will this auto-enable both new and old if `.rb` files are found?

config/engines.yml
```diff
@@ -207,6 +207,16 @@ rubocop:
     - \.rb$
   default_ratings_paths:
     - "**.rb"
+rubocop-old:
+  image: codeclimate/codeclimate-rubocop-old
+  description: A Ruby static code analyzer, based on the community Ruby style guide. Version 0.35.1 of RuboCop.
+  community: false
+  upgrade_languages:
+   - Ruby
```
pbrisbin: With this auto-enable both new and old if a classic config specified `Ruby`?

config/engines.yml
jpignata: Tested and the answer is yep. Yanking.

wfleming:
Can we tag it as something more descriptive than `old`? like `rubocop-v36`? Or even `rubocop:v36`?

jpignata:
@wfleming I'd be fine with `-v35`. I'm not sure we can support a `:v35` tag of the engine, but if you have a clear idea of how to go this route I'd be down for it.

wfleming:
> I'm not sure we can support a :v35 tag of the engine, but if you have a clear idea of how to go this route I'd be down for it.

I may not be thinking of something, but I think if you just put the image as `codeclimate/codeclimate-rubocop:v35` in the yml here & in builder, it'll work. That value is used directly for running & pulling engines in CLI, and running in builder. So as long as we push that tag out to GCR & Docker Hub, I think it'll be fine.

config/engines.yml
```diff
@@ -207,6 +207,13 @@ rubocop:
     - \.rb$
   default_ratings_paths:
     - "**.rb"
+rubocop-old:
+  image: codeclimate/codeclimate-rubocop-old
+  description: A Ruby static code analyzer, based on the community Ruby style guide. Version 0.35.1 of RuboCop.
```
ABaldwinHunter: what do you think about calling this the rubocop version instead of old?

ABaldwinHunter: ah just saw top level comment addressing

jpignata:
@wfleming So you're proposing:

```
rubocop-v35:
  image: codeclimate/codeclimate-rubocop:v35
  description: A Ruby static code analyzer, based on the community Ruby style guide. Version 0.35.1 of RuboCop.
  community: false
  enable_regexps:
  default_ratings_paths:
    - "**.rb"
```

ABaldwinHunter:
I think having v-35 in engine name instead of as a tag for now makes sense and would be fine.

That way there's no conflict with the way we use tags to identify versions of an engine.

This `old rubocop engine` we can think of as a separate engine under hood, instead of different version of the same imo.

ABaldwinHunter:
@jpignata that sounds good to me!

wfleming:
@jpignata Yep, that's pretty much exactly what I'm proposing.

Should probably remove the `enable_regexps`, though: otherwise we'll auto-enable both the old & new engines, which won't go well.

jpignata:
@wfleming That works for me.

`enable_regexps` is empty. It shouldn't auto-enable.

wfleming:
@jpignata Sorry, yeah, I edited my comment: I just read too hastily!

jpignata:
@wfleming Updated.

wfleming:
:sheep:

PR #352, opened 2016-02-18
Remove fixme issues from default .rubocop.yml

ABaldwinHunter:
This fixes a weird experience where users with an autogenerated
.rubocop.yml had fixme issues reported.

@codeclimate/review 

Note: an alternative I considered was changing the fixme examples to lower case (which our Fixme engine ignores). Didn't go that route because I thought the dashes were more intention revealing and better convey the original example words intended.

### DIFF
--- a/config/rubocop/.rubocop.yml
+++ b/config/rubocop/.rubocop.yml
@@ -510,9 +510,7 @@ Style/CommandLiteral:
   Enabled: false
 
 Style/CommentAnnotation:
-  Description: >-
-                 Checks formatting of special comments
-                 (TODO, FIXME, OPTIMIZE, HACK, REVIEW).
+  Description: 'Checks formatting of annotation comments.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#annotate-keywords'
   Enabled: false
 


config/rubocop/.rubocop.yml
```diff
@@ -512,7 +512,7 @@ Style/CommandLiteral:
 Style/CommentAnnotation:
   Description: >-
                  Checks formatting of special comments
-                 (TODO, FIXME, OPTIMIZE, HACK, REVIEW).
+                 (T-ODO, F-IXME, O-PTIMIZE, H-ACK, R-EVIEW).
```
GordonDiggs: I think this format is more confusing. Can we just delete the line? There's a url to explain this more if people are confused. Maybe we could also change the description to `Checks for presence of maintenance comments`? WDYT?

jpignata: What do you think about replacing the description with:

> Checks formatting of annotation comments.

This is the language the Ruby style guide users and sufficiently explains what the check is. I think a hyphen in the word would be confusing to a reader.

See: https://github.com/bbatsov/ruby-style-guide#comment-annotations

ABaldwinHunter: Sure either of those are fine by me.

I can go with `Checks formatting of annotation comments.`.

Personally, I prefer examples but I'm a newbie. ;)

ABaldwinHunter:
@GordonDiggs @jpignata ready for another look!

config/rubocop/.rubocop.yml
```diff
@@ -510,9 +510,7 @@ Style/CommandLiteral:
   Enabled: false
 
 Style/CommentAnnotation:
-  Description: >-
-                 Checks formatting of special comments
-                 (TODO, FIXME, OPTIMIZE, HACK, REVIEW).
+  Description: 'Checks for presence of maintenance comments.'
```
jpignata: Nit: It's not checking for presence. It's checking for formatting of those comments when present.

ABaldwinHunter:
@jpignata thanks! that was a good distinction to catch - I misunderstood the function before.

Updated.

jpignata:
:+1:

PR #354, opened 2016-02-18
Fix bug excluding a directory after partially excluding it

wfleming:
There was a bug here where if a directory was populated (probably due to
being partially excluded), and then later completely excluded (such as for
an engine-specific exclude), it would not actually be fully removed.

@codeclimate/review I chose to do a refactor first to (I hope) clarify the
behavior of this code & make it easier to read & understand. It's pretty much
copy-paste & some renaming, so I don't think it will make review terribly
difficult. The actual bug here is quite small: reviewing the two commits
seperately might be easiest.  I'm also happy to split into different PRs.

### DIFF
--- a/spec/cc/workspace/path_tree_spec.rb
+++ b/spec/cc/workspace/path_tree_spec.rb
@@ -44,6 +44,17 @@ class CC::Workspace
       end
     end
 
+    it "excludes directory after excluding only part of it" do
+      within_temp_dir do
+        make_fixture_tree
+
+        tree = PathTree.for_path(".")
+        tree.exclude_paths(["code/a/bar.rb"])
+        tree.exclude_paths(["code/"])
+        expect(tree.all_paths.sort).to eq [".git/", "foo.txt", "lib/"]
+      end
+    end
+
     it "handles including nonexistent files" do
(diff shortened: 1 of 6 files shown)


pbrisbin:
LGTM

spec/cc/workspace/path_tree_spec.rb
```diff
@@ -58,7 +69,7 @@ class CC::Workspace
       within_temp_dir do
         make_fixture_tree
 
-        tree = PathTree.new(".")
+        tree = PathTree.for_path(".")
         tree.exclude_paths(["code/does-not-exist"])
         expect(tree.all_paths.sort).to eq [".git/", "code/a/", "code/foo.rb", "foo.txt", "lib/"]
```
ABaldwinHunter: This might be outside scope of this PR, but maybe we could consider in the future blowing up here instead? Not sure which is better.

wfleming: If you're suggesting that attempting to exclude a non-existent path should blow up, we can't really do that. The use case is that you may have files locally you want excluded, but those files may be in your `.gitignore`, so they won't exist on .com.

PR #356, opened 2016-02-24
Add Reek engine

wfleming:
This engine is ready to be added as available.

@codeclimate/review

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -191,6 +191,16 @@ radon:
     - \.py$
   default_ratings_paths:
     - "**.py"
+reek:
+  image: codeclimate/codeclimate-reek
+  description: "Reek examines Ruby classes, modules, and methods and reports any code smells it finds."
+  community: true
+  upgrade_languages:
+    - Ruby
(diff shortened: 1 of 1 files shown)


config/engines.yml
```diff
@@ -191,6 +191,16 @@ radon:
     - \.py$
   default_ratings_paths:
     - "**.py"
+reek:
+  image: codeclimate/codeclimate-reek
+  description: "Reek examines Ruby classes, modules, and methods and reports any code smells it finds."
+  community: true
+  upgrade_languages:
+    - Ruby
```
GordonDiggs: Is it worth it to set the upgrade and enable patterns if it's not going to be enabled by default?

wfleming: We've tended to do it on other engines just because it makes it an easier change if we decide we do want to enable them by default, I think. The same question would apply to the `enable_regexps` field, and we add that for every engine.

GordonDiggs: Cool - makes sense

GordonDiggs:
LGTM

PR #361, opened 2016-03-01
rm faraday from gemspec

wfleming:
It is not used: I find now mention of it in code, and specs pass just fine.

cc @codeclimate/review


jpignata:
LGTM

PR #368, opened 2016-03-07
Correct unit in engine timeout message

ABaldwinHunter:
@codeclimate/review 

alternatively:

```
message = "engine #{engine_name} ran for #{data.duration / 1000} seconds
```


GordonDiggs:
I think presenting it as seconds would be better (your alternative)

ABaldwinHunter:
@GordonDiggs Updated. Ready for another look!

GordonDiggs:
:+1:

Issue #299, opened 2016-01-14
Engines rely on hard-coded /code paths.


DavidEBest:
The CodeClimate engines are hard-coded to require the code to live in the `/code` directory. Could that be parameterized?

The reason I ask is that we are attempting to build a Concourse (Concourse.CI) pipeline for continuous compliance scanning. We like the CC approach to Dockerized engines, but Concourse does not allow you to create folders at the root of the file system. Making this change would allow us to easily use the same configuration both inside the CC Platform, and in our own environment.

Thoughts?

GordonDiggs:
Hey @DavidEBest the `/code` location is a requirement in [our spec](https://github.com/codeclimate/spec/blob/master/SPEC.md#input) and is unlikely to change, as it provides a lot of convenience as well as security. It's important to note that `/code` is the directory _inside_ the container - it can be any directory outside the container.

dblandin:
Hey @DavidEBest,

We recently made some enhancements to mounted code and tmp paths in #369 which may be of interest to you.

You can use the `CODECLIMATE_CODE` environment variable (renamed from `CODE_PATH`) to specify a host directory containing application code to be analyzed. This will default to the host machine's current working directory.

You can also use the `CODECLIMATE_TMP` environment variable to specify a host directory that the CLI to use for temporary files during analysis. This will default to `/tmp/cc` on the host machine.

You can use these new environment variables with the recently released version [v0.22.1](https://github.com/codeclimate/codeclimate/releases/tag/v0.22.1) of the Code Climate CLI. Let me know if you have any questions!

DavidEBest:
Thanks for the update! I'll give it a shot.

PR #369, opened 2016-03-08
Refactor handling of mounted code and tmp paths

pbrisbin:
This was some work I did on a plane ride a few weeks ago and forgot to PR until
now. The main driver for this change is that is allows users to stop us from
creating a root-owned `/tmp/cc` on their systems. As part of doing so, I was
able to (I think) clean up the code a bit and solve an issue where our test
suite would also create (or be impaired by) the root-owned temp directory.

I've tested local analysis, changing the temp location, and even changing the
code location to analyze some other, non-current directory. I still need to
test that builder can use this interface as-is.

/cc @codeclimate/review

---

Adds a "MountedPath" abstraction that provides a single identity for the case
where we have an in-container path and need to know the on-host path it
represents for the purpose of re-mounting into other containers.

Removes the hap-hazard FILESYSTEM_DIR and inconsistently-named CODE_PATH
environment variables. FILESYSTEM_DIR goes away as MountedPath can maintain
that functionality itself. CODE_PATH is replaced with CODECLIMATE_CODE to be
consistent with the new CODECLIMATE_TMP and existing CODECLIMATE_DEBUG.

We continue to read CODE_PATH if CODECLIMATE_CODE is unspecified. This is to
support any existing users with their own wrapper scripts in the wild
(interestingly, Builder fits in this category).

MountedPath internally makes use of a CODECLIMATE_DOCKER variable to set the
local and in-container paths as equal if not being run in a container. This
makes local development and testing more convenient.

Having CODECLIMATE_TMP means we don't have to force a root-owned /tmp/cc on our
users' machines. Users can set this variable to whatever they want. We make use
of this in the specs.

Builder should work as-is. It assumes /tmp/cc, but that was maintained as the
default when CODECLIMATE_TEMP is not set. We would now be free to use that
variable to make a build-specific temporary directory though.

One unknown: the test subcommand was messing with the CODE_PATH and
FILESYSTEM_DIR variables as part of running its analysis. This code is untested
(ha) and I don't understand it, so I'm not sure what needs to change to support
the MountedPath approach.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,5 @@ RUN chmod +x /bin/docker
 
 COPY . /usr/src/app
 
-ENV FILESYSTEM_DIR /code
-
+ENV CODECLIMATE_DOCKER 1
 ENTRYPOINT ["/usr/src/app/bin/codeclimate"]
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -43,7 +43,7 @@ module CC
         end
 
         write_config_file
-        CLI.debug "engine config: #{File.read(config_file).inspect}"
+        CLI.debug "engine config: #{config_file.read.inspect}"
         container.run(container_options)
       ensure
         delete_config_file
@@ -60,7 +60,7 @@ module CC
           "--net", "none",
           "--rm",
           "--volume", "#{@code_path}:/code:ro",
-          "--volume", "#{config_file}:/config.json:ro",
+          "--volume", "#{config_file.host_path}:/config.json:ro",
           "--user", "9000:9000"
         ]
       end
@@ -70,16 +70,15 @@ module CC
       end
 
       def write_config_file
-        FileUtils.mkdir_p(File.dirname(config_file))
-        File.write(config_file, @config.to_json)
+        config_file.write(@config.to_json)
       end
 
       def delete_config_file
-        File.delete(config_file) if File.file?(config_file)
+        config_file.delete if config_file.file?
       end
 
       def config_file
-        @config_file ||= File.join("/tmp/cc", SecureRandom.uuid)
+        @config_file ||= MountedPath.tmp.join(SecureRandom.uuid)
       end
 
       def output_filter
--- a/lib/cc/analyzer/mounted_path.rb
+++ b/lib/cc/analyzer/mounted_path.rb
@@ -1,0 +1,80 @@
+module CC
+  module Analyzer
+    class MountedPath
+      DEFAULT_CODECLIMATE_TMP = "/tmp/cc".freeze
+
+      def self.code
+        host_prefix = ENV["CODECLIMATE_CODE"]
+        host_prefix ||= ENV["CODE_PATH"] # deprecated
+
+        if ENV["CODECLIMATE_DOCKER"]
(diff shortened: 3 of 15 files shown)


lib/cc/analyzer/engine.rb
```diff
@@ -60,7 +60,7 @@ def container_options
           "--net", "none",
           "--rm",
           "--volume", "#{@code_path}:/code:ro",
-          "--volume", "#{config_file}:/config.json:ro",
+          "--volume", "#{config_file.local_path}:/config.json:ro",
```
dblandin: What do you think of renaming `local_path` to `host_path`? When scanning other lines, reading "local" prompted the question "local to what?".

lib/cc/analyzer/mounted_path.rb
```diff
@@ -0,0 +1,80 @@
...
+  module Analyzer
+    class MountedPath
+      DEFAULT_CODECLIMATE_TMP = "/tmp/cc".freeze
+
+      def self.code
+        local_prefix = ENV["CODECLIMATE_CODE"]
+        local_prefix ||= ENV["CODE_PATH"] # deprecated
+
+        if ENV["CODECLIMATE_DOCKER"]
+          new(local_prefix, "/code")
```
dblandin: Same question here for `local_prefix` -> `host_prefix`.

lib/cc/analyzer/engine.rb
pbrisbin: I like it. Will update throughout.

Dockerfile
```diff
@@ -15,6 +15,5 @@ RUN chmod +x /bin/docker
 
 COPY . /usr/src/app
 
-ENV FILESYSTEM_DIR /code
-
+ENV CODECLIMATE_DOCKER 1
```
wfleming: Am I correct that this is only false for local dev/local tests? If that's the case, I'd suggest dropping it completely & making sure tests provide appropriate ENV. Since docker is the CLI's natural environ, special-casing for the other case (esp. if it affects tests) seems like it makes tests less honest.

pbrisbin: If not being run via docker, we can't use the container-side paths of `/code` and `/tmp/cc`, we have to set them as whatever the host side is. I couldn't find any way to do that without:

- Having two new environment variables for the container side (with defaults) making 4 total
- Having a single environment variable to represent the single conditional (as I've done)

I may be missing something; if you have something in mind, please describe it a little more concretely.

wfleming: Bleh, you're right, I didn't think it through all the way. Yeah, this seems like the less troubling option.

wfleming:
This is great! The `MountedPath` is a nice abstraction I can't believe we didn't see earlier.

The `test` functionality isn't very used now, and needs real work to be legitimately useful at any rate, so I wouldn't worry about it excessively. That said, we shouldn't break existing functionality even if it's not widely used, so we should do whatever needs doing to figure out if this is breaking it. @dblandin tried using it for something, so he might have a simple testcase.

dblandin:
> The test functionality isn't very used now, and needs real work to be legitimately useful at any rate, so I wouldn't worry about it excessively.

I wouldn't worry about it. It's not used at the moment.

This LGTM and addresses a few open issues including #245, #246, and #299.

PR #372, opened 2016-03-09
Add ShellCheck engine definition

pbrisbin:
/cc @codeclimate/review

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -249,6 +249,16 @@ rubymotion:
   enable_regexps:
   default_ratings_paths:
     - "**.rb"
+
+shellcheck:
+  image: codeclimate/codeclimate-shellcheck
+  description: A static analysis tool for shell scripts
+  community: true
+  enable_regexps:
+    - \.sh$
+  default_ratings_paths:
+    - "**.sh"
+
 scss-lint:
   image: codeclimate/codeclimate-scss-lint
   description: Configurable tool for writing clean and consistent SCSS.
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -249,16 +249,14 @@ rubymotion:
   enable_regexps:
   default_ratings_paths:
     - "**.rb"
-
 shellcheck:
   image: codeclimate/codeclimate-shellcheck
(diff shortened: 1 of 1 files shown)


config/engines.yml
```diff
@@ -249,6 +249,16 @@ rubymotion:
   enable_regexps:
   default_ratings_paths:
     - "**.rb"
+
```
wfleming: Nit, but right now we don't have spaces between engine definitions. Seems like a nice addition to me, but if we're going to do it I think we should do all of them at once.

config/engines.yml
```diff
@@ -249,6 +249,16 @@ rubymotion:
...
+shellcheck:
+  image: codeclimate/codeclimate-shellcheck
+  description: A static analysis tool for shell scripts
```
wfleming: :lipstick: description fields should be full sentences with periods at the end.

config/engines.yml
pbrisbin: My bad, I thought I saw space between nearby ones. I was seeing things.

config/engines.yml
pbrisbin: I can add the period, are you also saying I need to make this a grammatically correct sentence? "ShellCheck is a..."

config/engines.yml
```diff
@@ -249,6 +249,16 @@ rubymotion:
   enable_regexps:
   default_ratings_paths:
     - "**.rb"
+
+shellcheck:
```
wfleming: I think this should be sorted after `scss-lint`, shouldn't it?

pbrisbin: I matched builder's assuming that one was sorted; guess not. Will fix here at least.

config/engines.yml
wfleming: Good point. Other descriptions in the file are currently fragments, so for now I think just adding the period is fine, and we can wait for @scriberty to decide it's bad & needs changing.

config/engines.yml
wfleming: Oops. Builder might have been my bad, sorry.

wfleming:
LGTM

PR #374, opened 2016-03-16
Exclude files that begin with a dash

GordonDiggs:
These cause problems for the `find` command we run, so it's best we just
exclude them.

@codeclimate/review


pbrisbin:
Seems reasonable. I wish `find` had something like the `--` argument separator, but it doesn't and from what I can find online, working around this can get pretty hacky.

PR #377, opened 2016-03-16
Export CODECLIMATE variables in wrapper script

pbrisbin:
I'm not sure what's changed but it seems these variables need to be exported
(not just set) for docker to find them when using the value-less --env form.

/cc @codeclimate/review


GordonDiggs:
LGTM

dblandin:
This fix works for me :+1:

PR #380, opened 2016-03-16
Use Analyzer.logger within Analyzer classes

pbrisbin:
Using CLI.debug caused an awkward scenario where Builder had to require cc/cli
to bring it into scope when using Analyzer classes. This change uses the
existing Analyzer.logger interface to provide the debug functionality. Said
functionality now also uses print and bookends messages with newlines to avoid
mangling due to threading with the analysis spinner.

Note: The CLI logger only provides debug, because in the context of CLI any
non-DEBUG output should be displayed via normal, user-friendly channels.

/cc @codeclimate/review

### DIFF
--- a/lib/cc/cli.rb
+++ b/lib/cc/cli.rb
@@ -17,15 +17,25 @@ module CC
     autoload :ValidateConfig, "cc/cli/validate_config"
     autoload :Version, "cc/cli/version"
 
-    def self.debug(message, values = {})
-      if ENV["CODECLIMATE_DEBUG"]
-        if values.any?
-          message << " "
-          message << values.map { |k, v| "#{k}=#{v.inspect}" }.join(" ")
-        end
+    class Logger
+      def debug(message, values = {})
+        if ENV["CODECLIMATE_DEBUG"]
+          if values.any?
+            message << " "
+            message << values.map { |k, v| "#{k}=#{v.inspect}" }.join(" ")
+          end
 
-        $stderr.puts("[DEBUG] #{message}")
+          # Leading and trailing newlines to prevent mangling
+          $stderr.print("\n[DEBUG] #{message}\n")
+        end
+      end
+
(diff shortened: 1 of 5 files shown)


lib/cc/cli.rb
```diff
@@ -17,15 +17,25 @@ module CLI
     autoload :ValidateConfig, "cc/cli/validate_config"
     autoload :Version, "cc/cli/version"
 
-    def self.debug(message, values = {})
-      if ENV["CODECLIMATE_DEBUG"]
-        if values.any?
-          message << " "
-          message << values.map { |k, v| "#{k}=#{v.inspect}" }.join(" ")
+    class Logger
```
dblandin: Is it worth moving this to its own file? `cc/cli/logger`?

dblandin:
One minor comment. LGTM to me either way.

lib/cc/cli.rb
```diff
@@ -17,15 +17,25 @@ module CLI
...
+          end
+
+          # Leading and trailing newlines to prevent mangling
+          $stderr.print("\n[DEBUG] #{message}\n")
         end
+      end
 
-        $stderr.puts("[DEBUG] #{message}")
+      # Any non-DEBUG output is handled not via logging
+      def method_missing(*)
```
wfleming: Is this actually used right now?  Since `#debug` doesn't follow the same contract as ruby's `Logger`, I don't see a reason for a shim like this unless we already have usage relying on it (which I don't believe we do).

pbrisbin: There are other calls to `Analyzer.logger.{warn,info}` reachable by the CLI, so this shim is necessary. The interface mismatch is worth thinking about though.

pbrisbin:
@wfleming brings up a good point. I'm going to close this and leave the weird require in Builder for now, to be cleaned up by the sub-command rewrite of CLI.

PR #382, opened 2016-03-18
Remove check as gem executable

### LINKED ISSUE #381: Check is marked as executable within gemspec
The 'check' command is marked as an executable within the gemspec. However, this is solely for executable ruby files, stated here: http://guides.rubygems.org/specification-reference/#executables

This is an issue when tooling that analyzes gemspecs assumes the executables listed are ruby scripts.

GordonDiggs:
Executables should only be ruby scripts as per
http://guides.rubygems.org/specification-reference/#executables

Fixes #381

@codeclimate/review


jpignata:
:+1:

PR #384, opened 2016-03-18
Revert "Install codeclimate-wrapper from image"

pbrisbin:
This reverts commit e26b09eaa5e57445f63d8aba71dee95cb507a2b0.

This doesn't work on docker-machine because the host path is in the VM and we
need to install locally. We'll have to take the risk of getting the wrapper out
of sync.

/cc @codeclimate/review


GordonDiggs:
LGTM

PR #386, opened 2016-03-18
Allow Marketing site updates to fail

pbrisbin:
Instead of dying, print a noticable warning. The automated step fails on my
machine for a strange zsh+chruby issue, but I'm happy to run it manually.

/cc @codeclimate/review


wfleming:
:+1:

PR #388, opened 2016-03-21
Require YAML in Analyzer and CLI

dblandin:
YAML is not loaded by default and now that we're using `.safe_load` from Psych
instead of the `safe_yaml` gem, we have load it explicitly.

```console
$ codeclimate analyze
error: (NameError) uninitialized constant CC::Analyzer::EngineRegistry::YAML
```


GordonDiggs:
:+1:

PR #398, opened 2016-04-01
Ensure Workspace#clone clones everything

### LINKED ISSUE #397: Engine-specific wildcard and multi-level exclude_paths are globally applied.
I'm trying to disable just the `duplication` engine for all nested `test` directories in a project, but it appears that wildcard exclusions and multi-level paths get applied to all engines.

This yaml produces (for example) 16 warnings.
``` yaml
engines:
  duplication:
    enabled: true
    exclude_paths:
    - tasks/
    config:
      languages:
      - ruby
  reek:
    enabled: true
ratings:
  paths:
  - "**.rb"
exclude_paths:
- "**/node_modules/**"
- tmp/
```

If I replace `tasks/` with `"**/test/**"` or `tasks/filter-team-data/test`, I receive 15 warnings. The additional warning that is filtered is a reek warning on `tasks/filter-team-data/test/test_team_data_filterer.rb`.

I'm surprised to see a reek warning change based on a modification to `duplication`'s `exclude_paths`. Am I doing something wrong?

I am running `version 0.24.0` on https://github.com/18F/concourse-compliance-testing.

wfleming:
`Hash#dup` does not `dup` values: I don't know why I thought it did. As a
result, excluding files deeper than 1 level in the workspace would apply
to the global workspace copy, not only the engine-specific workspace copy.

The fix here manually deep-clones the `children` hash, and `freezes`
some keys for good measure.

cc @codeclimate/review. This is intended to fix #397.

### DIFF
--- a/spec/cc/workspace_spec.rb
+++ b/spec/cc/workspace_spec.rb
@@ -90,6 +90,8 @@ module CC
           lib/foo/baz.rb
           lib/quix/a.rb
           lib/quix/b.rb
+          skeleton/torso/ribcage/heart.rb
+          skeleton/torso/belly/liver.rb
           spec/foo/bar_spec.rb
           spec/foo/baz_spec.rb
           spec/foo_spec.rb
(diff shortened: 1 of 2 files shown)


spec/cc/workspace_spec.rb
```diff
@@ -90,6 +90,8 @@ module CC
           lib/foo/baz.rb
           lib/quix/a.rb
           lib/quix/b.rb
+          skeleton/torso/ribcage/heart.rb
+          skeleton/torso/belly/liver.rb
```
pbrisbin: I approve of this fixture scheme

pbrisbin:
LGTM

afeld:
That was fast! :clap:

PR #399, opened 2016-04-01
Don't try to show spinner when debug is on

wfleming:
The TTY spinner doesn't play well with debug output, and can create some odd interspersing of text.

Skipping the spinner results in the "Running..." status line not being output at all, so
I added the engine name to the relevant debug lines for clarity.

cc @codeclimate/review


jpignata:
lgtm

PR #403, opened 2016-04-05
Don't crash `init` in an empty directory.

wfleming:
Running `init` in an empty directory correctly set up an empty config
yml, but then crashed trying to look for engine configs.

cc @codeclimate/review

### DIFF
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -64,9 +64,8 @@ module CC
       end
 
       def available_engine_configs
-        engine_names = existing_cc_config.engines.select do |_, config|
-          config.enabled?
-        end.keys
+        engines = existing_cc_config.engines || {}
+        engine_names = engines.select { |_, config| config.enabled? }.keys
 
         all_paths = engine_names.flat_map do |engine_name|
           engine_directory = File.expand_path("../../../../config/#{engine_name}", __FILE__)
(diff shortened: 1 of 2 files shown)


lib/cc/cli/init.rb
```diff
@@ -64,9 +64,8 @@ def create_default_engine_configs
       end
 
       def available_engine_configs
-        engine_names = existing_cc_config.engines.select do |_, config|
-          config.enabled?
-        end.keys
+        engines = existing_cc_config.engines || {}
```
GordonDiggs: `existing_cc_config` returns nil when the file doesn't exist, right? So calling `engines` on that would raise an error?

wfleming: This was discussed a bit on [#312](https://github.com/codeclimate/codeclimate/pull/312#discussion_r50721192) when this was implemented. `existing_cc_config` is guaranteed to exist when this is called: either it will have been generated by this class just now, or it already exists (and was upgraded from classic, if it needed to be).

The crash that did occur was that `engines` is `nil` instead of `{}` in `CC::Yaml` when it's empty. Maybe that should be fixed instead.

GordonDiggs: Ah cool. Makes sense! I think returning nil is fine. Coercing it into `{}` might harm us more down the road.

GordonDiggs:
LGTM

PR #404, opened 2016-04-05
Debug mode improvements

wfleming:
The stderr here is a straight feature addition that I think will be important for the usefulness of the debug output.

The changes to string-escaping are a bit trickier, and probably merit some discussion. When all the output was escaped using `#inspect`, I personally found it difficult to read directly in the terminal.
It also made it harder to copy-paste directly into a JSON file. I recognize that the probable intent was to make it very easy to copy into a REPL, but since Ruby allows using `%[]` instead of quotes for literal strings, I don't think making the output easier to read/use elsewhere blocks that use case. The current choice of manually-escaping newlines/tabs might not be quite right, though. Thoughts, @codeclimate/review?

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -34,7 +34,7 @@ module CC
         )
 
         container.on_output("\0") do |raw_output|
-          CLI.debug "#{name} engine output: #{raw_output.inspect}"
+          CLI.debug("#{name} engine output: #{raw_output.strip}")
           output = EngineOutput.new(raw_output)
 
           unless output_filter.filter?(output)
@@ -43,8 +43,10 @@ module CC
         end
 
         write_config_file
-        CLI.debug "#{name} engine config: #{config_file.read.inspect}"
-        container.run(container_options)
+        CLI.debug("#{name} engine config: #{config_file.read}")
+        container.run(container_options).tap do |result|
+          CLI.debug("#{name} engine stderr: #{result.stderr}")
+        end
       ensure
         delete_config_file
       end
--- a/spec/cc/analyzer/engine_spec.rb
+++ b/spec/cc/analyzer/engine_spec.rb
@@ -6,7 +6,9 @@ module CC::Analyzer
       it "passes the correct options to Container" do
         container = double
         allow(container).to receive(:on_output).and_yield("")
-        allow(container).to receive(:run)
+        allow(container).to receive(:run).and_return(
+          Container::Result.new(0, false, 1, false, 10, ""),
+        )
 
         expect(Container).to receive(:new) do |args|
           expect(args[:image]).to eq "codeclimate/foo"
@@ -32,7 +34,7 @@ module CC::Analyzer
           "--rm",
           "--volume", "/code:/code:ro",
           "--user", "9000:9000",
-        ))
+        )).and_return(Container::Result.new(0, false, 1, false, 10, ""))
 
         expect(Container).to receive(:new).and_return(container)
         engine = Engine.new("", {}, "/code", {}, "a-label")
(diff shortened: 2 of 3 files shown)


lib/cc/analyzer/engine.rb
```diff
@@ -43,8 +43,10 @@ def run(stdout_io, container_listener)
         end
 
         write_config_file
-        CLI.debug "#{name} engine config: #{config_file.read.inspect}"
-        container.run(container_options)
+        CLI.debug("#{name} engine config: #{config_file.read}")
+        container.run(container_options).tap do |result|
+          CLI.debug("#{name} engine stderr: #{result.stderr}")
```
pbrisbin: :clap: I've always wanted this.

lib/cc/analyzer/engine.rb
```diff
@@ -34,7 +34,7 @@ def run(stdout_io, container_listener)
         )
 
         container.on_output("\0") do |raw_output|
-          CLI.debug "#{name} engine output: #{raw_output.inspect}"
+          CLI.debug("#{name} engine output: #{raw_output.gsub("\n", "\\n").gsub("\t", "\\t")}")
```
GordonDiggs: WDYT of `gsub(/[\n\t]/, " ")`?

pbrisbin:
I'm not sure if I have an opinion on the string escaping. I'd have to see some screenshots, I think. If that's too annoying though, I'm fine merging and discussing later (or just living with it) if I find I don't like it.

So... pretty much LGTM.

lib/cc/analyzer/engine.rb
GordonDiggs: There's also `JSON.pretty_generate` but I think you want it to all be one line?

wfleming: > WDYT of gsub(/[\n\t]/, " ")?

Could do, but I'm hesitant to eliminate information about the *kind* of whitespace present in something calling itself a debug mode. That whitespace could also be present *within* the issue, where it's kind may be more important to interpreting the results.

> There's also JSON.pretty_generate but I think you want it to all be one line?

We definitely want this on one line. Also, this debug output is before anything else has tried to parse the output has JSON, I believe. I wouldn't want invalid-engine-output to start raising from here instead of from where we want it to.

GordonDiggs: Cool - all good points. What you have seems like a good way to go for now

GordonDiggs:
LGTM pending green builds

lib/cc/analyzer/engine.rb
wfleming: We could simply not escape it: I'm not escaping it in stderr below right now.

The only downside of that, for me, is it can be less obvious where engine output begins or ends when it has leading or trailing whitespace.

GordonDiggs: You could wrap it in quotes to make that more evident if that would help

wfleming: > You could wrap it in quotes to make that more evident if that would help

I was considering that, but I think we need a delimiter other than quotes, since pretty much all engine output will contain quotes.

wfleming:
@pbrisbin 

Before (escaping everything):
screen shot 2016-04-05 at 11 30 19

After (not escaping everything):
screen shot 2016-04-05 at 11 30 28

wfleming:
Tweaked it a bit, and decided to just strip engine output before echoing without doing any other manipulation: newlines in issue descriptions are already escaped (because it has to be valid JSON). We handle leading/trailing whitespace before we parse the issues, so it doesn't seem very useful for debugging anyway.

@pbrisbin @GordonDiggs any problems with that thinking?

GordonDiggs:
Works for me! :shipit:

PR #406, opened 2016-04-10
Fix analyzer path bug

rolandhawk:
When trying to add multiple paths with common parent, somehow some of
the paths are not added correctly. This is because the parent node is
always initialized even if it has been initialized before.

Example:
I have these files:
````
.git/FETCH_HEAD
.git/refs/heads/master
code/a/bar.rb
code/a/baz.rb
code/foo.rb
foo.txt
lib/thing.rb
````
Then run:
````
codeclimate analyze code/foo.rb code/a/bar.rb
````
It only evaluates `code/a/bar.rb`


wfleming:
Thanks for catching this, @rolandhawk!

PR #408, opened 2016-04-11
Rename ESLint config file in light of filename deprecation

dblandin:
The `.eslintrc` file, which can be either JSON or YAML, has been deprecated in favor of more explicit filenames. Documentation can be found [here][].

This PR renames the default config file to `.eslintrc.yml`.

[here]: http://eslint.org/docs/user-guide/configuring#configuration-file-formats


GordonDiggs:
Can we use the `.js` one? It has the highest precedence (and is what we suggest in our styleguide)

dblandin:
> Can we use the .js one? It has the highest precedence (and is what we suggest in our styleguide)

Sounds good to me. I'll convert the config to JS.

GordonDiggs:
@dblandin Something to think about: how will this change the inferred config experience for people?

wfleming:
This has been discussed before: we explicitly stuck with `.eslintrc` in the past because it has the *lowest* precedence, ensuring we would never overwrite a user's explicit settings.

wfleming:
The problem is that we run `init` on .com even if a `.codeclimate.yml` is present, because we want to provide our default engine configs when explicit configs are not present.

The copy-configs code currently pretty much just says "copy each file in a directory if it doesn't already exist". In the ESLint case, that's problematic because there are 5 standard files they could have configuration in, and older configs are likely to be the older `.eslintrc` name (since I think that used to be the only name used). That's why we've continued to name our config `.eslintrc`: it's a less optimal first-run setup since we're giving the user a deprecated filename, but it won't stomp over explicit user config.

I don't think we can reasonably change this without first adding features to to the copy-configs code  to allow more sophisticated conditions for copying a file.

dblandin:
> we intentionally stuck with .eslintrc in the past because it has the lowest precedence
> ...
> I don't think we can reasonably change this without first adding features to to the copy-configs code to allow more sophisticated conditions for copying a file.

Makes sense to me. Thanks for the context! I'll go ahead and close this PR.

PR #409, opened 2016-04-15
Issue Validations

GordonDiggs:
This ports over the issue validations we have in Builder, as well as adding 3 new ones:

- Making sure a reported file exists
- Making sure the categories reported are valid
- Making sure the description is present (required by the spec)

@codeclimate/review

### DIFF
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -15,6 +15,7 @@ module CC
     autoload :Filesystem, "cc/analyzer/filesystem"
     autoload :Formatters, "cc/analyzer/formatters"
     autoload :Issue, "cc/analyzer/issue"
+    autoload :IssueCategoryValidation, "cc/analyzer/issue_category_validation"
     autoload :IssueCheckNamePresenceValidation, "cc/analyzer/issue_check_name_presence_validation"
     autoload :IssuePathPresenceValidation, "cc/analyzer/issue_path_presence_validation"
     autoload :IssueRelativePathValidation, "cc/analyzer/issue_relative_path_validation"
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -18,6 +18,7 @@ module CC
     autoload :IssueCategoryValidation, "cc/analyzer/issue_category_validation"
     autoload :IssueCheckNamePresenceValidation, "cc/analyzer/issue_check_name_presence_validation"
     autoload :IssueDescriptionPresenceValidation, "cc/analyzer/issue_description_presence_validation"
+    autoload :IssuePathExistenceValidation, "cc/analyzer/issue_path_existence_validation"
     autoload :IssuePathPresenceValidation, "cc/analyzer/issue_path_presence_validation"
     autoload :IssueRelativePathValidation, "cc/analyzer/issue_relative_path_validation"
     autoload :IssueSorter, "cc/analyzer/issue_sorter"
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -17,6 +17,7 @@ module CC
     autoload :Issue, "cc/analyzer/issue"
     autoload :IssueCategoryValidation, "cc/analyzer/issue_category_validation"
     autoload :IssueCheckNamePresenceValidation, "cc/analyzer/issue_check_name_presence_validation"
+    autoload :IssueDescriptionPresenceValidation, "cc/analyzer/issue_description_presence_validation"
     autoload :IssuePathPresenceValidation, "cc/analyzer/issue_path_presence_validation"
     autoload :IssueRelativePathValidation, "cc/analyzer/issue_relative_path_validation"
     autoload :IssueSorter, "cc/analyzer/issue_sorter"
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -15,13 +15,19 @@ module CC
     autoload :Filesystem, "cc/analyzer/filesystem"
     autoload :Formatters, "cc/analyzer/formatters"
     autoload :Issue, "cc/analyzer/issue"
+    autoload :IssueCheckNamePresenceValidation, "cc/analyzer/issue_check_name_presence_validation"
+    autoload :IssuePathPresenceValidation, "cc/analyzer/issue_path_presence_validation"
+    autoload :IssueRelativePathValidation, "cc/analyzer/issue_relative_path_validation"
     autoload :IssueSorter, "cc/analyzer/issue_sorter"
+    autoload :IssueTypeValidation, "cc/analyzer/issue_type_validation"
+    autoload :IssueValidator, "cc/analyzer/issue_validator"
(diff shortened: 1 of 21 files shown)


lib/cc/analyzer/issue_category_validation.rb
```diff
@@ -0,0 +1,30 @@
+module CC
+  module Analyzer
+    class IssueCategoryValidation < Validation
+      CATEGORIES = [
+        "Bug Risk",
```
wfleming: I think we're supposed to freeze these individual values.

lib/cc/analyzer.rb
```diff
@@ -15,13 +15,22 @@ module Analyzer
...
     autoload :IssueSorter, "cc/analyzer/issue_sorter"
+    autoload :IssueTypeValidation, "cc/analyzer/issue_type_validation"
+    autoload :IssueValidator, "cc/analyzer/issue_validator"
     autoload :LocationDescription, "cc/analyzer/location_description"
     autoload :LoggingContainerListener, "cc/analyzer/logging_container_listener"
     autoload :MountedPath, "cc/analyzer/mounted_path"
     autoload :RaisingContainerListener, "cc/analyzer/raising_container_listener"
     autoload :SourceBuffer, "cc/analyzer/source_buffer"
     autoload :StatsdContainerListener, "cc/analyzer/statsd_container_listener"
+    autoload :Validation, "cc/analyzer/validation"
```
ABaldwinHunter: What do you think about autoloading all of the other validators in `validation.rb`?

pbrisbin:
LGTM

lib/cc/analyzer.rb
GordonDiggs: If we move them under that namespace that would work, but these need to be autoloaded under the module to which they belong (in this case, `Analyzer`)

lib/cc/analyzer/issue_validator.rb
```diff
@@ -0,0 +1,49 @@
+module CC
+  module Analyzer
+    class IssueValidator
+      CHECKS = [
+        IssueCategoryValidation,
+        IssueCheckNamePresenceValidation,
+        IssueDescriptionPresenceValidation,
+        IssuePathExistenceValidation,
```
wfleming: There's an interesting ordering issue here that might prompt us to not sort all of these alphabetically: checking for path presence *after* checking for path existence seems wrong to me.

I think either we should re-order these to `presence -> relative -> existence` (and add a comment explaining why), or collapse all of these into a single `IssuePathValidation`.

lib/cc/analyzer/issue_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+      ].freeze
+
+      attr_reader :error
+
+      def initialize(issue)
+        @issue = issue
+      end
+
+      def valid?
+        @error.blank?
```
wfleming: API nit: most `ActiveModel`-esque interfaces allow you to call `#valid?` without explicitly calling `validate`. Should we do the same? It's our own little world, so we don't have to, but OTOH it seems like something likely to trip somebody up down the road, and IMHO it is a nice little feature.

lib/cc/analyzer/issue_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      private
+
+      attr_reader :issue
+
+      def invalid_messages
+        @invalid_messages ||= CHECKS.each_with_object([]) do |check, result|
+          validator = check.new(issue)
+          result << validator.message unless validator.valid?
+          CLI.debug("#{check} #{validator.valid?}")
```
wfleming: Is this worth keeping after shipping this? Won't the CLI very quickly be echoing the above messages telling you that it's invalid?

lib/cc/analyzer/issue_validator.rb
GordonDiggs: When we first introduced these, we made the choice that they should 

1. Not be order dependent
2. Be able to run independently of each other
3. All be run all the time

This keeps the classes small (responding to just a `valid?` and `message` method). See that discussion [here](https://github.com/codeclimate/builder/pull/77#discussion_r34052959).

I don't like the idea of doing multiple checks in one validation class - I think it will cause the classes to become bloated and do too many things (additionally, having a single coherent message to be returned from them will be harder)

lib/cc/analyzer/issue_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      def validate
+        if issue && invalid_messages.any?
+          @error = {
+            message: "#{invalid_messages.join("; ")}: `#{issue}`.",
+            issue: issue,
+          }
+          false
+        else
+          true
```
wfleming: I think there's a not-great edge case here for issues that are unparseable: the `IssueValidator` get's initialized with `nil`, which then hits this branch because `issue && invalid_messages.any?` also evaluates to `nil`. Is the unparseable case still covered somewhere that I'm just not seeing in the diff?

wfleming: From an API design perspective, I guess I also question whether considering a `nil` issue "valid" is semantically a good idea. Personally, I don't think it is.

lib/cc/analyzer/issue_validator.rb
wfleming: Fair enough. I see the other two validations are well guarded against `nil` paths, so that makes sense to me.

wfleming:
The original card prompting this work was about path validations, and I could be wrong, but I think one of the things around that was not just that the *file path* could be invalid, but also that the *location* (line numbers/offsets) could be invalid. Are you planning to investigate/address that separately, or did you conclude it wasn't necessary?

lib/cc/analyzer/issue_validator.rb
GordonDiggs: Yeah the unparseable case is handled [elsewhere](https://github.com/codeclimate/codeclimate/blob/gd-issue-validation/lib/cc/analyzer/issue.rb#L50) (we let JSON.parse raise in this repo). In builder we handle it separately. Because it's a different error code (E12 vs E13), I think not having a `Parseable JSON` validation is better

wfleming: Sounds good. I agree about not having an extra validation be better. I had seen [a rescue](https://github.com/codeclimate/codeclimate/pull/409/files#diff-d5f814a23156b4c15e5dfdef6074109eL27) in the diff, and forgot that we do that parse elsewhere *without* the rescue.

wfleming: Still a bit squirrely out about considering `nil` to be a valid issue, though :)

GordonDiggs:
@codeclimate/review ready for another look!

GordonDiggs:
@wfleming ah that wasn't clear from the card. I can do that validation in a follow-up PR!

wfleming:
My only remaining nit is that I think intentionally calling `IssueValidator.new(nil)` is a smell. I'd love to see that not be the case eventually, but I'm fine with it not being done now. Aside from that, this is awesome!

PR #412, opened 2016-04-15
Add validation that the positions for an issue are valid

GordonDiggs:
Following the directions from [the spec][1]

[1]: https://github.com/codeclimate/spec/blob/master/SPEC.md#locations

@codeclimate/review

### DIFF
--- a/lib/cc/analyzer/issue_location_existence_validation.rb
+++ b/lib/cc/analyzer/issue_location_existence_validation.rb
@@ -1,0 +1,59 @@
+module CC
+  module Analyzer
+    class IssueLocationExistenceValidation < Validation
+      def valid?
+        if location["lines"]
+          valid_lines?
+        elsif location["positions"]
+          location["positions"].is_a?(Hash) &&
+            valid_position?(location["positions"]["begin"]) &&
+            valid_position?(location["positions"]["end"])
+        else
+          false
+        end
+      end
+
+      def message
+        "Location does not exist in the file"
+      end
+
+      private
+
+      def location
+        object.fetch("location", {})
+      end
+
+      def contents
+        @contents ||= location["path"] && File.read(location["path"])
+      end
+
+      def lines
+        @lines ||= contents.lines
+      end
+
+      def valid_lines?
+        start = location["lines"]["begin"]
+        stop = location["lines"]["end"]
+
+        start && stop && valid_line?(start) && valid_line?(stop)
+      end
+
+      def valid_position?(position)
+        (position["offset"] && valid_offset?(position["offset"])) ||
+          (position["line"] && position["column"] && valid_linecol?(position["line"], position["column"]))
+      end
+
+      def valid_offset?(offset)
+        !contents[offset].nil?
+      end
+
+      def valid_line?(line)
+        !lines[line - 1].nil?
+      end
+
+      def valid_linecol?(line, column)
+        (text = lines[line - 1]) && !text[column - 1].nil?
+      end
+    end
+  end
+end
--- a/lib/cc/analyzer/issue_validator.rb
+++ b/lib/cc/analyzer/issue_validator.rb
@@ -5,6 +5,7 @@ module CC
         IssueCategoryValidation,
         IssueCheckNamePresenceValidation,
         IssueDescriptionPresenceValidation,
+        IssueLocationExistenceValidation,
         IssuePathExistenceValidation,
         IssuePathPresenceValidation,
         IssueRelativePathValidation,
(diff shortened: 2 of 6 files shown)


lib/cc/analyzer/issue_validator.rb
```diff
@@ -5,6 +5,7 @@ class IssueValidator
         IssueCategoryValidation,
         IssueCheckNamePresenceValidation,
         IssueDescriptionPresenceValidation,
+        IssueLocationExistenceValidation,
```
ABaldwinHunter: I slightly like using `presence` throughout here, instead of mixing with `existence`, but looks like we've already mixed a bit so, perhaps that's fine. WDYT?

ABaldwinHunter:
One question that might be outside of scope: have we considered how we might handle permitting valid issues without location in the future - like a security vulnerability that says you're missing x file. ?

Other than that LGTM

lib/cc/analyzer/issue_location_existence_validation.rb
```diff
@@ -0,0 +1,56 @@
...
+          valid_lines?
+        elsif location["positions"]
+          valid_position?(location["positions"]["begin"]) &&
+            valid_position?(location["positions"]["end"])
+        end
+      end
+
+      def message
+        "Location does not exist in the file"
+      end
```
ABaldwinHunter: Maybe this could be a constant?

lib/cc/analyzer/issue_validator.rb
GordonDiggs: Presence means "is it present in the output?", whereas existence means "does it actually exist where the output says it is?"

lib/cc/analyzer/issue_location_existence_validation.rb
GordonDiggs: Eh, I don't see a big win in that - do you? It has to be a method in order to conform to the `Validator` format

GordonDiggs:
> have we considered how we might handle permitting valid issues without location in the future - like a security vulnerability that says you're missing x file. ?

This will likely have to change to accomodate that, but we also would have to do a fair amount of work elsewhere to support it

lib/cc/analyzer/issue_validator.rb
ABaldwinHunter: Got it. Thanks

lib/cc/analyzer/issue_location_existence_validation.rb
ABaldwinHunter: 👍  makes sense

lib/cc/analyzer/issue_location_existence_validation.rb
```diff
@@ -0,0 +1,56 @@
+module CC
+  module Analyzer
+    class IssueLocationExistenceValidation < Validation
+      def valid?
+        if location["lines"]
+          valid_lines?
+        elsif location["positions"]
```
wfleming: Should there be an `else` after this to add an error in the event that the location is completely invalid and doesn't have either? Is that covered anywhere else?

lib/cc/analyzer/issue_location_existence_validation.rb
```diff
@@ -0,0 +1,56 @@
...
+          valid_position?(location["positions"]["begin"]) &&
```
wfleming: There's a `NoMethodError` risk if an engine does something really dumb like emit `positions: "foo"`. Maybe not worth addressing.

lib/cc/analyzer/issue_location_existence_validation.rb
wfleming: Oh, wait, you're implicitly returning `nil` if neither of these checks is `true`, which counts as not valid. Right? Nevermind, then.

wfleming:
This looks good to me. One thing I'd be interested in doing some testing of before shipping is performance impact on some repos: you're reading in the file afresh for each issue, which could fairly seriously slow down analysis on largish repos/repos with many issues. It'd be nice to have some assurance that impact is acceptable.

lib/cc/analyzer/issue_location_existence_validation.rb
GordonDiggs: Right - adding the explicit false though just because I like it.

lib/cc/analyzer/issue_location_existence_validation.rb
GordonDiggs: Good call - updated!

GordonDiggs:
@wfleming when you have a moment, I'm curious about your thoughts on https://codeclimate.slack.com/archives/dev/p1461000692000097

Discovered that behavior while trying to test the performance impact of this

GordonDiggs:
Going to close this and re-work. We decided that asserting that the format is correct is good enough

PR #416, opened 2016-04-18
Add validations for location and other_locations

GordonDiggs:
Make sure that they match the formats outlined in the spec:
https://github.com/codeclimate/spec/blob/master/SPEC.md#locations

@codeclimate/review

### DIFF
--- a/lib/cc/analyzer/issue_other_locations_format_validation.rb
+++ b/lib/cc/analyzer/issue_other_locations_format_validation.rb
@@ -1,0 +1,33 @@
+module CC
+  module Analyzer
+    class IssueOtherLocationsFormatValidation < Validation
+      CHECKS = [
+        IssueLocationFormatValidation,
+        IssuePathExistenceValidation,
+        IssuePathPresenceValidation,
+        IssueRelativePathValidation,
+      ].freeze
+
+      def valid?
+        if object["other_locations"]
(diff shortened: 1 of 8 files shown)


wfleming:
Nit: we're not validating the `path` of `other_locations` entries, are we? Should we?

GordonDiggs:
@wfleming ah yeah we should. I'll update

GordonDiggs:
@wfleming updated!

lib/cc/analyzer/issue_other_locations_format_validation.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      def message
+        "Other locations are not formatted correctly"
+      end
+
+      private
+
+      def other_location_valid?(location)
+        [IssueLocationFormatValidation, IssuePathExistenceValidation, IssuePathPresenceValidation].all? do |klass|
```
wfleming: Should we also do the `IssueRelativePathValidation` here?

wfleming:
One note about another validation you probably want in the `other_locations` stack. Otherwise, LGTM.

lib/cc/analyzer/issue_other_locations_format_validation.rb
GordonDiggs: Yep!

PR #419, opened 2016-04-22
Add SourceFingerprint based on path, check_name, and source

dblandin:
Currently, the default fingerprint for an issue is computed by creating an MD5 digest based on the path of the file which introduced the issue and the check name for the issue emitted.

The result of this fingerprint for many engines is that issues emitted with the same check name for the same file are given the same fingerprint. On codeclimate.com, we only show issues with unique fingerprints, so even if 100 issues were emitted for a file, if all issues share the same check name and thus the same fingerprint, we only show one issue.

Additionally, a fixed or introduced issue is often ignored if there are current issues with the same fingerprint already on the default branch.

This PR introduces new fingerprinting behavior to address these issues, first added to a [couple][markdownlint] [engines][shellcheck], that incorporates relevant source code into the fingerprint. ~~This implementation allows us to enable the new fingerprinting behavior with an environment variable so we can roll it out gradually.~~ Eventually, this will become the new default fingerprinting method for engines.

[markdownlint]: https://github.com/jpignata/codeclimate-markdownlint
[shellcheck]: https://github.com/filib/codeclimate-shellcheck

### DIFF
--- a/lib/cc/analyzer.rb
+++ b/lib/cc/analyzer.rb
@@ -31,6 +31,8 @@ module CC
     autoload :MountedPath, "cc/analyzer/mounted_path"
     autoload :RaisingContainerListener, "cc/analyzer/raising_container_listener"
     autoload :SourceBuffer, "cc/analyzer/source_buffer"
+    autoload :SourceExtractor, "cc/analyzer/source_extractor"
+    autoload :SourceFingerprint, "cc/analyzer/source_fingerprint"
     autoload :StatsdContainerListener, "cc/analyzer/statsd_container_listener"
     autoload :Validation, "cc/analyzer/validation"
 
--- a/lib/cc/analyzer/issue.rb
+++ b/lib/cc/analyzer/issue.rb
@@ -34,24 +34,20 @@ module CC
         end
       end
 
+      def path
+        parsed_output.fetch("location", {}).fetch("path", "")
+      end
+
       private
 
       attr_reader :output
 
       def default_fingerprint
-        digest = Digest::MD5.new
-        digest << path
-        digest << "|"
-        digest << check_name.to_s
-        digest.to_s
+        SourceFingerprint.new(self).compute
       end
 
       def parsed_output
         @parsed_output ||= JSON.parse(output)
-      end
-
-      def path
-        parsed_output.fetch("location", {}).fetch("path", "")
       end
     end
   end
--- a/lib/cc/analyzer/source_extractor.rb
+++ b/lib/cc/analyzer/source_extractor.rb
@@ -1,0 +1,79 @@
+module CC
+  module Analyzer
+    class SourceExtractor
+      InvalidLocationPositions = Class.new(StandardError)
+
+      def initialize(source)
+        @source = source
+      end
+
+      def extract(location)
+        if (lines = location["lines"])
+          extract_from_lines(lines)
+        elsif (positions = location["positions"])
+          extract_from_positions(positions)
+        end
+      end
+
+      private
+
+      attr_reader :location, :source
+
+      def extract_from_lines(lines)
+        begin_index = lines.fetch("begin") - 1
+        end_index = lines.fetch("end") - 1
+        range = (begin_index..end_index)
+
+        source.each_line.with_object("").with_index do |(source_line, memo), index|
+          memo << source_line if range.include?(index)
+        end
+      end
+
+      def extract_from_positions(positions)
+        positions = convert_to_offsets(positions)
+        begin_offset = positions.fetch("begin").fetch("offset")
+        end_offset = positions.fetch("end").fetch("offset")
+        length = end_offset - begin_offset
+
+        source[begin_offset, length + 1]
+      end
+
+      def convert_to_offsets(positions)
+        positions.each do |key, value|
+          next if value["offset"]
+
+          validate_position_format!(value)
+
+          positions[key] = {
+            "offset" => to_offset(value["line"] - 1, value["column"] - 1),
+          }
+        end
+
+        positions
+      end
+
+      def validate_position_format!(position)
+        unless position.key?("line") && position.key?("column")
+          message = "Location positions must have either line/column or offset form"
+
+          raise InvalidLocationPositions, message
+        end
+      end
+
+      def to_offset(line, column, offset = 0)
+        source.each_line.with_index do |source_line, index|
+          offset +=
+            if line == index
+              column
+            else
+              source_line.length
+            end
+
+          break if index >= line
+        end
+
+        offset
+      end
+    end
+  end
+end
--- a/spec/cc/analyzer/issue_spec.rb
+++ b/spec/cc/analyzer/issue_spec.rb
@@ -2,8 +2,8 @@ require "spec_helper"
 
 module CC::Analyzer
   describe Issue do
-    it "allows access to keys as methods" do
-      output = {
+    let(:output) do
+      {
         "categories" => ["Style"],
         "check_name" => "Rubocop/Style/Documentation",
         "description" => "Missing top-level class documentation comment.",
@@ -16,8 +16,11 @@ module CC::Analyzer
         },
         "remediation_points" => 10,
         "type" => "issue",
-      }.to_json
-      issue = Issue.new(output)
+      }
+    end
+
+    it "allows access to keys as methods" do
+      issue = Issue.new(output.to_json)
 
       expect(issue.respond_to?("check_name")).to eq true
       expect(issue.check_name).to eq("Rubocop/Style/Documentation")
@@ -25,65 +28,30 @@ module CC::Analyzer
 
     describe "#fingerprint" do
       it "adds a fingerprint when it is missing" do
-        output = {
-          "categories" => ["Style"],
-          "check_name" => "Rubocop/Style/Documentation",
-          "description" => "Missing top-level class documentation comment.",
-          "location"=> {
-            "lines" => {
-              "begin" => 32,
-              "end" => 40,
-            },
-            "path" => "lib/cc/analyzer/config.rb",
-          },
-          "remediation_points" => 10,
-          "type" => "issue",
-        }.to_json
-        issue = Issue.new(output)
+        issue = Issue.new(output.to_json)
 
-        expect(issue.fingerprint).to eq "9d20301efe0bbb8f87fb4eb15a71fc81"
+        expect(issue.fingerprint).to eq "433fae1189b03bcd9153dc8dce209fa5"
       end
 
       it "doesn't overwrite fingerprints within output" do
-        output = {
-          "categories" => ["Style"],
-          "check_name" => "Rubocop/Style/Documentation",
-          "description" => "Missing top-level class documentation comment.",
-          "fingerprint" => "foo",
-          "location"=> {
-            "lines" => {
-              "begin" => 32,
-              "end" => 40,
-            },
-            "path" => "lib/cc/analyzer/config.rb",
-          },
-          "remediation_points" => 10,
-          "type" => "issue",
-        }.to_json
-        issue = Issue.new(output)
+        output["fingerprint"] = "foo"
+
+        issue = Issue.new(output.to_json)
 
         expect(issue.fingerprint).to eq "foo"
+      end
+
+      it "uses the source fingerprint if env variable is present" do
+        issue = Issue.new(output.to_json)
+
+        expect(issue.fingerprint).to eq "433fae1189b03bcd9153dc8dce209fa5"
       end
     end
 
     describe "#as_json" do
       it "merges in defaulted attributes" do
-        output = {
-          "categories" => ["Style"],
-          "check_name" => "Rubocop/Style/Documentation",
-          "description" => "Missing top-level class documentation comment.",
-          "location"=> {
-            "lines" => {
-              "begin" => 32,
-              "end" => 40,
-            },
-            "path" => "lib/cc/analyzer/config.rb",
-          },
-          "remediation_points" => 10,
-          "type" => "issue",
-        }
         expected_additions = {
-          "fingerprint" => "9d20301efe0bbb8f87fb4eb15a71fc81",
+          "fingerprint" => "433fae1189b03bcd9153dc8dce209fa5",
         }
         issue = Issue.new(output.to_json)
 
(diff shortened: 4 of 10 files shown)


lib/cc/analyzer/issue.rb
```diff
@@ -39,20 +39,16 @@ def fingerprint
...
       def default_fingerprint
-        digest = Digest::MD5.new
-        digest << path
-        digest << "|"
-        digest << check_name.to_s
-        digest.to_s
+        if ENV["CODECLIMATE_SOURCE_FINGERPRINT"]
+          ::CC::CLI::SourceFingerprint.new(parsed_output).compute
+        else
+          ::CC::CLI::PathFingerprint.new(parsed_output).compute
```
pbrisbin: I think this is telling you these objects should live under the `Analyzer` namespace. Is there a reason you went with `CLI`?

lib/cc/cli/path_fingerprint.rb
```diff
@@ -0,0 +1,33 @@
...
+        @output = output
+      end
+
+      def compute
+        md5 = Digest::MD5.new
+        md5 << key
+        md5.hexdigest
+      end
+
+      def key
```
pbrisbin: Does this need to be public?

lib/cc/cli/source_fingerprint.rb
```diff
@@ -0,0 +1,69 @@
...
+            (lines["begin"]..lines["end"])
+          end
+        end
+      end
+
+      def extract_source
+        contents = ""
+
+        File.open(path) do |file|
+          file.each_line.with_index do |line, line_number|
```
pbrisbin: You may or may not enjoy this kind of conciseness, but I thought it was a neat trick nonetheless (to get rid of the `content` temp):

```rb
irb(main):001:0> file = "foo\nbar\nbaz\n"
irb(main):011:0> file.each_line.with_object("memo").with_index do |(l, m), ln|
irb(main):012:1* p l, m, ln
irb(main):013:1> end
"foo\n"
"memo"
0
"bar\n"
"memo"
1
"baz\n"
"memo"
2
=> "memo"
```

lib/cc/cli/source_fingerprint.rb
```diff
@@ -0,0 +1,69 @@
...
+        output["check_name"]
+      end
+
+      def path
+        output.fetch("location", {}).fetch("path", "")
+      end
+
+      def source
+        @source ||= begin
+          if line_range
```
wfleming: Any particular reason we're only supporting `lines`, and not `positions`?

lib/cc/analyzer/issue.rb
dblandin: Ahh, that was an oversight. I'll move it into Analyzer.

spec/cc/analyzer/issue_spec.rb
```diff
@@ -16,72 +16,44 @@ module CC::Analyzer
...
       end
+
+      it "uses the source fingerprint if env variable is present" do
+        ENV["CODECLIMATE_SOURCE_FINGERPRINT"] = "true"
+
+        issue = Issue.new(output.to_json)
+
+        expect(issue.fingerprint).to eq "ad79867c19494d6a1a567f804fc575ca"
+
+        ENV.delete("CODECLIMATE_SOURCE_FINGERPRINT")
```
pbrisbin: Should this be `ensure`d? Maybe behind a block-based `with_env` helper?

lib/cc/cli/path_fingerprint.rb
dblandin: It was nice to write assertions against the key in the spec. That's the only reason it's public.

lib/cc/cli/source_fingerprint.rb
dblandin: Nice, I like it.

lib/cc/cli/source_fingerprint.rb
dblandin: Nope, I can bake in position support as well.

spec/cc/analyzer/issue_spec.rb
dblandin: `ENV.delete` isn't an exception raising method, so I think it's fine as it. Doesn't seem worth it to extract as a helper method just yet.

```ruby
>> ENV.delete("CODECLIMATE_SOURCE_FINGERPRINT")
=> nil
```

lib/cc/cli/path_fingerprint.rb
```diff
@@ -0,0 +1,33 @@
...
+        "#{path}|#{check_name}"
```
GordonDiggs: Why use `|` separators instead of adding each item to the digest with `<<` separately?

lib/cc/cli/source_fingerprint.rb
wfleming: You might find the `SourceBuffer` class helpful.

lib/cc/cli/source_fingerprint.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      def compute
+        md5 = Digest::MD5.new
+        md5 << key
+        md5.hexdigest
+      end
+
+      def key
+        key = "#{path}|#{check_name}"
+        key << "|#{source.gsub(/\s+/, "")}" if source
```
GordonDiggs: This is where I'm concerned about using the `|` separators - it could lead to collisions here (maybe thats ok)

lib/cc/cli/path_fingerprint.rb
```diff
@@ -0,0 +1,33 @@
+require "digest/md5"
+
+module CC
```
wfleming: Nit: there is a lot of hash access going on in these two fingerprint classes. Would it be nicer in the long run to add more accessors to `Issue`?

spec/cc/analyzer/issue_spec.rb
pbrisbin: Yeah but if, for example `issue.fingerprint` raises, it won't be called. RSpec will hold the exception as a failure and run other specs with a sullied `ENV`.

dblandin: Ahh, I see what you mean. Yes, that makes sense. I'll make that change.

lib/cc/cli/path_fingerprint.rb
dblandin: Could you explain the benefit of adding the separators separately? I don't think they're even necessary, but it makes for a more readable key.

spec/cc/analyzer/issue_spec.rb
pbrisbin: I know I've written this somewhere:

```rb
with_env(k, v)
  old_v = ENV[k]
  ENV[k] = v
  yield
ensure
  ENV[k] = old_v
end
```

You could also just stub `ENV#[]` I suppose.

lib/cc/cli/source_fingerprint.rb
dblandin: Not sure I get what you mean about the collisions. Could you elaborate?

lib/cc/cli/path_fingerprint.rb
GordonDiggs: Hmm I thought it would do some work the help with collisions, but it doesn't. I think `|` is a little naive of a separator (especially since it's so common in source listings). WDYT of using `\0` instead?

lib/cc/cli/source_fingerprint.rb
GordonDiggs: See my above comment - I was wrong about collisions, but still have ideas about separators

lib/cc/cli/path_fingerprint.rb
dblandin: I think I'm leaning towards removing the separator entirely. I don't think it's very helpful. codeclimate-markdownlint doesn't [use one either](https://github.com/jpignata/codeclimate-markdownlint/blob/master/lib/cc/engine/markdownlint.rb#L93-L97).

pbrisbin: I'm still not clear on the concern here, can you give an example of two issues' source that would collide because of containing `|`?

lib/cc/cli/path_fingerprint.rb
pbrisbin: I consider it an anti-pattern to make methods public only for testing. Not a big deal though.

lib/cc/cli/path_fingerprint.rb
dblandin: Using accessors on the `Issue` sounds fine to me. The just pass the issue to the fingerprinting class?

```
SourceFingerprint.new(self).compute
```

lib/cc/cli/path_fingerprint.rb
dblandin: That's fair. Totally willing to move this to private scope.

lib/cc/cli/source_fingerprint.rb
```diff
@@ -0,0 +1,69 @@
...
+module CC
+  module CLI
+    class SourceFingerprint
+      def initialize(output)
+        @output = output
+      end
+
+      def compute
+        md5 = Digest::MD5.new
+        md5 << key
```
ABaldwinHunter: Nice! I like the work here and that you broke out the fingerprint computations into two separate functions, so that the old mechanism's still available. 

One question: Does this mean that anytime content changes the fingerprint will change? Could that create negative behavior changes in reporting new issues instead of `worse` issues - in fact perhaps not allowing issues to get worse in some cases? 

I suppose engines can implement their own fingerprints to account for some of those cases - like method length.

lib/cc/cli/path_fingerprint.rb
GordonDiggs: Hmm the more I think about it, I can't really come up with a good example. So, might be fine!

lib/cc/cli/source_fingerprint.rb
dblandin: That's a good question and I think that we're okay with the new behavior. If source code relevant to an existing issue is changed, we shouldn't consider the existing issue up-to-date. If a new analysis determines that the same kind of issue (identical check_name) is present for the updated source, code, we should emit a fresh issue with a new fingerprint. So in that case, you'd get a fixed issue and a new issue.

ABaldwinHunter: Ah okay. I didn't know that we'd considered and discussed the impact of that change.

One downside might be that more PRs will fail. And editing an existing issue - even improving it - will show up as a `new` issue.

But it's hard to have a perfect system so maybe this tradeoff is okay

dblandin: That's true. The new behavior will produce more issues, and potentially cause more failing status checks, but I think that's a favorable outcome. We want to reveal issues, not sweep them under the rug.

If an engine author wishes to override the default fingerprinting behavior, they can easily do so by providing that key in the issue, but I think it's a more useful default that what we have now.

ABaldwinHunter: 👍

dblandin:
@wfleming I pushed up a commit with support for location positions either in the line/column or offset form. Let me know what you think.

dblandin:
@codeclimate/review This is ready for another look! I addressed all of the comments brought up in the last round.

lib/cc/analyzer/source_extractor.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      def to_offset(line, column)
+        offset = 0
+
+        source.each_line.with_index do |source_line, index|
+          if line == index
+            offset += column
+          else
+            offset += source_line.length
+          end
```
pbrisbin: FWIW: I've started to come around to:

```rb
offset +=
  if line == index
    column
  else
    source_line.length
  end
```

Which appeases rubocop without misaligning the if/else/end depending on the size of the LHS of the assignment.

lib/cc/analyzer/source_extractor.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        source.each_line.with_object("").with_index do |(line, memo), index|
+          memo << line if range.include?(index)
+        end
+      end
+
+      def extract_from_positions(positions)
+        positions = convert_to_offsets(positions)
+
+        length = positions["end"]["offset"] - positions["begin"]["offset"]
```
pbrisbin: The nested hash access seems ripe for `nil`-errors. Are you confident these keys will be here? Would `Hash#fetch` be too syntactically noisy?

spec/cc/analyzer/issue_spec.rb
```diff
@@ -16,72 +16,44 @@ module CC::Analyzer
...
+
+    it "allows access to keys as methods" do
+      issue = Issue.new(output.to_json)
 
       expect(issue.respond_to?("check_name")).to eq true
       expect(issue.check_name).to eq("Rubocop/Style/Documentation")
     end
 
     describe "#fingerprint" do
+      include EnvHelpers
```
pbrisbin: I thought I converted this test suite to RSpec? I think we typically inject our helper modules into all specs via `RSpec.configure` immediately after the module definition, and only do this sort of thing when we're forced to by minitest.

lib/cc/analyzer.rb
```diff
@@ -29,8 +29,11 @@ module Analyzer
     autoload :LocationDescription, "cc/analyzer/location_description"
     autoload :LoggingContainerListener, "cc/analyzer/logging_container_listener"
     autoload :MountedPath, "cc/analyzer/mounted_path"
+    autoload :PathFingerprint, "cc/analyzer/path_fingerprint"
```
pbrisbin: Is this still relevant?

pbrisbin:
Some minor comments. Looks good enough to RC and move onto a builder branch IMO.

lib/cc/analyzer/source_extractor.rb
dblandin: I think the `#convert_to_offsets` method should ensure that these keys will be present but I'll take a look through this again.

spec/cc/analyzer/issue_spec.rb
dblandin: Looks like I missed this line in the spec_helper which loads all of the helpers:

```ruby
Dir.glob("spec/support/**/*.rb").each(&method(:load))
```

lib/cc/analyzer/source_extractor.rb
dblandin: Looks a little strange, but seems better to me! 👍

lib/cc/analyzer.rb
dblandin: Nope! Will remove.

PR #424, opened 2016-05-05
Update default fingerprinting

dblandin:
This PR introduces a few fixes based on testing the new default fingerprinting:

- Validate that issue line or position values are integers
- Lazily compute the default fingerprint
- Handle invalid UTF-8 bytes

### DIFF
--- a/lib/cc/analyzer/issue.rb
+++ b/lib/cc/analyzer/issue.rb
@@ -24,7 +24,7 @@ module CC
       end
 
       def fingerprint
-        parsed_output.fetch("fingerprint", default_fingerprint)
+        parsed_output.fetch("fingerprint") { default_fingerprint }
       end
 
       # Allow access to hash keys as methods
--- a/lib/cc/analyzer/issue_location_format_validation.rb
+++ b/lib/cc/analyzer/issue_location_format_validation.rb
@@ -30,13 +30,13 @@ module CC
       def valid_position?(position)
         position &&
           (
-            (position["line"] && position["column"]) ||
-            position["offset"]
+            [position["line"], position["column"]].all? { |value| value.is_a?(Integer) } ||
+            position["offset"].is_a?(Integer)
           )
       end
 
       def valid_lines?(lines)
-        lines.is_a?(Hash) && lines.key?("begin") && lines.key?("end")
+        lines.is_a?(Hash) && [lines["begin"], lines["end"]].all? { |value| value.is_a?(Integer) }
       end
     end
   end
--- a/spec/cc/analyzer/source_fingerprint_spec.rb
+++ b/spec/cc/analyzer/source_fingerprint_spec.rb
@@ -61,6 +61,23 @@ module CC::Analyzer
 
         expect(fingerprint.compute).to eq("eef541a28f83417a45808139d58b631d")
       end
+
+      it "clears out invalid UTF-8 bytes" do
+        output["location"] = {
+          "lines" => {
+            "begin" => 1,
+            "end" => 1
+          },
+          "path" => "spec/fixtures/stub.rb"
+        }
+
+        allow(File).to receive(:read).and_return("hi \255".force_encoding(Encoding::UTF_8))
+
+        issue = Issue.new(output.to_json)
+        fingerprint = SourceFingerprint.new(issue)
+
+        expect(fingerprint.compute).to eq("717ddedd698e51037903bad1d2b06c1b")
+      end
     end
   end
 end
--- a/lib/cc/analyzer/source_fingerprint.rb
+++ b/lib/cc/analyzer/source_fingerprint.rb
@@ -21,7 +21,10 @@ module CC
 
       def relevant_source
         source = SourceExtractor.new(raw_source).extract(issue.location)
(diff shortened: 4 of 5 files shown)


lib/cc/analyzer/issue_location_format_validation.rb
```diff
@@ -30,13 +30,13 @@ def valid_positions?(positions)
       def valid_position?(position)
         position &&
           (
-            (position["line"] && position["column"]) ||
-            position["offset"]
+            [position["line"], position["column"]].all? { |value| value.is_a?(Integer) } ||
+            position["offset"] && position["offset"].is_a?(Integer)
```
pbrisbin: Does this mean you're going to replace the errors we saw in production with engine validation errors?

If so, would it be better to cast?

spec/cc/analyzer/source_fingerprint_spec.rb
```diff
@@ -61,6 +61,23 @@ module CC::Analyzer
...
+      it "clears out invalid UTF-8 bytes" do
+        output["location"] = {
+          "lines" => {
+            "begin" => 1,
+            "end" => 1
+          },
+          "path" => "spec/fixtures/source_invalid_utf8.rb"
+        }
+
+        allow(File).to receive(:read).and_return("hi \255".force_encoding(Encoding::UTF_8))
```
pbrisbin: If you're stubbing read, why did you have to create the fixture file?

lib/cc/analyzer/issue.rb
```diff
@@ -24,7 +24,7 @@ def as_json(*)
       end
 
       def fingerprint
-        parsed_output.fetch("fingerprint", default_fingerprint)
+        parsed_output["fingerprint"] || default_fingerprint
```
pbrisbin: Did you actually see cases of `{ fingerprint: nil }` or `{ fingerprint: false }`?

dblandin: I went into more detail in the commit message. This is more about not going through the trouble of reading a file / computing the default fingerprint if we already have one in the issue output.

spec/cc/analyzer/source_fingerprint_spec.rb
dblandin: Ahh, that's right. The fixture file wasn't working so I switched tactics. I'll take out the fixture file.

lib/cc/analyzer/issue.rb
pbrisbin: Ah sorry. How about `parsed_output.fetch("fingerprint") { default_fingerprint }` (the block is only evaluated on missing keys)?

lib/cc/analyzer/issue_location_format_validation.rb
dblandin: I don't think it's better to cast in this case because any string could be cast to a `0`. You could have `"offset": "ohmygod"` and still get a `0` with `to_i`. 

The bugsnag error I saw actually pointed to a bug in the radon engine. I don't think we should have generated this issue at all and it's the only issue I saw with a string location value.

```json
{
    "categories": [
        "Bug Risk"
    ],
    "check_name": "Complexity",
    "content": {
        "body": "We encountered an error attempting to analyze this line."
    },
    "description": "Error: invalid syntax (<unknown>, line 102)",
    "fingerprint": "d2fec2141485c30d047b12f9f06283ee",
    "location": {
        "lines": {
            "begin": "102",
            "end": "102"
        },
        "path": "packages/grid_control_gui/xmpp/features.py"
    },
    "remediation_points": 1000000,
    "type": "issue"
}
```

lib/cc/analyzer/issue.rb
dblandin: I like it 👍

lib/cc/analyzer/issue_location_format_validation.rb
wfleming: I think our spec is pretty clear that these values should be numbers: I would not consider an engine sending us anything else as compliant, personally, and there's always bad edge cases with casting. I'd like to know which engines would be impacted by this change & action fixing them before shipping this, but IMHO this is a positive change.

pbrisbin: I see. If you're confident that this change won't cause a bunch of unexpected validation errors, then I'm totally good with the strict approach. My concern was that we had a lot of engines reporting string locations and the system has just been OK with it to this point.

dblandin: I think it's worth introducing and seeing what shakes out. We might want to fix the known Radon bug first as it's an engine we auto-enable.

dblandin: I just opened up a PR for Radon: https://github.com/rubik/radon/pull/89

GordonDiggs: Would it be enough to check `position["offset"].is_a?(Integer)`, since it would imply the presence check?

dblandin: Good call. I'm doing that elsewhere too. Will update.

PR #425, opened 2016-05-06
Support Docker for Mac

### LINKED ISSUE #421: Docker for Mac support
The beta version of Docker for Mac (using `docker` v1.11.x) runs natively on OS X and changes the locations of certain items expected by the `codeclimate-wrapper` script. For example, the Docker socket file at `/var/run/docker.sock` no longer exists on OS X (the container expects this to be passed in).

Will support be added to the wrapper script for Docker for Mac?

sgerrand:
💁  These changes support Docker's socket file existing in a [different location under Docker for Mac](https://beta.docker.com/docs/mac/docker-toolbox/#the-docker-for-mac-environment).

Closes #421

### DIFF
--- a/codeclimate-wrapper
+++ b/codeclimate-wrapper
@@ -18,15 +18,25 @@ EOF
   exit 1
 }
 
+docker_for_mac() {
+  test -S /var/tmp/docker.sock > /dev/null 2>&1
+}
+
 socket_missing() {
-  invalid_setup "/var/run/docker.sock must exist as a Unix domain socket"
+  invalid_setup "$socket_path must exist as a Unix domain socket"
 }
 
 invalid_docker_host() {
   local host=$1
 
-  invalid_setup "invalid DOCKER_HOST=$host, must be unset or unix:///var/run/docker.sock"
+  invalid_setup "invalid DOCKER_HOST=$host, must be unset or unix://$socket_path"
 }
+
+socket_path=/var/run/docker.sock
+
+if docker_for_mac; then
+  socket_path=/var/tmp/docker.sock
+fi
 
 if [ -n "$DOCKER_MACHINE_NAME" ] && command -v docker-machine > /dev/null 2>&1; then
   docker-machine ssh $DOCKER_MACHINE_NAME -- \
@@ -36,8 +46,8 @@ if [ -n "$DOCKER_MACHINE_NAME" ] && command -v docker-machine > /dev/null 2>&1; 
     'test -n "$DOCKER_HOST" -a "$DOCKER_HOST" != "unix:///var/run/docker.sock"' > /dev/null 2>&1 \
     && invalid_docker_host $(docker-machine ssh $DOCKER_MACHINE_NAME -- 'echo "$DOCKER_HOST"')
 else
-  test -S /var/run/docker.sock || socket_missing
-  test -n "$DOCKER_HOST" -a "$DOCKER_HOST" != "unix:///var/run/docker.sock" \
+  test -S $socket_path || socket_missing
+  test -n "$DOCKER_HOST" -a "$DOCKER_HOST" != "unix://$socket_path" \
     && invalid_docker_host "$DOCKER_HOST"
 fi
 
@@ -49,7 +59,7 @@ docker_run() {
     --env CODECLIMATE_DEBUG \
     --volume "$CODECLIMATE_CODE":/code \
     --volume "$CODECLIMATE_TMP":/tmp/cc \
-    --volume /var/run/docker.sock:/var/run/docker.sock \
+    --volume "$socket_path":/var/run/docker.sock \
     "$@"
 }
 


codeclimate-wrapper
```diff
@@ -18,6 +18,29 @@ EOF
   exit 1
 }
 
+modern_docker_version() {
+  docker --version | awk '{
```
pbrisbin: `| grep -q "Docker version 1.11"` is functionally equivalent to this `awk` script right?

sgerrand: Pretty much.

codeclimate-wrapper
```diff
@@ -28,35 +51,38 @@ invalid_docker_host() {
...
     && invalid_docker_host $(boot2docker ssh -- 'echo "$DOCKER_HOST"')
 else
-  test -S /var/run/docker.sock || socket_missing
+  socket_exists || socket_missing
   test -n "$DOCKER_HOST" -a "$DOCKER_HOST" != "unix:///var/run/docker.sock" \
     && invalid_docker_host "$DOCKER_HOST"
 fi
 
 docker_run() {
+  local socket_path=$(socket_path)
```
pbrisbin: I think there are some subtle bugs in the docker-machine case since this invocation of `socket_path` will run on the local system, not the docker-machine VM, so it won't find either socket.

What do you think about restructuring your change like this:

```sh
# Fix our bug, stop hardcoding a literal ;)
socket_path=/var/run/docker.sock

# Modify it for your case
#
# Note: the predicate has to be smarter than just "modern_docker_version". For example,
# I run 1.11 on Linux and the socket is in the default location
if docker_for_mac
  socket_path=/var/tmp/docker.sock
fi

# Now, the rest of the script is exactly as it was on master, except that anywhere
# we've hardcoded the socket path, use $socket_path.
```

sgerrand: That makes sense, I'll add something similar in.

sgerrand: Done in 7b4ad87 and 4ad8a6b. ♻️

sgerrand: > # Note: the predicate has to be smarter than just "modern_docker_version". For example,
> # I run 1.11 on Linux and the socket is in the default location

Done in 2ca5f7f. ♻️

codeclimate-wrapper
```diff
@@ -18,32 +18,52 @@ EOF
...
 
+docker_for_mac() {
+  test -S /var/tmp/docker.sock > /dev/null 2>&1
+}
+
+socket_exists() {
+  test -S $socket_path > /dev/null 2>&1
+}
+
+socket_path() {
```
pbrisbin: This approach still won't work for docker-machine (or boot2docker). The invocation at L47 will evaluate on the local machine and `exit 1` (which should be `return 1` anyway) since the socket is in the VM. That function is also incorrect anyway. This'll trip over into `socket_missing` when it shouldn't.

The invocations of `socket_exists` at L53 and L60 would also fail for similar reasons, but I don't think it'd get that far.

Something like this should work:

```sh
socket_path=/var/run/docker.sock

if [ -S /var/tmp/docker.sock ]; then
  socket_path=/var/tmp/docker.sock
fi
```

Leaving the rest of the script completely as-is, except for replacing the hardcoded socket with `$socket_path`.

This is very intentional in _not_ moving things out into named functions like `socket_path` or `socket_exists` because that'd change the semantics: currently, we're giving arguments to the `ssh` commands, your changes make them actual local invocations _to produce_ arguments for the `ssh` commands -- which won't work.

codeclimate-wrapper
```diff
@@ -56,7 +76,7 @@ docker_run() {
     --env CODECLIMATE_DEBUG \
     --volume "$CODECLIMATE_CODE":/code \
     --volume "$CODECLIMATE_TMP":/tmp/cc \
-    --volume /var/run/docker.sock:/var/run/docker.sock \
+    --volume $socket_path:/var/run/docker.sock \
```
pbrisbin: 💄 quotes: `"$socket_path":/var/run/docker.sock`

codeclimate-wrapper
pbrisbin: Also, sorry for the back and forth; we really appreciate you taking the time to work on this. Please let me know if you'd rather I commit what I'm thinking to this branch (or another) for your review instead.

codeclimate-wrapper
sgerrand: 🙇  Updated in 0e69153.

codeclimate-wrapper
sgerrand: @pbrisbin: Thanks for the tips, no worries about the back and forth.

> Leaving the rest of the script completely as-is, except for replacing the hardcoded socket with `$socket_path`.

To be clear: approaching this problem by changing the socket file location _does not work_ due to the differences between how `docker-machine` and Docker for Mac behave. There are _no_ Docker related environment variables set by Docker for Mac and the Docker Machine executables (`boot2docker` and `docker-machine`) may still exist on the user path, so the conditionals which testing for `command -v docker-machine` and `command -v boot2docker` need to be short circuited early.

sgerrand: 📝  The current version of Docker for Mac (v1.11.1-beta10) symlinks `/var/run/docker.sock` to `/var/tmp/docker.sock`, so the changes which are really needed in this script is:
* setting `$socket_path` to the right path; and
* shortcircuiting for Docker for Mac users before testing for `boot2docker`.

The test for `docker-machine` are guarded by `if [ -n "$DOCKER_MACHINE_NAME" ]` and the [Docker for Mac documentation](https://beta.docker.com/docs/mac/docker-toolbox/#setting-up-to-run-docker-for-mac) directs the user to unset every environment variable prefixed with `DOCKER_`.

pbrisbin: Ah good note.

What about this:

1. If the current version is symlinking to `/var/run/docker.sock` we should just use that and not bother an alternate `$socket_path`
1. If the documententation directs uses to unset `$DOCKER_`, we should rely on that and not worry about short-circuiting before checking for docker-machine
1. This only leaves the potential for a user having `boot2docker` on `$PATH`, but actually intending to use Docker Machine -- I think it's OK not to handle this edge-case

If those three things are true / appropriate -- do we need any changes at all?

codeclimate-wrapper
```diff
@@ -18,32 +18,52 @@ EOF
...
 }
 
-if [ -n "$DOCKER_MACHINE_NAME" ] && command -v docker-machine > /dev/null 2>&1; then
+socket_path=$(socket_path) || socket_missing
+
+if docker_for_mac; then
+  socket_exists || socket_missing
+elif [ -n "$DOCKER_MACHINE_NAME" ] && command -v docker-machine > /dev/null 2>&1; then
   docker-machine ssh $DOCKER_MACHINE_NAME -- \
-    test -S /var/run/docker.sock > /dev/null 2>&1 || socket_missing
```
sgerrand: > technically your changes are just passing a string literal of "socket_exists" to the ssh commands, which won't work at all.

You're completely correct, apologies for letting these two lines through.

codeclimate-wrapper
sgerrand: >If the current version is symlinking to `/var/run/docker.sock` we should just use that and not bother with an alternate `$socket_path`

The current script tests whether this file is a socket file, which fails as it's a symlink.

> If the documententation directs users to unset `$DOCKER_`, we should rely on that and not worry about short-circuiting before checking for docker-machine

Agreed. 👍 

> This only leaves the potential for a user having `boot2docker` on `$PATH`, but actually intending to use Docker Machine -- I think it's OK not to handle this edge-case since boot2docker is deprecated

That sounds like an argument for removing that test altogether, but I'm not really fussed either way.

sgerrand: 📝  Docker Toolbox for Mac does still install `boot2docker`. I'll check where it gets installed to.

pbrisbin: > At least one of them is false in my opinion. My view is that the socket path should be changed.

Makes sense. It also wouldn't work because we'd have to mount in both the symlink and the source for it to work.

Separately, I've gotten the green-light to remove the `boot2docker` code entirely -- which I'll now do.

Are we in agreement then that the few lines I've been pasting is the answer? Just adjust `$socket_path` up top and leave the rest of the script alone?

sgerrand: > Are we in agreement then that the few lines I've been pasting is the
answer? Just adjust $socket_path up top and leave the rest of the script
alone?

Yes, that's fine with me.

sgerrand: > Are we in agreement then that the few lines I've been pasting is the
> answer? Just adjust $socket_path up top and leave the rest of the script 
> alone?

Yes, that's fine with me.

sgerrand:
@pbrisbin: Thanks for all of your feedback through this process. Since the recent changes in Docker for Mac (as of beta10) and the removal of `boot2docker` (in c28015e7447f420dab0c0cbc0dcb38a7c068881b) I've found that `codeclimate-wrapper` now works fine with Docker for Mac. No changes required!

PR #429, opened 2016-05-06
Don't mutate issue location data during source extraction

dblandin:
This fixes a bug where the source extraction process would mutate the issue location data.


GordonDiggs:
LGTM

PR #433, opened 2016-05-09
Add Issue template

pbrisbin:
Not really a template, but a convenient place to show about-to-report users:

- Some common reasons for not reporting Issues
- Guidelines for reporting Issues

/cc @codeclimate/review @codeclimate/copy

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,64 @@
+Hi there!
+
+Sorry you're having trouble with our CLI, and thanks for taking the time to
+report the issue to us.
+
+There are a few reasons _not_ to open an issue here. Please consider the
+following:
+
+- Is this an issue with codeclimate.com?
+
+  If you're having trouble with our hosted analysis available on
+  codeclimate.com, please use our [Help & Support](https://codeclimate.com/help)
+  form.
+
+- Is this an issue with a specific engine?
+
+  Please report issues about a specific engine to that engine's repository. It
+  can usually be found at `codeclimate/codeclimate-<engine>`. If you can't find
+  the engine's repository (or Issues aren't enabled there), please do report
+  your issue here.
+
+- Is this an issue with your Docker installation in general?
+
+  While we try hard to smooth over some of this, there will always be
+  system-specific issues and edge-cases that we can't work around. If you're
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,66 @@
...
+  can usually be found at `codeclimate/codeclimate-<engine>`. If you can't find
+  the engine's repository (or Issues aren't enabled there), please do report
+  your issue here.
+
+- Is this an issue with your Docker installation in general?
+
+  While we try hard to smooth over some of this, there will always be
+  system-specific issues and edge-cases that we can't work around. If you're
+  unable to use Docker at all (i.e. commands like `docker version` or `docker
+  ps` fail), please try other channels (Stack Overflow, IRC, etc) before
```
GordonDiggs: WDYT of also suggesting `docker run hello-world`?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,66 @@
...
+  CODECLIMATE_DEBUG=1 codeclimate analyze
+  ```
+
+  If this output is large, consider using a service like gist.github.com.
+
+- Indicate what you expected to happen and what happened instead
+- Include your operating system, how you run Docker (e.g. Docker Machine) and
+  the version of Docker you're using.
+
+  Including the output of commands like
```
GordonDiggs: `Include the output of the following commands`?

GordonDiggs:
A few thoughts but LGTM

ISSUE_TEMPLATE.md
pbrisbin: That makes it sound more mandatory, which I didn't feel was necessary -- but would be fine with me, if you do.

ISSUE_TEMPLATE.md
pbrisbin: Should I put this in the parenthetical list with `version` and `ps`? Otherwise, I think i'd have to re-word more of the paragraph to include that and have it still flow nicely.

GordonDiggs: Yeah I think putting it in that list is a good idea

GordonDiggs:
👍

PR #435, opened 2016-05-11
Update validation to ensure project relative issue paths

dblandin:
This PR updates the `IssueRelativePathValidation` to ensure that issue paths
are relative to the code directory.

Both the issue path and code directory paths are expanded before the relative
check.

### DIFF
--- a/lib/cc/analyzer/issue_relative_path_validation.rb
+++ b/lib/cc/analyzer/issue_relative_path_validation.rb
@@ -10,7 +10,7 @@ module CC
       end
 
       def message
-        "Path must be relative"
+        "Path must be relative to the project directory"
       end
 
       private
--- a/lib/cc/analyzer/issue_relative_path_validation.rb
+++ b/lib/cc/analyzer/issue_relative_path_validation.rb
@@ -1,8 +1,12 @@
+require "pathname"
+
 module CC
   module Analyzer
     class IssueRelativePathValidation < Validation
       def valid?
-        path && !path.start_with?("/")
+        path &&
+          !path.start_with?("/") &&
+          relative_to?(MountedPath.code.container_path)
       end
 
       def message
@@ -10,6 +14,13 @@ module CC
       end
 
       private
+
+      def relative_to?(directory)
+        expanded_base = Pathname.new(directory).expand_path.to_s
+        expanded_path = Pathname.new(path).expand_path.to_s
+
+        expanded_path.start_with?(expanded_base)
+      end
 
       def path
         object.fetch("location", {})["path"]
--- a/spec/cc/analyzer/issue_relative_path_validation_spec.rb
+++ b/spec/cc/analyzer/issue_relative_path_validation_spec.rb
@@ -3,12 +3,26 @@ require "spec_helper"
 module CC::Analyzer
   describe IssueRelativePathValidation do
     describe "#valid?" do
-      it "returns true" do
-        expect(IssueRelativePathValidation.new("location" => { "path" => "foo.rb" })).to be_valid
+      it "returns true if path is relative to the project directory" do
+        expect(IssueRelativePathValidation.new("location" => {
+          "path" => "spec/fixtures/source.rb"
+        })).to be_valid
       end
 
-      it "returns false" do
-        expect(IssueRelativePathValidation.new("location" => { "path" => "/foo.rb" })).not_to be_valid
+      it "returns false if path is absolute" do
+        expect(IssueRelativePathValidation.new("location" => {
+          "path" => "#{MountedPath.code.container_path}/spec/fixtures/source.rb"
+        })).not_to be_valid
+      end
+
+      it "returns false if relative path moves up directories" do
+        expect(IssueRelativePathValidation.new("location" => {
+          "path" => "../../foo.rb"
+        })).not_to be_valid
+
+        expect(IssueRelativePathValidation.new("location" => {
+          "path" => "foo/../../../../bar"
+        })).not_to be_valid
       end
     end
   end
--- a/spec/cc/analyzer/issue_validator_spec.rb
+++ b/spec/cc/analyzer/issue_validator_spec.rb
@@ -27,7 +27,7 @@ module CC::Analyzer
         validator = IssueValidator.new({})
         expect(validator).not_to be_valid
         expect(validator.error).to eq(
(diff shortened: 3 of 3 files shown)


spec/cc/analyzer/issue_relative_path_validation_spec.rb
```diff
@@ -3,13 +3,17 @@
...
+        expect(IssueRelativePathValidation.new("location" => { "path" => "spec/fixtures/source.rb" })).to be_valid
       end
 
-      it "returns false" do
+      it "returns false if path is absolute" do
         expect(IssueRelativePathValidation.new("location" => { "path" => "/foo.rb" })).not_to be_valid
       end
+
+      it "returns false if relative path moves up directories" do
+        expect(IssueRelativePathValidation.new("location" => { "path" => "../../foo.rb" })).not_to be_valid
```
brynary: Can we also get a spec for: `foo/../../../../bar`?

lib/cc/analyzer/issue_relative_path_validation.rb
```diff
@@ -11,6 +11,13 @@ def message
 
       private
 
+      def relative_to?(directory)
```
pbrisbin: @brynary had a good suggestion to use `Pathname` for this, just to avoid messing up any of the fiddly bits in dealing with path names. I thought there was a `relative_to?`-like method on it already, but it doesn't look like it. There might be a way to use [`#relative_path_from`](http://ruby-doc.org/stdlib-2.2.0/libdoc/pathname/rdoc/Pathname.html#method-i-relative_path_from) though?

wfleming:
Should we update the message as well?  "Path must be relative" is...not a very helpful message if the path is `../foo.rb`.

lib/cc/analyzer/issue_relative_path_validation.rb
dblandin: I looked at `Pathname` and didn't find a useful method for this check. `Pathname#relative_path_from` looked promising but didn't end up offering a way to enforce that a path is a child of a base directory.

For example:

```ruby
>> base = Pathname.new(Dir.pwd)
>> Pathname.new("/tmp/example.txt").relative_path_from(base)
=> #<Pathname:../../../../tmp/example.txt>

Comment (wfleming) [inline lib/cc/analyzer/issue_relative_path_validation.rb]: `relative_path_from` doesn't behave very well in terms of trying to resolve a relative path in terms of an absolute path: it errors out. I don't think that's really what it's intended for anyway, since it doesn't actually touch the file system, it's just manipulating the path independent of the real FS.

`#cleanpath` may be useful: you could check that `path == Pathname.new(path).cleanpath.to_s`.

Comment (dblandin) [inline lib/cc/analyzer/issue_relative_path_validation.rb]: Updated the expand logic to use `Pathname#realpath` which has the same cleaning behavior as `cleanpath` but also follows symlinks.

File: spec/cc/analyzer/issue_relative_path_validation_spec.rb (hunk 1 above)
Comment (dblandin) [inline spec/cc/analyzer/issue_relative_path_validation_spec.rb]: Yep! Added.

**Comment by dblandin** [comment]
> Should we update the message as well? "Path must be relative" is...not a very helpful message if the path is ../foo.rb.

Good idea. I've updated the message to "Path must be relative to the project directory"

File: lib/cc/analyzer/issue_relative_path_validation.rb [hunk 3]
```diff
@@ -1,16 +1,28 @@
+require "pathname"
+
 module CC
   module Analyzer
     class IssueRelativePathValidation < Validation
       def valid?
-        path && !path.start_with?("/")
+        path &&
+          !path.start_with?("/") &&
+          File.exist?(path) &&
```
pbrisbin: This seems a little out of scope for this object.

dblandin: `Pathname#realpath` touches the filesystem and will error if the file doesn't exist.

```
Errno::ENOENT:
       No such file or directory @ realpath_rec - /private/tmp/source.rb
```

I could rescue that exception in `#relative_to?` instead. What do you think?

pbrisbin: Where did we land on dependent validations? Assuming this validation runs after the existence-validation, and we're OK with relying on order-dependcies, I'd say we do nothing here.

dblandin: Ah that's right. I forgot about that validation. I'll remove this `exist?` check.

lib/cc/analyzer/issue_relative_path_validation.rb
wfleming: Q: how does it handle dangling symlinks? (a symlink that points at a non-existent file.)

dblandin: I backed out of the symlink changes for this particular PR. It seems complex enough to deal with separately.

dblandin: I think that we'll have to update the `IssuePathExistenceValidation` to address symlinks and then we should be safe to use `Pathname#realpath` here.

pbrisbin: I'm interested in having an in-chat conversation about how we have decided / are deciding to "address symlinks". I think there's some nuance there and I know we've fought this battle before.

wfleming: > I'm interested in having an in-chat conversation about how we have decided / are deciding to "address symlinks". I think there's some nuance there and I know we've fought this battle before.

Definitely worth having a discussion (maybe in-person tomorrow at this point). I think our hand is being forced a bit now: we've fought parts of this battle, but also deferred the war (so to speak). With these stricter validations, and the behavior of content-based-fingerprinting, we now *must* "handle" symlinks in one way or another.

lib/cc/analyzer/issue_relative_path_validation.rb
```diff
@@ -1,16 +1,32 @@
...
 
+      def relative_to?(directory)
+        expanded_base = expand_path(directory)
+        expanded_path = expand_path(path)
+
+        expanded_path.start_with?(expanded_base)
+      end
+
+      def expand_path(path)
+        cleaned = Pathname.new(path).cleanpath.to_s
```
wfleming: Any reason not to use `Pathname#expand_path`? I think it pretty much would basically replace this whole method from what i've tested of its behavior.

dblandin: Works for me! Updated.

wfleming:
LGTM

PR #438, opened 2016-05-20
Validate that issue-path is a file

pbrisbin:
Emitting issues for directories is not something we currently support. Doing so
was not causing visible problems until we added source-based fingerprinting
which attempts to read the path and errors on directories:

https://bugsnag.com/code-climate/builder/errors/5736508dbe3f29e6a6c1cde8?event_id=573e507dba76e62611aa1445

Note: this still returns true for symlinks; discussion for how to handle those
cases is still open and this patch should not impact that behavior either way.

/cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/issue_path_is_file_validation.rb
+++ b/lib/cc/analyzer/issue_path_is_file_validation.rb
@@ -1,0 +1,19 @@
+module CC
+  module Analyzer
+    class IssuePathIsFileValidation < Validation
+      def valid?
+        File.file?(path)
+      end
+
+      def message
+        "Path is not a file: '#{path}'"
+      end
+
+      private
+
+      def path
+        object.fetch("location", {}).fetch("path", "")
+      end
+    end
+  end
+end
--- a/lib/cc/analyzer/issue_validator.rb
+++ b/lib/cc/analyzer/issue_validator.rb
@@ -8,6 +8,7 @@ module CC
         IssueLocationFormatValidation,
         IssueOtherLocationsFormatValidation,
         IssuePathExistenceValidation,
+        IssuePathIsFileValidation,
         IssuePathPresenceValidation,
         IssueRelativePathValidation,
         IssueTypeValidation,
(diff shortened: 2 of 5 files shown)


lib/cc/analyzer/issue_validator.rb
```diff
@@ -8,6 +8,7 @@ class IssueValidator
         IssueLocationFormatValidation,
         IssueOtherLocationsFormatValidation,
         IssuePathExistenceValidation,
+        IssuePathIsFileValidation,
```
wfleming: Should probably validate this after presence & relative path validations?

GordonDiggs: Like latin, order does not matter here. All the validations get run every time

lib/cc/analyzer/issue_path_is_file_validation.rb
```diff
@@ -0,0 +1,19 @@
+module CC
+  module Analyzer
+    class IssuePathIsFileValidation < Validation
+      def valid?
+        File.file?(path)
```
wfleming: This is `false` for symlinks. That's a kind of significant product change that should probably be discussed.

GordonDiggs: I believe this will also return false for symlinks. I think that's fine but wanted to be sure that's ok with you

pbrisbin: No one reads my commit messages ;)

lib/cc/analyzer/issue_validator.rb
wfleming: > Like latin, order does not matter here.

👏

lib/cc/analyzer/issue_path_is_file_validation.rb
wfleming: "Trust but verify".

GordonDiggs:
LGTM

ABaldwinHunter:
I hesitate over this change, because I believe it's moving in the opposite direction of what we want: a way to handle `file`less issues gracefully - but I'm okay with it for the time being since I agree that:
> Emitting issues for directories is not something we currently support.

GordonDiggs:
@ABaldwinHunter definitely a valid concern. I think that there's going to be a large amount of change to be made whenever we decide to support that, though, and a lot of it will be in this area. I agree it does move in that direction, but not a huge amount

pbrisbin:
@ABaldwinHunter it's a good point. IMO this is just trading an unclear error and a U10 for a clear one and a not-U10 for the time being. Agreed with Gordon that there will be much larger, sweeping changes to actually support issues on non-existent files and reverting this bit would make up a very small percentage of such a change.

ABaldwinHunter:
Makes sense. 👍

PR #440, opened 2016-05-20
Add HTML formatter

mrb:
This is going to be very helpful for engine developers, primarily because it gives them a chance to see HTML rendered output versions of their description and content body.

screenshot 2016-05-20 18 02 19

### DIFF
--- a/lib/cc/analyzer/formatters.rb
+++ b/lib/cc/analyzer/formatters.rb
@@ -2,11 +2,13 @@ module CC
   module Analyzer
     module Formatters
       autoload :Formatter, "cc/analyzer/formatters/formatter"
+      autoload :HTMLFormatter, "cc/analyzer/formatters/html_formatter"
       autoload :JSONFormatter, "cc/analyzer/formatters/json_formatter"
       autoload :PlainTextFormatter, "cc/analyzer/formatters/plain_text_formatter"
       autoload :Spinner, "cc/analyzer/formatters/spinner"
 
       FORMATTERS = {
+        html: HTMLFormatter,
         json: JSONFormatter,
         text: PlainTextFormatter,
       }.freeze
--- a/lib/cc/analyzer/formatters/html_formatter.rb
+++ b/lib/cc/analyzer/formatters/html_formatter.rb
@@ -1,0 +1,82 @@
+require "redcarpet"
+require "active_support/number_helper"
+
+module CC
+  module Analyzer
+    module Formatters
+      class HTMLFormatter < Formatter
+        class ReportTemplate
+          include ERB::Util
+          attr_accessor :template, :issues, :issues_by_path
+
+          TEMPLATE_PATH = File.expand_path(File.join(File.dirname(__FILE__), "templates/html.erb"))
+
+          def initialize(issue_count, issues_by_path)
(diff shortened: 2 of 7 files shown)


lib/cc/analyzer/formatters/html_formatter.rb
```diff
@@ -0,0 +1,104 @@
...
+            issues << json
+          when "warning"
+            warnings << json
+          else
+            raise "Invalid type found: #{json["type"]}"
+          end
+        end
+
+        def finished
+          puts "<h3>#{pluralize(issues.size, "issue")} found in #{issues_by_path.size} files.</h3>"
```
GordonDiggs: All these `puts` make me sad. How hard would it be to pull something like ERB into this to handle these?

mrb: @gordondiggs probably not too hard! I'll give it a shot.

mrb:
@GordonDiggs Ready for re-review

lib/cc/analyzer/formatters.rb
```diff
@@ -4,11 +4,13 @@ module Formatters
       autoload :Formatter, "cc/analyzer/formatters/formatter"
       autoload :JSONFormatter, "cc/analyzer/formatters/json_formatter"
       autoload :PlainTextFormatter, "cc/analyzer/formatters/plain_text_formatter"
+      autoload :HTMLFormatter, "cc/analyzer/formatters/html_formatter"
       autoload :Spinner, "cc/analyzer/formatters/spinner"
 
       FORMATTERS = {
         json: JSONFormatter,
         text: PlainTextFormatter,
+        html: HTMLFormatter,
```
wfleming: 💄 this list should probably be sorted, as should the autoload list above.

spec/cc/analyzer/formatters/html_formatter_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'spec_helper'
+
+module CC::Analyzer::Formatters
+  describe HTMLFormatter do
+    include Factory
+
+    let(:formatter) do
+      filesystem ||= CC::Analyzer::Filesystem.new(
```
wfleming: FWIW the return value of `let`s are memoized by rspec, so I don't think the `||` is doing anything for you here?

lib/cc/analyzer/formatters/templates/html.erb
```diff
@@ -0,0 +1,37 @@
...
+  </head>
+  <body>
+  <div class="container">
+    <h2>Code Climate Analysis</h2>
+
+    <h3><%= pluralize(@issue_count, "issue") %> found in <%= @issues_by_path.size %> files.</h3>
+
+    <% @issues_by_path.each do |path, file_issues| %>
+      <hr>
+      <b><a name='<%= path %>'><%= path %></a> (<%= pluralize(file_issues.size, "issue") %>)</b>
```
GordonDiggs: Maybe use an `h4` here?

GordonDiggs:
One minor thing and I agree with Will's comment but otherwise LGTM! This will be awesome to have

wfleming:
Definitely really helpful for community authors!

PR #441, opened 2016-05-25
Support engine image channels

pbrisbin:
- Rewrite registry to replace image:x with channels[stable]:x
- Update engines-config-builder to look for channels.stable
- Improve error messaging for ImageRequired errors to indicate the cause may be
  an invalid channel
- Support -e engine:channel on the command line

Individual commits have more details.

This is a fully backwards compatible change. Builder should pick up the new
logic in engines-config-builder, but it'll need to rewrite its own manifest to
support channels as well as add its own error handling for ImageRequired.

Here in CLI, we specify that tag-less images on Docker Hub are for the stable
channel. When we have non-stable channel images, those will likely be tagged as
the channel on Docker Hub. In Builder, the channel keys will always specify
"fully qualified" (i.e. :b123) images. Changes to our tooling will happen
elsewhere later to ensure the correct images are in the correct places as part
of releasing an engine with multiple channels.

/cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -3,6 +3,19 @@ require "securerandom"
 module CC
   module Analyzer
     class EnginesConfigBuilder
+      class RegistryAdapter < SimpleDelegator
+        # Calling this is guarded by Registry#key?(name) so we can assume
+        # metadata itself will be present. We own the YAML loaded into the
+        # registry, so we can also assume the "channels" key will be present. We
+        # can't assume it will have a key for the given channel, but the nil
+        # value for the returned image key will trigger the desired error
+        # handling.
+        def fetch(name, channel)
+          metadata = self[name]
+          metadata.merge("image" => metadata["channels"][channel])
+        end
+      end
+
       Result = Struct.new(
         :name,
         :registry_entry,
@@ -12,7 +25,7 @@ module CC
       )
 
       def initialize(registry:, config:, container_label:, source_dir:, requested_paths:)
-        @registry = registry
+        @registry = RegistryAdapter.new(registry)
         @config = config
         @container_label = container_label
         @requested_paths = requested_paths
@@ -23,7 +36,8 @@ module CC
         names_and_raw_engine_configs.map do |name, raw_engine_config|
           label = @container_label || SecureRandom.uuid
           engine_config = engine_config(raw_engine_config)
-          Result.new(name, @registry[name], @source_dir, engine_config, label)
+          engine_metadata = @registry.fetch(name, raw_engine_config.channel)
+          Result.new(name, engine_metadata, @source_dir, engine_config, label)
         end
       end
 
--- a/lib/cc/cli/help.rb
+++ b/lib/cc/cli/help.rb
@@ -14,7 +14,7 @@ module CC
 
       def commands
         [
-          "analyze [-f format] [-e engine] <path>",
+          "analyze [-f format] [-e engine(:channel)] <path>",
           "console",
           "engines:disable #{underline("engine_name")}",
           "engines:enable #{underline("engine_name")}",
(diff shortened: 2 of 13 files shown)


pbrisbin:
I snuck in a small commit to exit 1 on errors, which I noticed we weren't doing while testing the `ImageRequired` error handling.

lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -3,6 +3,19 @@
 module CC
   module Analyzer
     class EnginesConfigBuilder
+      class RegistryAdapter < SimpleDelegator
```
pbrisbin: This logic could've been done as just a private method here, but I thought the adapter class was nice because it hints that, design-wise, this is the interface `Registry` _should_ have, but I've elected not to go down that route because the current CLI/Builder split of responsibilities makes that difficult and investing the time into addressing the difficulty doesn't make sense when we hope to move to something entirely different (the sub-command approach) eventually.

lib/cc/analyzer/engines_config_builder.rb
```diff
@@ -3,6 +3,19 @@
...
+        # Calling this is guarded by Registry#key?(name) so we can assume
+        # metadata itself will be present. We own the YAML loaded into the
+        # registry, so we can also assume the "channels" key will be present. We
+        # can't assume it will have a key for the given channel, but the nil
+        # value for the returned image key will trigger the desired error
+        # handling.
+        def fetch(name, channel)
+          metadata = self[name]
```
GordonDiggs: I don't see a big win in this assignment over doing `self[name].merge(...)`. WDYT?

lib/cc/cli/help.rb
```diff
@@ -14,7 +14,7 @@ def run
 
       def commands
         [
-          "analyze [-f format] [-e engine] <path>",
+          "analyze [-f format] [-e engine(:channel)] <path>",
```
GordonDiggs: This is awesome - we should highlight this when documenting upgrading to the new eslint

lib/cc/cli/runner.rb
```diff
@@ -8,8 +8,8 @@ def self.run(argv)
         new(argv).run
       rescue => ex
         $stderr.puts("error: (#{ex.class}) #{ex.message}")
-
         CLI.debug("backtrace: #{ex.backtrace.join("\n\t")}")
+        exit 1
```
GordonDiggs: Drive by change?

GordonDiggs: EDIT: looks like it's in its own commit, so I'm fine with it (unless it causes enough issues to be worth pulling for now)

GordonDiggs:
Some tiny nits. This looks awesome

lib/cc/analyzer/engines_config_builder.rb
pbrisbin: I don't see any wins either way. If you prefer dropping the temp, that's fine with me. The name comes in handy in the comment that describes the assumptions so I'd have to reword that too.

GordonDiggs: I don't feel that strongly - fine to leave it as is

lib/cc/cli/runner.rb
pbrisbin: I commented when I added this: it was motivated by the work here, but is its own thing, so I figured same PR separate commit was appropriate. It seems to be (somehow) triggering these JUnit reporter errors though so I'm going to pull it out. Baffling.

pbrisbin:
Forgot to make `engines:install` work. Stand by.

pbrisbin:
Made the updates for `engines:install` and `test`. Could use another quick look (/cc @GordonDiggs). I noticed that `engines:list` might benefit from including channel info, but the formatting was non-obvious so I'm punting for now.

pbrisbin:
Ping @codeclimate/review for a last look please.

GordonDiggs:
👍

PR #446, opened 2016-05-26
Fix type error in channel lookup

pbrisbin:
A channel read via CC::Yaml will be a Yaml::Nodes::Scalar, not a String, so
(even though printing things looks like it'd all work), looking up into the
registry Hash always fails: Scalar("stable") != "stable".

When there's no channel in the config, CC::Yaml defaults to the String "stable"
(not a Scalar representing "stable"), so the behavior for a defaulted channel
was tested and worked, and the behavior for an invalid channel was tested and
worked. However, the behavior for a non-defaulted, and valid channel doesn't
work without this fix.

/cc @codeclimate/review

### DIFF
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -12,7 +12,7 @@ module CC
         # handling.
         def fetch(name, channel)
           metadata = self[name]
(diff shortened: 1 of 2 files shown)


wfleming:
👍

ABaldwinHunter:
nice

Issue #92, opened 2015-08-14
Add Engine for SCSS-Lint?


elwayman02:
Are there any plans to add scss-lint to the engines list? csslint isn't really all that helpful as an engine in the days of LESS & SASS...it just throws a bunch of useless errors for most projects.

GordonDiggs:
@elwayman02 We do have plans to add it! It's not quite ready yet but I will update this issue when we release it

elwayman02:
Thanks @GordonDiggs!

dylangrafmyre:
:+1: 
The maintainer of scss-lint also has slim-lint and haml-lint. Maybe you can add 3 for the price of one :)

GordonDiggs:
@elwayman02 @dylangrafmyre 

SCSS Lint is now available as of version 0.6.0! To enable it, you can run `codeclimate engines:enable scss-lint`

elwayman02:
Awesome, glad to hear it!

dylangrafmyre:
Great! Thank You!

aliceinwire:
:+1

PR #448, opened 2016-05-26
Add eslint-2 channel for eslint engine

pbrisbin:
/cc @codeclimate/review

I locally built and pushed an image from codeclimate/codeclimate-eslint#79 so
it's now available to try. Consider this a quiet release for wider testing,
it's entirely opt-in.

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -84,6 +84,7 @@ duplication:
 eslint:
   channels:
     stable: codeclimate/codeclimate-eslint
+    eslint-2: codeclimate/codeclimate-eslint:eslint-2
   description: A JavaScript/JSX linting utility.
   community: false
   upgrade_languages:


abritinthebay:
Does this mean I can opt-in on a project *right now* if I edit my yml file?

pbrisbin:
@abritinthebay yes, that's correct. Note that we're still testing this image out so I can't say for sure that it'll work, but you're welcome to try it:

Once you've updated to the latest CLI, you can add:

```yaml
eslint:
  enabled: true
  channel: eslint-2
```

abritinthebay:
The latest CLI? Will it not also work on the web dashboard..?

pbrisbin:
Oh yes sorry. Same caveat about it not being thoroughly tested yet, but you can commit that yaml change and it will run eslint on our hosted analysis too.

abritinthebay:
hmmm I'm getting "We had trouble running the unknown engine."

Here's the engines section of my yml:

```
engines:
  csslint:
    enabled: false
  duplication:
    enabled: false
  eslint:
    enabled: true
    channel: eslint-2
    config: .eslintrc.js
  fixme:
    enabled: true
  scss-lint:
    enabled: true
```

pbrisbin:
Sorry, you were just a few minutes too soon as I was in the process of pushing that feature out for hosted analysis. It should work if you try again now.

abritinthebay:
haha, what's more new that _bleeding edge_ ;) 

So right now it fails with `Error: Cannot find module 'estraverse-fb'

Which appears to be a babel-eslint dependency issue...?

pbrisbin:
Damn. Unfortunately, I'm not immediately sure what the issue is there. I'll open up an issue on the engine repo to track it, but it probably won't get addressed today. Thank you very much for trying it out, and apologies that it's not working on your project yet.

abritinthebay:
Worked it out but only you can fix it. For ESlint 2.x you need babel-eslint@6.x.x not the 5.x branch that works with regular ESlint.

abritinthebay:
see here: https://github.com/babel/babel-eslint/issues/267

abritinthebay:
I'm _very_ willing to alpha test this and find bugs - it's very important for my team here at Bleacher Report to get this in our toolchain on CC :)

pbrisbin:
Thanks! I've rebuilt a new image from #92 with that fix and pushed it to Docker Hub, this means you could test the fix via the CLI right now. If you can only use hosted, you'll have wait just a bit longer as I go through the process of getting the image out there as well.

abritinthebay:
I could use CLI locally but I'm totally willing to wait - don't have it fully set up to use the CLI atm.

pbrisbin:
OK, that fix is now out. Feel free to try again.

abritinthebay:
It works! 

I'm getting a lot of `Definition for rule 'babel/some-rule-name' was not found` which I guess means it's not loading the babel eslint plugin, but everything else seems to work.

This is great :) if that odd babel plugin issue starts working it'll be *perfect* !

abritinthebay:
Specifically it seems to complain about `babel/no-await-in-loop`... but not the other babel rules we have in there... hmmm.

abritinthebay:
for now I've just set that to not be checked in the `checks` section of the config.

pbrisbin:
That's great to hear!

Tomorrow we'll be writing up documentation and dropping comments on all the OSS Issues where this support has been requested. Then we'll start collecting and addressing things like that. Thanks again for being an early tester (and finding that babel-eslint fix).

PR #450, opened 2016-05-27
Support engine channels in dev mode

dblandin:
`--dev` mode broke following our introduction of engine channels. This PR updates the registry to produce a structure compatible with the new `RegistryAdapter` when in dev mode.


wfleming:
👍

PR #452, opened 2016-05-27
Use engine name + channel in logs, container name, and metrics

dblandin:
This allows us to monitor and track engine channels separately.


wfleming:
LGTM modulo CC issues.

wfleming:
Still  LGTM for an RC

PR #454, opened 2016-06-02
Don't error creating the fingerprint when the file doesn't exist

GordonDiggs:
This is sometimes run before issue validations, so it should correctly
handle a missing file

@codeclimate/review

### DIFF
--- a/lib/cc/analyzer/source_fingerprint.rb
+++ b/lib/cc/analyzer/source_fingerprint.rb
@@ -28,7 +28,12 @@ module CC
       end
 
       def raw_source
-        @raw_source ||= File.read(issue.path)
+        @raw_source ||=
+          if File.file?(issue.path)
+            File.read(issue.path)
+          else
+            ""
+          end
       end
     end
   end
--- a/spec/cc/analyzer/source_fingerprint_spec.rb
+++ b/spec/cc/analyzer/source_fingerprint_spec.rb
@@ -71,12 +71,28 @@ module CC::Analyzer
           "path" => "spec/fixtures/stub.rb"
         }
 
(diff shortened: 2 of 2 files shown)


lib/cc/analyzer/source_fingerprint.rb
```diff
@@ -28,7 +28,12 @@ def relevant_source
       end
 
       def raw_source
-        @raw_source ||= File.read(issue.path)
+        @raw_source ||=
+          if File.exist?(issue.path)
```
pbrisbin: Thoughts on `File.file?` which is "exists and is not a directory". The exceptions that prompted some other directory-related work were from such a case (but didn't fix it because of this unknown order situation).

GordonDiggs: Good idea - done!

pbrisbin:
LGTM

dblandin:
@GordonDiggs Do you know when a fingerprint would be generated for an issue before the issue validations (including `IssuePathIsFileValidation`) run?

Based on [this snippet](https://github.com/codeclimate/codeclimate/blob/8169da9a9c9698c64a53a9a1d8fe007fdf3a65b6/lib/cc/analyzer/engine.rb#L40-L47) it seems like we may instead be outputting the issue json, and creating a source fingerprint, even if the output is invalid:

```ruby
unless output.valid?
  stdout_io.failed("#{qualified_name} produced invalid output: #{output.error[:message]}")
  container.stop
end

  stdout_io.write(output.to_json) || container.stop
end
```

Would it make sense to fix that instead of duplicating the `File.file?` check?

PR #457, opened 2016-06-06
Allow negated exclude patterns

wfleming:
This allows negated exclude patterns, [similar to git's behavior](https://git-scm.com/docs/gitignore#_pattern_format).


GordonDiggs:
Implementation LGTM if we decide we want this

wfleming:
Credit for the implementation here goes to @maxjacobson. I just cleaned it up a bit & added a couple other tests for sanity.

maxjacobson:
👋 oh hey. Thanks for cleaning it up 😊

PR #461, opened 2016-06-16
Pass engine parameter ENV vars through wrapper

GordonDiggs:
These can be useful for debugging / authoring, so pass them through

@codeclimate/review


wfleming:
👍

PR #464, opened 2016-06-22
Refactor validations into their own module

wfleming:
I was about to add a new validation, but decided to do this initially as
cleanup: we have enough validations that it feels appropriate to put
them in their own namespace.

cc @codeclimate/review


lib/cc/analyzer/issue_validations.rb
```diff
@@ -0,0 +1,25 @@
...
+      autoload :PathIsFileValidation, "cc/analyzer/issue_validations/path_is_file_validation"
+      autoload :PathPresenceValidation, "cc/analyzer/issue_validations/path_presence_validation"
+      autoload :RelativePathValidation, "cc/analyzer/issue_validations/relative_path_validation"
+      autoload :TypeValidation, "cc/analyzer/issue_validations/type_validation"
+      autoload :Validation, "cc/analyzer/issue_validations/validation"
+
+      def self.validations
+        constants.sort.map do |const_name|
+          const_get(const_name)
+        end.select do |klass|
```
wfleming: I disagree with Rubocop on this because I prefer functional transform chaining to temporary vars. Other people's opinion?

pbrisbin: I do try and avoid it because I think it looks a bit off. In this case, I'd be open to `constants.sort.map(&method(:const_get)).select do |klass|`. Just an idea, I'd also be find disabling the rule if that's how everyone feels.

wfleming: I totally forgot about the `&method` trick in this case. That'll probably work here for now.

maxjacobson:
LGTM

PR #465, opened 2016-06-22
Validate Issue Content key

wfleming:
We occassionally see invalid contents emited that will blow things up
further down the line (like a `null` body, for example). This adds a
validation to catch these things earlier.

cc @codeclimate/review

### DIFF
--- a/spec/cc/analyzer/issue_validations/content_validation_spec.rb
+++ b/spec/cc/analyzer/issue_validations/content_validation_spec.rb
@@ -1,0 +1,31 @@
+require "spec_helper"
+
+module CC::Analyzer::IssueValidations
+  describe ContentValidation do
+    describe "#valid?" do
+      it "is valid without any content" do
+        expect(ContentValidation.new({})).to be_valid
+      end
+
+      it "is valid with content with body" do
+        expect(ContentValidation.new("content" => { "body" => "hi"})).to be_valid
+      end
+
+      it "is not valid with content as string" do
+        expect(ContentValidation.new("content" => "hi")).not_to be_valid
+      end
+
+      it "is not valid with content as nil" do
+        expect(ContentValidation.new("content" => nil)).not_to be_valid
+      end
+
+      it "is not valid with content without body" do
+        expect(ContentValidation.new("content" => {})).not_to be_valid
+      end
+
+      it "is not valid with content with nil body" do
+        expect(ContentValidation.new("content" => { "body" => nil })).not_to be_valid
+      end
+    end
+  end
+end
--- a/lib/cc/analyzer/issue_validations.rb
+++ b/lib/cc/analyzer/issue_validations.rb
@@ -3,6 +3,7 @@ module CC
     module IssueValidations
       autoload :CategoryValidation, "cc/analyzer/issue_validations/category_validation"
       autoload :CheckNamePresenceValidation, "cc/analyzer/issue_validations/check_name_presence_validation"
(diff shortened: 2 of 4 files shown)


spec/cc/analyzer/issue_validations/content_validation_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      it "is valid with content with body" do
+        expect(ContentValidation.new("content" => { "body" => "hi"})).to be_valid
+      end
+
+      it "is not valid with content as string" do
+        expect(ContentValidation.new("content" => "hi")).not_to be_valid
+      end
+
+      it "is not valid with content as nil" do
+        expect(ContentValidation.new("content" => nil)).not_to be_valid
```
GordonDiggs: Does this case cause issues?

wfleming: Honestly not sure. It doesn't comply with the spec, though, so may as well catch it?

wfleming: I guess there's a valid question of "if this *doesn't* cause issues, how common is it currently in the wild?"

GordonDiggs: Yeah I'm cool with leaving it, but we should keep an eye on the error codes

GordonDiggs:
LGTM

spec/cc/analyzer/issue_validations/content_validation_spec.rb
pbrisbin: I think it'd be worth doing a quick query (on the secondary of course) for issues with `content: nil`. I know the extension always checks for presence before displaying it, so I thought it was valid to omit. I'm fairly sure not everything has ReadUp content and I'd hate to start throwing validation errors.

wfleming: It is valid to omit. It is not valid to include with a `null` value. Minor semantic difference. I'll check secondary tomorrow.

pbrisbin: Ohhh. OK, you can ignore me then, or go ahead with investigation. Up to you.

wfleming: So `db.smells.count({ content: { $exists: true, $in: [null] } })` is not going to run in a reasonable amount of time (my connection got broken by fab after an hour or so). Can't do `mapReduce` on slave (apparently), so I don't know of a good way to "background" the job. Any other ideas about how to query for this?

pbrisbin: Protip: Add an `_id: { $gt: BSON::ObjectId.from_time(something_reasonable) }`. We don't need to check if there's _ever_ been such issues, but knowing if any have been emitted recently is valuable.

wfleming: Got back to this today. For future reference, it seemed like I had to provide an index hint to `count` to get a response in reasonable time.

```
db.runCommand({ 
   count: 'smells',
   query: { _id: { $gt: ObjectId('577144910000000000000000') }, content: { $exists: true, $in: [null] } },
   hint: { _id: 1 }
})

Anyway: no issues in the last day that matched this query, which seems like a large enough data set to me. (That's ~88.7M issues.)
```

pbrisbin: :+1:

PR #472, opened 2016-07-20
Rescue POSIX::Spawn::Timeout when killing containers

pbrisbin:
When we kill containers due to resource limitations (usually time), we give the
docker-kill and docker-wait commands two minutes to complete. If they don't
complete in time, this exception occurs, halts the general shutdown process,
and ultimately bubbles up as an opaque exception (or U10 error) to the user.

It's better to rescue this experience and let the shutdown process continue as
it was, which will result in a more meaningful exception (or E-class error) to
the user.

A "zombie" container will still be left running, but by recording and viewing
the added metrics, we can identify a repeat offender (e.g. a particularly
un-killable engine) and address that problem at its root.

/cc @codeclimate/review


GordonDiggs:
LGTM. FWIW we've seen this behavior on an enterprise instance as well

maxjacobson:
👍 I like the logging strategy

PR #478, opened 2016-08-16
Put waitpid2 in a thread, so we can kill it on #stop

pbrisbin:
If an engine container runs away and our docker-kill/wait does not succeed,
this waitpid2 can block forever. To address this we put it in an
immediately-joined thread. We kill this thread as part of #stop. If the child
had exited, the kill will be a no-op. Otherwise, it will allow builder to move
on without waiting forever.

/cc @codeclimate/review


GordonDiggs:
LGTM

PR #483, opened 2016-08-22
Add Time Arg

ABaldwinHunter:
@codeclimate/review

### DIFF
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -40,6 +40,8 @@ module CC
             @engine_options << @args.shift
           when "--dev"
             @dev_mode = true
+          when "--timeout"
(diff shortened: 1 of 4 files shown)


lib/cc/cli/analyze.rb
```diff
@@ -40,6 +40,8 @@ def process_args
             @engine_options << @args.shift
           when "--dev"
             @dev_mode = true
+          when "--timeout"
+            ENV["CONTAINER_TIMEOUT_SECONDS"] = @args.shift
```
GordonDiggs: Why is this preferable to using the env variable directly?

ABaldwinHunter: @GordonDiggs good call.

Thought I saw a case where ENV wasn't working but I think it was just that the engine ran for less than 1 second. 

Thanks!

PR #484, opened 2016-08-23
Don't update marketing site data on deploy

wfleming:
New pricing page doesn't show this data. We don't need this anymore.

cc @codeclimate/review


pbrisbin:
Yessssss

PR #495, opened 2016-10-05
Add beta channel for duplication engine

ABaldwinHunter:
@codeclimate/review 

Longer term, I wonder if it makes sense to keep this registry up to date with available channels, or to attempt to install and analyze with them as a docker tag indiscriminately.


maxjacobson:
LGTM

Agreed that this is slightly tedious

PR #496, opened 2016-10-07
Fix a typo in README

maxjacobson:
It seems like this comment-in-a-comment was copied from the earlier
example, but doesn't pertain here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,14 +139,16 @@ The following is a brief explanation of each available command.
 * To increase the amount of time each engine container may run (default 15 min):
 
   ```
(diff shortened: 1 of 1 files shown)


wfleming:
LGTM

(C++ style comments next to shell commands seems like a mistake in any case.)

wfleming:
Actually, wait, this isn't valid shell anyway, is it? Underscores in numbers won't work setting this in ENV AFAIK. If these are intended to be runnable commands, should that be changed?

maxjacobson:
@wfleming heh, good point! Updated

wfleming:
LGTM (for realz).

PR #505, opened 2016-10-14
Update README.md

gerard2p:
Added step to use the command in windows 10


gerard2p:
Sorry this is actually my first pull request and i edit it directly from  gihub, I've also a way to use codeclimate with the linter-codeclime, I'll make a pull request to that repo.

dblandin:
Instead of adding these detailed instructions for Windows which dive into docker-specific commands, what do you think of creating a Windows-compatible version of the codeclimate-wrapper script? The existing script is automatically installed via homebrew (OSX) and an AUR (Arch Linux) package: https://github.com/codeclimate/codeclimate/blob/master/codeclimate-wrapper

GordonDiggs:
Closing due to staleness. Feel free to reopen if this picks up again

PR #506, opened 2016-10-15
Update Readme fixing code Path

mmornati:
The standard readme proposes the environment variable to define where is the code to analyse (for the docker machine) as $PWD. In reality the code, internal to the docker, is into the */code* folder which is mounted on $PWD.
I think leaving it in that way can bring to confusion...


wfleming:
Good catch, thanks @mmornati

Issue #422, opened 2016-05-05
Fail with Error Code if Analysis Engine Finds Issues?


joncursi:
If the analysis engine finds issues with your code, no error code is returned. The results are printed to console and the process is marked as a success. This makes it impossible to bake CodeClimate CLI into automated workflows that listen for such failures (aka when things "break").

pbrisbin:
I'm genuinely surprised by this, but have confirmed that's what's happening. This was either an oversight or a regression; either way, we'll look to fix soon. Thanks!

joncursi:
@pbrisbin is this still on the roadmap?

pbrisbin:
Hi @joncursi, sorry I was a little over-eager with my use of "soon" previously. We agree this is an issue, but we have no concrete plans to work on it quite yet.

devth:
I'm interested in this issue and potentially helping out.

1. isn't this contradictory to the [spec](https://github.com/codeclimate/spec/blob/master/SPEC.md#output)?
2. why was https://github.com/codeclimate/codeclimate/pull/508 closed?

pbrisbin:
Hey @devth thanks for offering to take a swing at this!

1. This doesn't contradict the SPEC. The SPEC is referring to engines themselves, which should not exit non-zero when they find issues; the behavior of the CLI is not part of the SPEC
1. That PR just got stale. It looks like there were some open discussion threads but, as is usually the case with stale PRs, I don't really remember the specifics

devth:
I think individual engines should be allowed to fail based on config. For example:

```yaml
engines:
  markdownlint:
    enabled: true
  fixme:
    enabled: true
    allow_failure: true
  shellcheck:
    enabled: true
```

In this case `allow_failure: true` is set on the `fixme` engine. Reasoning is that it can be useful to see those issues in the build log, but I don't want to fail the build because of them.

The default would be `allow_failure: false` meaning if any issues were found for an engine then `analyze` would `exit 1`.

Thoughts?

pbrisbin:
That seems like a worthwhile extension, but I would suggest doing it separately. Personally, I think exiting non-zero when issues are found is a saner default which could be implemented separate from, and before having the option to turn it off engine-by-engine.

I could be convinced that a global flag to go back to the old behavior would be useful so we don't just start erroring folks' builds, but engine-by-engine seems like clear feature-creep that can be tackled after.

dahlb:
[...]

PR #508, opened 2016-10-18
Exit with a return code of 1 if issues are found

dblandin:
Most CI processes are expected to exit with a return code of 0 in the
success mode and with a return code of 1 in the failure mode.

This commit leverages the formatter's knowledge of issues emitted to
determine the appropriate return code.

Addresses #422 

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -38,7 +38,7 @@ module CC
           output = EngineOutput.new(raw_output)
 
           unless output.valid?
-            stdout_io.failed("#{qualified_name} produced invalid output: #{output.error[:message]}")
+            stdout_io.errored("#{qualified_name} produced invalid output: #{output.error[:message]}")
             container.stop
           end
 
--- a/lib/cc/analyzer/formatters/plain_text_formatter.rb
+++ b/lib/cc/analyzer/formatters/plain_text_formatter.rb
@@ -42,11 +42,11 @@ module CC
           end
         end
 
-        def failed(output)
-          spinner.stop("Failed")
-          puts colorize("\nAnalysis failed with the following output:", :red)
+        def errored(output)
+          spinner.stop("Errored")
+          puts colorize("\nAnalysis errored with the following output:", :red)
           puts output
-          exit 1
+          exit 2
         end
 
         private
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -20,6 +20,9 @@ module CC
           runner.run
         end
 
+        unless formatter.empty?
+          exit 1
+        end
       rescue EnginesRunner::InvalidEngineName => ex
         fatal(ex.message)
       rescue EnginesRunner::NoEnabledEngines
--- a/lib/cc/analyzer/formatters/formatter.rb
+++ b/lib/cc/analyzer/formatters/formatter.rb
@@ -40,6 +40,10 @@ module CC
         def failed(_output)
         end
 
+        def empty?
+          issues.empty?
+        end
+
         InvalidFormatterError = Class.new(StandardError)
 
         private
--- a/lib/cc/analyzer/formatters/formatter.rb
+++ b/lib/cc/analyzer/formatters/formatter.rb
@@ -37,7 +37,7 @@ module CC
         def close
         end
 
-        def failed(_output)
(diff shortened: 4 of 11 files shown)


lib/cc/cli/analyze.rb
```diff
@@ -20,6 +20,9 @@ def run
           runner.run
         end
 
+        unless formatter.empty?
+          exit 1
```
wfleming: We also `exit 1` when analysis errored (which we unfortunately chose to call "failed" in `Formatter`). To distinguish those cases maybe we should exit with a different code for this case?

pbrisbin: I agree. Looking forward to when Builder just invokes `codeclimate` as a (containerized) process, the exit status will be part of the interface, and a specific status for issues found makes sense.

lib/cc/analyzer/formatters/plain_text_formatter.rb
```diff
@@ -42,11 +42,11 @@ def engine_running(engine, &block)
...
 
-        def failed(output)
-          spinner.stop("Failed")
-          puts colorize("\nAnalysis failed with the following output:", :red)
+        def errored(output)
+          spinner.stop("Errored")
+          puts colorize("\nAnalysis errored with the following output:", :red)
           puts output
-          exit 1
+          exit 2
```
pbrisbin: Are you saying an error is `2` and reserving issues found as `1`?

Looking at http://tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF

> 1	Catchall for general errors
> 2	Misuse of shell builtins (according to Bash documentation)
>
> According to the above table, exit codes 1 - 2, 126 - 165, and 255 [1] have special meanings, and should therefore be avoided for user-specified exit parameters

I think it's fine to use `1` for general errors. For issues-found, it looks like 3-125 or 166-255 is available; I'd prefer going above 165, myself.

lib/cc/analyzer/engine.rb
```diff
@@ -38,7 +38,7 @@ def run(stdout_io, container_listener)
           output = EngineOutput.new(raw_output)
 
           unless output.valid?
-            stdout_io.failed("#{qualified_name} produced invalid output: #{output.error[:message]}")
+            stdout_io.errored("#{qualified_name} produced invalid output: #{output.error[:message]}")
```
pbrisbin: I'm :+1: on this cleanup. Are there builder-side changes to conform to the new Formatter interface?

lib/cc/analyzer/formatters/plain_text_formatter.rb
wfleming: We could use `170` for for "analysis failed/issues found", and do increments of 10 for each category when we do error-class specific status codes. (e.g. `180` for a G10, `181` for a G11, `190` for a `C10`, etc.)

pbrisbin: I like it, but I'm not sure how much mileage we'd get out of that idea since most of the individual error codes can be determined externally. A non-zero exit from analyze and validate-config are probably the only two cases that could use differentiation.

wfleming: We could, probably: I think keeping them unique at the CLI level is desirable, though, if we're going to differentiate them at all. The grouping-by-10 suggestion is primarily so that each category would still be contiguous if we decided to add a new code to a category.

pbrisbin: I guess what I'm saying is, aside from analyze and validate config, there really is no other cases where the CLI would be involved:

- `G*` these are a clone in a totally different container
- `E*` these are just error vs timeout which builder knows (error is non-zero exit, timeout is no exit at at all, same for output exceeded, etc)

Not super important now; 170 for issues-found is fine with me regardless of future planning.

wfleming: Oh, got it. You're quite right about `G`. I think `E` timeouts vs errors would still be CLI territory: CLI already knows about timeouts & in the "thin-builder" world I think that might continue to be the case?

Yeah, not relevant for this PR anyway: just starting with `170` here is plenty.

dblandin:
Going to go ahead and close this PR for now. Can bring it up later if we feel like it's an worthwhile enhancement.

PR #512, opened 2016-11-01
Update default inferred rubocop config

maxjacobson:
This one was renamed, so it emits a deprecation warning now.

We should do a more thorough audit at some point -- There are some new
cops that have been introduced which are disabled and not listed, so
people don't easily know how to enable them.


wfleming:
Cool. LGTM. We should probably find a way to dogfood inferred configs better for this kind of issue.

PR #523, opened 2016-11-10
Add dependencies command

wfleming:
This adds a new command called `dependencies`, which will process entries in the `dependencies.files` key & download files.

cc @codeclimate/review this is mostly the same code, just tweaked a bit, but review would be appreciated!

### DIFF
--- a/lib/cc/cli.rb
+++ b/lib/cc/cli.rb
@@ -11,6 +11,7 @@ module CC
     autoload :Command, "cc/cli/command"
     autoload :Console, "cc/cli/console"
     autoload :Engines, "cc/cli/engines"
+    autoload :Dependencies, "cc/cli/dependencies"
     autoload :Help, "cc/cli/help"
     autoload :Init, "cc/cli/init"
     autoload :Runner, "cc/cli/runner"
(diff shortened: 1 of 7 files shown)


pbrisbin:
For the UX, I imagined it'd just be `codeclimate dependencies`, just because today it only has a `files` sub-key and only takes that action, I still see the command as broader and open to future extension. WDYT?

pbrisbin:
`codeclimate dependencies --files` might be another options, if you really want to make the "I'm only doing `files` things" explicit.

wfleming:
> For the UX, I imagined it'd just be codeclimate dependencies. Just because today it only has a files sub-key and only takes that action, I still see the command as broader and open to future extension. WDYT?

@pbrisbin I think we'd still want to be able to execute individual sub-parts of `dependencies` independently. Were you thinking a git-style `codeclimate dependencies files` for that case, and a plain `codeclimate dependencies` do (potentially) everything?

pbrisbin:
Yes, I think `codeclimate dependencies` would do everything (which is today just the `files` stuff). Then if/when we extend further we can decide if:

- A sub-sub command makes sense (your suggestion)
- A flag makes sense (my second suggestin)
- It actually doesn't ever make sense to do "part" of the dependencies step (what I think is likely)

I don't see it backing us into anything, personally.

wfleming:
@pbrisbin updated

pbrisbin:
Cool, LGTM.

PR #525, opened 2016-11-11
Rename dependencies command to prepare

wfleming:
This is following the key change in the YAML: https://github.com/codeclimate/codeclimate-yaml/pull/37

cc @codeclimate/review


lib/cc/cli/prepare.rb
```diff
@@ -38,16 +38,16 @@ def allow_internal_ips?
         @args.include?("--allow-internal-ips")
       end
 
-      def files
-        @files ||= config.dependencies && config.dependencies.files
+      def fetches
+        @fetches ||= config[:prepare] && config[:prepare].fetch
```
wfleming: @pbrisbin You were on the right track with thinking there was a method name collision, but the collision was (surprisingly) with `#prepare`, not fetch. `Node#prepare` in `cc-yaml` is basically a method nodes can use to set themselves up with initial state. So when the code here called `config.prepare` (what I had originally), it wasn't fetching my `Prepare` node, but calling the default empty `#prepare` method which returns `nil`, resulting in thinking there were no fetch entries configured. Using `#[]` is a workaround here to actually access the node we want :hurtrealbad:.

As for why the presenting symptom was silent RSpec failure, I've addressed & explained that in 7e302d3.

wfleming:
@codeclimate/review this is ready for review

GordonDiggs:
LGTM

PR #534, opened 2016-12-14
Fix README help for `analyze` command

wfleming:
The docs as written-here were confusing. For one thing, they were incorrectly
implying that `path` is a required argument: it's not, and providing a path like
`.` has suprising behavior for a lot of users.

This updates the short-help to make `path` more obviously optional, and adds a
paragraph to the fuller description below explaining the behavior of `analyze`
more fully both when `path` is provided & when it is not.

cc @codeclimate/review this has come up in a few GH issues, and I realized our docs were misleading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,10 @@ A list of available commands is accessible by running `codeclimate` or
 $ codeclimate help
 
 Available commands:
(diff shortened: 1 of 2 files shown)


toddmohney:
LGTM

PR #536, opened 2016-12-15
Fix explicit docker example in README

wfleming:
The example value given for `CODECLIMATE_CODE` in the explicit `docker run` example is wrong & won't work (unless you actually happen to be editing a project located at `/code` on your computer.) This ENV var needs to be the code path *on the host*, because it's used to construct the mount paths for sub-containers like engines.

cc @codeclimate/review


ABaldwinHunter:
Nice fix

Issue #541, opened 2017-01-09
Improve HTML formatter


GordonDiggs:
Our [HTML formatter](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/analyzer/formatters/html_formatter.rb) is fine, but its output could be improved. We should mostly match the format of [the website's issue view](https://codeclimate.com/github/codeclimate/codeclimate/issues). We can omit the following:

- Filtering
- Disable engine/check/issue modal
- Create ticket buttons
- Pagination
- Links to the file and engine (they can just be text)
- Displaying other instances of duplication

Otherwise, the issue display should be able to be mimicked pretty closely.

PR #544, opened 2017-01-17
Update HTML formatter to better match Web version

pointlessone:
image

## What's included

* Design matching web reports
* Code syntax highlighting
* Offending lines highlighting
* Code block expansion
* Filtering
* Duplication instances
* Extended descriptions (when available)

## Performance

Reports can get pretty big. This iteration is tested with over 400 issues in the report. It loads to a usable state in under a second in Firefox, Chrome and Safari.

Code highlighting is done when code block first time enters view. Layout is also optimized for minimum reflows during loading. Filtering is also snappy (under 300ms in Firefox).

---

Related issue: #541

### DIFF
--- a/lib/cc/analyzer/formatters/html_formatter.rb
+++ b/lib/cc/analyzer/formatters/html_formatter.rb
@@ -3,14 +3,164 @@ require "redcarpet"
 module CC
   module Analyzer
     module Formatters
-      class HTMLFormatter < Formatter
-        LANGUAGES = {
-        }
+      class HTMLFormatter < Formatter # rubocop: disable Metrics/ClassLength
+        LANGUAGES = Hash.new { |_, ext| ext }.
+                    merge(
+                      # abap
+                      # ada
+                      "appraisals" => "ruby",
+                      "as" => "actionscript",
+                      "asm" => "nasm",
+                      "bas" => "basic",
+                      # c
+                      "c++" => "cpp",
+                      "capfile" => "ruby",
+                      "cc" => "cpp",
+                      "cfc" => "markup",
+                      "cfm" => "markup",
+                      "coffee" => "coffeescript",
+                      "cp" => "cpp",
+                      # cpp
+                      "cr" => "crystal",
+                      "cs" => "csharp",
+                      "css" => %w[css css-extras],
+                      "cu" => "cpp",
+                      "cxx" => "cpp",
+                      # d
+                      # dart
+                      # diff
+                      "dockerfile" => "docker",
+                      "dpr" => "pascal",
+                      "erl" => "erlang",
+                      "ex" => "elixir",
+                      "f" => "fortran",
+                      "f90" => "fortran",
+                      "f95" => "fortran",
+                      "feature" => "gherkin",
+                      "for" => "fortran",
+                      "fs" => "fsharp",
+                      "fsi" => "fsharp",
+                      "fsscript" => "fsharp",
+                      "fsx" => "fsharp",
+                      "gemfile" => "ruby",
+                      "gemspec" => "ruby",
+                      # glsl
+                      # go
+                      # groovy
+                      "gvy" => "groovy",
+                      "h" => "c",
+                      "h++" => "cpp",
+                      # haml
+                      # handlebars
+                      "hbr" => "handlebars",
+                      "hh" => "cpp",
+                      "hpp" => "cpp",
+                      "hs" => "haskell",
+                      "htm" => "markup",
+                      "html" => "markup",
+                      "hx" => "haxe",
+                      "hxml" => "haxe",
+                      "icn" => "icon",
+                      "ijs" => "j",
+                      # ini
+                      "iol" => "jolie",
+                      # java
+                      "jl" => "julia",
+                      "js" => "javascript",
+                      # json
+                      # jsx
+                      "kt" => "kotlin",
+                      "kts" => "kotlin",
+                      # less
+                      "lhs" => "haskell",
+                      "lol" => "lolcode",
+                      "lols" => "lolcode",
+                      "ls" => "livescript",
+                      # lua
+                      "m" => "objective-c",
+                      "mab" => "ruby",
+                      # makefile
+                      # markdown
+                      "md" => "markdown",
+                      # mel
+                      "mkd" => "markdown",
+                      "ml" => "ocaml",
+                      "mli" => "ocaml",
+                      "mm" => "objective-c",
+                      # nim
+                      # nix
+                      "nsi" => "nsis",
+                      "ol" => "jolie",
+                      # oz
+                      "pas" => "pascal",
+                      "patch" => "diff",
+                      "pde" => "processing",
+                      "php" => %w[php php-extras],
+                      "php3" => %w[php php-extras],
+                      "php4" => %w[php php-extras],
+                      "php5" => %w[php php-extras],
+                      "phtml" => %w[php php-extras],
+                      "pl" => "perl",
+                      "pp" => "puppet",
+                      "prawn" => "ruby",
+                      "pro" => "prolog",
+                      # properties
+                      # pure
+                      "py" => "python",
+                      "py3" => "python",
+                      "pyw" => "python",
+                      "q" => "qore",
+                      "qm" => "qore",
+                      "qtest" => "qore",
+                      # r
+                      "rake" => "ruby",
+                      "rakefile" => "ruby",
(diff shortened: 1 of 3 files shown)


lib/cc/analyzer/formatters/templates/html.erb
```diff
@@ -1,53 +1,697 @@
...
+          <label for="engine-filter">Engine</label>
+          <select id="engine-filter">
+            <option value="">All Engines</option>
+            <%- engines.each do |engine| -%>
+            <option value="<%= param engine %>"><%= engine %></option>
+            <%- end -%>
+          </select>
+        </div>
+        <ul id="smells">
+          <%- issues.each do |issue| -%>
```
wfleming: Would it be possible to get an empty state for when there are no issues to show?

There is, of course, the overall case of no issues having been found at all, but it's also possible to reach an "empty issues" state by selecting a combination of engine/category filters (e.g. style/duplication will yield no results for any repo that happens to include both style issues & duplication issues): it would be great if we could also show some copy for that instead of just a blank page.

lib/cc/analyzer/formatters/html_formatter.rb
```diff
@@ -1,34 +1,280 @@
...
+            markdown(content)
+          end
+
+          def code(path)
+            h filesystem.read_path(path)
+          end
+
+          def location(issue)
+            (lines(issue["location"]["lines"]) ||
+             positions(issue)).uniq.sort.join("-")
```
pointlessone: Indeed. Well spotted.

lib/cc/analyzer/formatters/html_formatter.rb
```diff
@@ -1,34 +1,280 @@
...
+          end
+
+          def other_locations(issue)
+            locations = issue.fetch("other_locations", [])
+            Hash[
+              locations.map do |loc|
+                [loc["path"], lines(loc["lines"]).uniq.sort.join("-")]
```
pointlessone: Is this documented anywhere?

I wonder how this is reflected in the UI. Should I just display them one after another in a plain list or is there any hierarchy to it?

lib/cc/analyzer/formatters/templates/html.erb
```diff
@@ -1,53 +1,697 @@
...
+          1);
+        });
+      });
+    </script>
+    <style>
+      <%- categories.each do |category| -%>
+      .filter-category-<%= param category %> > li,
+      <%- end -%>
+      <%- engines.each do |engine| -%>
+      .filter-engine-bug-<%= param engine %> > li,
```
pointlessone: This is a… bug. It shouldn't be there.

lib/cc/analyzer/formatters/html_formatter.rb
```diff
@@ -1,34 +1,280 @@
...
-require "active_support/number_helper"
 
 module CC
   module Analyzer
     module Formatters
-      class HTMLFormatter < Formatter
+      class HTMLFormatter < Formatter # rubocop: disable Metrics/ClassLength
+        LANGUAGES = Hash.new { |_, ext| ext }.
+                    merge(
+                      # abap
```
pointlessone: The commented out are the ones that map to themselves.

The default is to use file extension as the syntax used to highlight that file. So entry like `'abap' => 'abap'` is redundant. But I left comments for the ones that are actually supported by Prism.

This list is completely based on what's supported by Prism, not what CC can actually  analyze.

lib/cc/analyzer/formatters/templates/html.erb
pointlessone: I'm not a designer so comments on visual part are welcome.

Here's how the report looks when no issues match selected filters:
screen shot 2017-01-19 at 14 35 15

And here's how it looks when no issues have been found:
screen shot 2017-01-19 at 14 36 05

lib/cc/analyzer/formatters/html_formatter.rb
wfleming: Ahh, right, I saw that default-val code but didn't make the connection. Thanks for the explanation.

lib/cc/analyzer/formatters/html_formatter.rb
wfleming: > Is this documented anywhere?

Yes, the spec [covers this](https://github.com/codeclimate/spec/blob/master/SPEC.md#issues). Sorry if that wasn't explained earlier in the process or if I saddled you with fixing a pre-existing bug in this instance.

> I wonder how this is reflected in the UI

Plain list (which I think is what you went with?) is totally fine.

wfleming:
Thanks for addressing those issues, @pointlessone. I turned up something else doing some local QA, that I didn't encounter yesterday so this might have been introduced as part of the recent changes?

It looks like each code block gets the source code of the *entire file*, which has a  severe impact on file size & browser render performance. Generating the HTML report for this repository turned up 71 issues, and the HTML file was 236 MB, which Chrome choked on (entirely independent of the Javascript: just parsing/rendering the HTML itself was problematic).

On the site we have a restriction in place for how many lines we'll render for a given issue for similar performance reasons: can we introduce something similar here? Specifically, can we include a maximum of 10 lines of code in the markup for any code block? It's fine if users need to go the actual source to see related lines or all of the content for issues that cover more than 10 lines. For issues less than 10 lines, it's also fine to just render the relevant line(s).

pointlessone:
Whoa! 236MB! This is unexpected.

Initially I only included the relevant lines (plus 2 lines of context on each side). Even though there was no heredocs, delimited comments or multiline strings in the issues the syntax highlighting still didn't look right all the time.

I decided to include full source for two main reasons:
1. I wonted to include "Expand" functionality as 2 lines of context is not always enough.
2. It's hard to always properly highlight syntax of a code cut at arbitrary points.

I tested it on this repo and with 61 issues it's only 1.2 MB (it was over 400 issues and just over 1.6 MB when I worked on it — wrong quotes on strings). The full source (`lib` and `spec`) is about 224 KB. It would need to be fully included over 1000 times to reach that size. I wonder how many issues are there on your repo and how big is the source.

wfleming:
Ah, I see what happened.

I had test coverage results locally in `coverage/index.html` (which isn't excluded from analysis because it's not normally there), and that test coverage file has several `TODO` comments in it (most of which seem to actually be from Simplecov). That coverage file is > 8 MB, and there were 21 issues found in it, so that's ~168 MB right there.

While this isn't an especially common case, and in this particular instance I should have excluded that dir from analysis, it's also not really *unusual* for customers to have quite large files that might have several issues in  them, and when that happens the "include all the source" strategy can get problematic quickly.

You make good points about context, so including context around smaller instances is entirely reasonable, but I still think we need to enforce a limit on the total number of lines put into the HTML for any given code block.

pointlessone:
Updated to include only  relevant piece of code up to 10 lines (plus up to 2 lines of context on each side). Removed "View more" links as that would've defeated the effort.

Report file size dropped significantly (from 1.2MB to 185KB in my test case).

lib/cc/analyzer/formatters/templates/html.erb
```diff
@@ -1,53 +1,637 @@
...
+            <%- issues.each do |issue| -%>
+            <li
+              data-categories="<%= params issue.categories %>"
+              data-engine="<%= param issue.engine_name %>">
+              <h2><%= h issue.description %></h2>
+              <div class="code">
+                <pre class="line-numbers <%= issue.source.syntaxes.map {|l| "language-#{l}" }.join ' ' %>"
+                     data-line="<%= issue.location %>"
+                     data-start="<%= issue.location.start %>"
+                     data-line-offset="<%= issue.location.line_offset %>"><code><%=
```
wfleming: Are these `data-line-offset` attrs actually used? It looked like 9ba3a28d4ccc2baec8fbd8b3c06439691c005254 removed the JS referring to them. Does `prism` use this internally or something?

pointlessone: `data-line-offset` is used by Prism's line highlight plugin. The one that gives different background to affected lines.

The removed code was based on that plugin and used it for the same calculations the plugin does. Removed code was used to properly position the whole source within a small window showing relevant piece of code. And it was removed since we don't do that any more (we don't position the code, we display the whole piece available for Prism now).

pointlessone:
I'd recommend squashing it before merging. I kept commits to (hopefully) make reviews a bit easier but there's no need to keep them as they're not very meaningful on their own.

PR #553, opened 2017-01-26
Extend help command

pointlessone:
Adds `codeclimate help command` which displays extended help for a command. It's possible to pass multiple arguments to `help` subcommand.

Help summary also shows short help for each of available commands.

### DIFF
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -1,6 +1,17 @@
+require "cc/cli/command"
+
 module CC
   module CLI
     class Analyze < Command
+      ARGUMENT_LIST = "[-f format] [-e engine[:channel]] [path]".freeze
+      SHORT_HELP = "Run analysis with the given arguments".freeze
+      HELP = "#{SHORT_HELP}\n" \
+        "\n" \
+        "    -f <format>, --format <format>   Format of output. Possible values: #{CC::Analyzer::Formatters::FORMATTERS.keys.join ", "}\n" \
+        "    -e <engine[:channel]>            Engine to run. Can be specified multiple times.\n" \
+        "    --dev                            Run in development mode.\n" \
+        "    path                             Path to check. Can be specified multiple times.".freeze
+
       include CC::Analyzer
 
       def initialize(_args = [])
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -9,7 +9,7 @@ module CC
         "\n" \
         "    -f <format>, --format <format>   Format of output. Possible values: #{CC::Analyzer::Formatters::FORMATTERS.keys.join ", "}\n" \
         "    -e <engine[:channel]>            Engine to run. Can be specified multiple times.\n" \
-        "    --dev                            Run in development mode.\n" \
+        "    --dev                            Run in development mode. Engines installed locally that are not in the manifest will be run.\n" \
         "    path                             Path to check. Can be specified multiple times.".freeze
 
       include CC::Analyzer
--- a/lib/cc/cli/command.rb
+++ b/lib/cc/cli/command.rb
@@ -7,7 +7,7 @@ module CC
   module CLI
     class Command
       CODECLIMATE_YAML = ".codeclimate.yml".freeze
-      NAMESPACE = name.split('::')[0..-2].join("::").freeze
+      NAMESPACE = name.split("::")[0..-2].join("::").freeze
 
       def self.abstract!
         @abstract = true
@@ -25,10 +25,12 @@ module CC
         all.find { |command| command.name == "#{NAMESPACE}::#{name}" || command.command_name == name }
       end
 
+      # rubocop: disable Style/ClassVars
       def self.inherited(subclass)
         @@subclasses ||= []
         @@subclasses << subclass
       end
+      # rubocop: enable Style/ClassVars
 
       def self.synopsis
         "#{command_name} #{self::ARGUMENT_LIST if const_defined?(:ARGUMENT_LIST)}".strip
--- a/lib/cc/cli/command.rb
+++ b/lib/cc/cli/command.rb
@@ -7,6 +7,48 @@ module CC
   module CLI
     class Command
       CODECLIMATE_YAML = ".codeclimate.yml".freeze
+      NAMESPACE = name.split('::')[0..-2].join("::").freeze
+
+      def self.abstract!
+        @abstract = true
+      end
+
+      def self.abstract?
+        @abstract == true
+      end
+
+      def self.all
+        @@subclasses.select { |klass| !klass.abstract? }
+      end
+
+      def self.[](name)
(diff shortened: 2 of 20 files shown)


lib/cc/cli/help.rb
```diff
@@ -3,30 +3,45 @@
...
-          "engines:remove #{underline("engine_name")}",
-          "prepare",
-          "help",
-          "init",
-          "validate-config",
-          "version",
-        ].freeze
+      def show_help(command_name)
+        command = Command[command_name]
+        say "Usage: #{$PROGRAM_NAME} #{command.synopsis}\n"
```
GordonDiggs: this `$PROGRAM_NAME` seems to be expanded into a full path:

```
Usage: /usr/src/app/bin/codeclimate analyze [-f format] [-e engine[:channel]] [path]
```

It will always be `codeclimate` so I would be ok with hardcoding that here

lib/cc/cli/analyze.rb
```diff
@@ -1,6 +1,17 @@
...
 module CC
   module CLI
     class Analyze < Command
+      ARGUMENT_LIST = "[-f format] [-e engine[:channel]] [path]".freeze
+      SHORT_HELP = "Run analysis with the given arguments".freeze
+      HELP = "#{SHORT_HELP}\n" \
+        "\n" \
+        "    -f <format>, --format <format>   Format of output. Possible values: #{CC::Analyzer::Formatters::FORMATTERS.keys.join ", "}\n" \
+        "    -e <engine[:channel]>            Engine to run. Can be specified multiple times.\n" \
+        "    --dev                            Run in development mode.\n" \
```
GordonDiggs: Might be good to expand on this. WDYT of "Run in development mode. Engines installed locally that are not in the manifest will be run"

lib/cc/cli/command.rb
```diff
@@ -7,6 +7,50 @@ module CC
...
     class Command
       CODECLIMATE_YAML = ".codeclimate.yml".freeze
+      NAMESPACE = name.split("::")[0..-2].join("::").freeze
+
+      def self.abstract!
+        @abstract = true
+      end
+
+      def self.abstract?
+        @abstract == true
```
GordonDiggs: 💄 probably enough to see if this is truthy and just return `@abstract`

lib/cc/cli/command.rb
```diff
@@ -7,6 +7,50 @@ module CC
...
+      end
+
+      def self.all
+        @@subclasses.select { |klass| !klass.abstract? }
```
GordonDiggs: Would it work to do

```
@@subclasses.reject(&:abstract?)
```

lib/cc/cli/console.rb
```diff
@@ -1,6 +1,8 @@
 module CC
   module CLI
     class Console < Command
+      SHORT_HELP = "Open a ruby console for the CLI. Only useful for people developing against the CLI".freeze
```
GordonDiggs: 📖  For the last sentence: "Useful for developing against the CLI"

lib/cc/cli/help.rb
```diff
@@ -3,30 +3,45 @@
 module CC
   module CLI
     class Help < Command
+      ARGUMENT_LIST = "[commad]".freeze
```
GordonDiggs: 📖  "commad" should be "command"

lib/cc/cli/help.rb
```diff
@@ -3,30 +3,45 @@
...
+      SHORT_HELP = "Show help.".freeze
```
GordonDiggs: 📖  "Display help information"

lib/cc/cli/help.rb
```diff
@@ -3,30 +3,45 @@
...
+      HELP = "#{SHORT_HELP}\n" \
+        "\n" \
+        "    no arguments   Show help summary for all commands.\n" \
+        "    [command]      Show help for specific commands. Can be scpecified multiple times.".freeze
```
GordonDiggs: 📖  "scpecified" should be "specified"

lib/cc/cli/init.rb
```diff
@@ -7,6 +7,12 @@ module CLI
     class Init < Command
       include CC::Analyzer
 
+      ARGUMENT_LIST = "[--upgrade]".freeze
+      SHORT_HELP = "Generate a configuration and install the right images based on the contents of your repo.".freeze
```
GordonDiggs: 📖  "Generate a configuration based on the contents of your repo."

We can leave out the image pulling for this

lib/cc/cli/init.rb
```diff
@@ -7,6 +7,12 @@ module CLI
...
+      HELP = "#{SHORT_HELP}\n" \
+        "\n" \
+        "    --upgrade    Upgrade existing configuration".freeze
```
GordonDiggs: 📖  "Upgrade a Code Climate Classic configuration"

lib/cc/cli/test.rb
```diff
@@ -68,6 +69,12 @@ def initialize(line, line_text)
     end
 
     class Test < Command
+      ARGUMENT_LIST = "<engine_name>".freeze
+      SHORT_HELP = "Test engine".freeze
```
GordonDiggs: 📖  "Test an engine"

lib/cc/cli/test.rb
```diff
@@ -68,6 +69,12 @@ def initialize(line, line_text)
...
+      HELP = "#{SHORT_HELP}\n" \
+        "\n"\
```
GordonDiggs: Could we add a line here that is something like "Validate that an engine is behaving correctly"

lib/cc/cli/validate_config.rb
```diff
@@ -3,6 +3,8 @@
 module CC
   module CLI
     class ValidateConfig < Command
+      SHORT_HELP = "Make sure your codeclimate.yml is valid.".freeze
```
GordonDiggs: 📖  "Validate your .codeclimate.yml`

lib/cc/cli/help.rb
pointlessone: `$PROGRAM_NAME` is just the first element of `$ARGV`. It depends on invocation. If the executable is in the `$PATH` and command is `codeclimate help` it fill be just `codeclimate`. If you're in the project dir and the command is `bin/codeclimate help` it will be `bin/codeclimate`. If invoked with full path it will be full path. I find it a bit more friendlier to mimic user's invocation style but I have no objections.

lib/cc/cli/command.rb
pointlessone: Lately I'm more inclined to use proper types. Query methods like this are supposed to return a boolean value. In Ruby that is either `true` or `false`. This variable is specifically set to `true` and there's no API to set it to anything else so all other values (including default `nil`) are `false`.

Getting `nil` from query method feels more like 🤷‍♀️ for an answer to a yes/no question.

I personally might've written `!!@abstract` but Rubocop doesn't like that.

PR #561, opened 2017-02-08
Check latest cli version

elcuervo:
TODO

- [x] - Cache last check
- [x] - Style warning on outdated versions


wfleming:
I realize this is WIP, so maybe this is planned, but IMHO we should also make this not fail out if the user isn't connected to the internet, or if the endpoint is down/unreachable. E.g timeouts, non-expected response bodies, non-200 codes.

lib/cc/cli/checker.rb
```diff
@@ -0,0 +1,80 @@
...
+    module Checker
+      TIMEOUT_CHECK = 60 * 60 # 1 Hour
+
+      def last_check
+        return { version: version, time: 0, outdated: false } unless has_config_file?("latest_version")
+        load_latest_version
+      end
+
+      def config(file)
+        dir = ENV.fetch("CODE_CLIMATE_CONFIG", File.expand_path(".code_climate", Dir.home))
```
GordonDiggs: Does this need to be mounted into the container?

lib/cc/cli/checker.rb
```diff
@@ -0,0 +1,80 @@
...
+          last_check["outdated"] = latest_version["outdated"] == true
+        end
+      end
+
+      def latest_version
+        @_remote ||= begin
+          url = ENV.fetch("CODE_CLIMATE_VERSIONS_URL", "https://versions.codeclimate.com")
+          CLI.debug("Checking #{url} for latest version")
+          values = { version: version, uname: `uname -a` }
+          resp = Faraday.get(url, values)
```
elcuervo: I used `net/http` before a refactor 🤕 I'll go back to that version

lib/cc/cli/checker.rb
```diff
@@ -0,0 +1,80 @@
...
+
+          JSON.parse(resp.body).merge("time" => Time.now.to_i)
+        end
+      end
+
+      def needs_update_to_version(latest)
+        warn Rainbow("~~~ Needs update to version #{latest}, currently #{version} ~~~").red
```
elcuervo: Perfect, it was just a placeholder.

pointlessone:
This is an updated version checker feature. It's more or less the same code just a bit refactored.

Notable changes:

* It uses UUID for identification; doesn't send out local username and other private information
* It implements XDG spec for configuration and caching
  * Config lives in `~/.config/codeclimate/config.yml` (by default)
  * Cache lives in `~/.cache/codeclimate/cache.yml` (by default)

⚠️ DO NOT MERGE ⚠️ 

It requires a small change to Versions API before it can be merged.

pbrisbin:
Hi @pointlessone there are some Docker complexities about storing config data which I'll have to work through with you to get this over the line. Could you do me a _huge_ favor? Please split this PR into two: one that implements the actual version checking code and another that deals with caching, XDG, etc.

The first PR (the one that actually does the version checking) should be guarded by a `--[no-]version-check` flag and default to _don't_. That should be a pretty quick review and we can ship that and get it out. 

Then we can follow up with the feature of caching last-checked etc, and making it on-by-default. Grab me in Slack before starting on that second PR and I can describe the Docker complexities we'll need to account for.

Is that OK?

pbrisbin:
@pointlessone I think this is superseded by the other two PRs, so I'm going to close this one.

PR #563, opened 2017-02-14
Update default .rubocop.yml for 0.47.1

maxjacobson:
These cops were renamed/moved

### DIFF
--- a/config/rubocop/.rubocop.yml
+++ b/config/rubocop/.rubocop.yml
@@ -81,10 +81,6 @@ Lint/EnsureReturn:
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-return-ensure'
   Enabled: true
 
-Lint/Eval:
-  Description: 'The use of eval represents a serious security risk.'
-  Enabled: true
-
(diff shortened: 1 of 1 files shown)


pbrisbin:
@maxjacobson is there anything blocking this from merge?

maxjacobson:
@pbrisbin yes. The codeclimate-rubocop engine is currently running rubocop 0.46.0. I'm planning to spend some time this week upgrading the codeclimate-rubocop engine from 0.46 to 0.47.1, after which we can ship this. I'd like to find a way to do it that tolerates outdated configuration more gracefully

pbrisbin:
Sounds good. This is further evidence that a better default-config system would allow the engines to themselves control what default configs are written for them.

jhabdas:
Update?

maxjacobson:
Update: we've added a channel to support RuboCop 0.48.1. We haven't switched to making it the default yet, but users can opt in by following the steps here: https://docs.codeclimate.com/docs/rubocop#section-using-rubocop-s-newer-versions

Thanks for the ping. Going to close this PR for now. We'll open a similar one when we change the default.

PR #564, opened 2017-02-15
Add Code Climate Checks

toddmohney:
(assume the feature is enabled)
With no existing `.codeclimate.yml` file:
- Running `codeclimate init` will generate a config file complete with _codeclimate checks_

With an existing `.codeclimate.yml` file
- Running `codeclimate init --upgrade` will add any missing _codeclimate checks_ to yer yahmal.

This an idempotent operation, so multiple upgrades will behave as expected (re: no dupes)

---

Holy smokes. It was suuuuuuuuper hard to figure out where the best place to make this change was. In the end I'm pretty happy with how it came out. Hi-fives to @pbrisbin for letting me torture him about this for several days.

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -50,6 +50,17 @@ checkstyle:
     - \.java$
   default_ratings_paths:
     - "**.java"
+complexity-ruby:
+  channels:
+    beta: codeclimate/codeclimate-complexity-ruby
+  description: Code Climate Complexity Checks for Ruby
+  community: false
+  code_climate_check: true
+  default_ratings_paths:
+    - "**.rb"
+  default_config:
+    languages:
+      - ruby
 csslint:
   channels:
     stable: codeclimate/codeclimate-csslint
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -52,7 +52,7 @@ checkstyle:
     - "**.java"
 complexity-ruby:
   channels:
-    beta: codeclimate/codeclimate-complexity-ruby
+    beta: codeclimate/codeclimate-complexity-ruby:b44
   description: Code Climate Complexity Checks for Ruby
   community: false
   code_climate_check: true
--- a/lib/cc/cli/config_generator.rb
+++ b/lib/cc/cli/config_generator.rb
@@ -38,10 +38,7 @@ module CC
       end
 
       def codeclimate_checks
-        return @codeclimate_checks if @codeclimate_checks
-
-        engines = engine_registry.list
-        @codeclimate_checks = engines.each_with_object({}) do |(name, config), result|
+        @codeclimate_checks ||= engine_registry.list.each_with_object({}) do |(name, config), result|
           if code_climate_check?(config)
             result[name] = config
           end
--- a/lib/cc/cli/config_generator.rb
+++ b/lib/cc/cli/config_generator.rb
@@ -37,6 +37,17 @@ module CC
         end
       end
 
+      def codeclimate_checks
+        return @codeclimate_checks if @codeclimate_checks
+
+        engines = engine_registry.list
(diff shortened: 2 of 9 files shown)


config/engines.yml
```diff
@@ -426,3 +426,14 @@ vint:
     - \.vim$
   default_ratings_paths:
     - "**.vim"
+complexity-ruby:
```
GordonDiggs: The engines in this file are sorted alphabetically. Can you move this into place?

config/engines.yml
```diff
@@ -426,3 +426,14 @@ vint:
...
+  channels:
+    stable: codeclimate/codeclimate-complexity-ruby
```
GordonDiggs: Should we have this in beta for now?

lib/cc/cli/config_generator.rb
```diff
@@ -63,6 +71,10 @@ def engine_eligible?(engine)
         !engine["community"] && engine["enable_regexps"].present? && files_exist?(engine)
       end
 
+      def code_climate_check?(engine)
+        engine["code_climate_check"] && ENV["BUILDER_RUN_CODECLIMATE_CHECKS"] == "true"
```
GordonDiggs: This ENV key shouldn't have the `BUILDER_` prefix

toddmohney:
@pbrisbin yep. you are correct on all counts.

toddmohney:
https://github.com/codeclimate/app/issues/3612

brynary:
@pbrisbin @toddmohney Pat is accurate that #2 is not desirable from a product standpoint. Normally it'd be an "implementation detail" but we expose generated configurations in the app and encourage people to download, commit and modify them, so anything we introduce into the configuration file is content that we can get "stuck" with (it's hard to get everyone to change their configurations).

As discussed, we do have the ability to ensure configuration files are as-needed because we will be hand-holding people through the setup of Pro. I wasn't envisioning the need to _go back_ and have them modify the configuration files again, but it's not a deal breaker, just more work for the customer success team.

My question is:

How far did we get, and what difficulties did we encounter implementing this on the read-side? (In essence, could we make the list of engines returned by the CLI include an extra element if the special ENV var is set?, for now)

toddmohney:
@pbrisbin @brynary I agree that this is not ideal. IMO, it is acceptable for now because it won't affect CLI customers in practice until we remove the feature flag. We have until then to come up with an alternative.

The motivation behind this implementation was to _use the pipeline as designed with minimal hacks_. This change is small and easy to move when we feel it would be necessary. Understanding the implication of pretty much _any_ change to the CLI and the resulting impacts on Builder is very difficult. We set the timebox for this work to 1 day, so this was the path of least resistance for me.

Happy to talk more about it, but what we have here will get us to a point where we can validate our work and unblock some stories. If small modifications need to be made before public release, I think that's acceptable.

toddmohney:
@brynary to answer your question:
> what difficulties did we encounter ... ?
Hacking something into the resulting engine list for analysis is doable (but gross, IMO). A problem arises, however, because the analysis step occurs _after_ required engines have been pulled. Now, a second hack is needed to get the engines which do not appear in the `.codeclimate.yml` file to be pulled. This step is implemented differently for the CLI and Builder. Not sure why.

It was my preference to make a small change that would functionally get us somewhere. With the alternative exhibiting signs of cascading hackery, I felt it best to work with the system as-is rather than try to fight it.

Again, I think I've made this easy to change as we explore our options.

pbrisbin:
The order dependency and code duplication/alteration with the pull step makes total sense, and is a clear reason why this is probably the least-bad option at this time. Thanks for clarifying that.

brynary:
Thanks for the additional explanation, Todd. Very helpful

PR #576, opened 2017-03-02
Check for all possible config names before generation

pointlessone:
CC CLI has default configs for a few engines it makes sure to not overwrite them but some engines accept a few files for configs (e.g. ESLint). In those cases CC CLI would generate a config even if there's a config for engine with a different name.

Here a new option is introduced in engine registry: `config_names`. It's a hash with keys being names of files in config template directory and values being arrays of possible alternative names.

See [example](https://github.com/pointlessone/codeclimate/blob/1b1580fc48cec93ae3ca310c1bd705bfd22d1526/config/engines.yml#L164-L169).

Template file name is included into a list of collision detection and doesn't need to be included into alternative names list. When the section is missing template file names used for collision detection anyway.

---

⚠️ 
This PR is for the `master` branch as is. There's a somewhat related PR #566. Either this or that PR will need to be updated depending on which is merged first.

### DIFF
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -74,14 +74,26 @@ module CC
 
       def create_default_engine_configs
         say "Generating default configuration for engines..."
-        available_engine_configs.each do |config_path|
-          file_name = File.basename(config_path)
-          if filesystem.exist?(file_name)
-            say "Skipping generating #{file_name} file (already exists)."
-          else
-            filesystem.write_path(file_name, File.read(config_path))
-            success "Config file #{file_name} successfully generated."
+        available_engine_configs.each do |(engine_name, config_paths)|
+          engine = engine_registry[engine_name]
+          config_mapping = Hash.new { |_, k| [k] }.merge(engine.fetch("config_files", {}))
+
+          config_paths.each do |config_path|
+            generate_config(config_path, config_mapping[File.basename(config_path)])
           end
+        end
+      end
+
+      def generate_config(config_path, possible_names)
+        file_name = File.basename(config_path)
+        existing_files = possible_names.select do |f|
+          filesystem.exist? f
+        end
+        if existing_files.any?
+          say "Skipping generating #{file_name}, existing file(s) found: #{existing_files.join(", ")}"
+        else
+          filesystem.write_path(file_name, File.read(config_path))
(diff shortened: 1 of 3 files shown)


lib/cc/cli/init.rb
```diff
@@ -74,27 +74,40 @@ def create_codeclimate_checks
...
       def create_default_engine_configs
         say "Generating default configuration for engines..."
-        available_engine_configs.each do |config_path|
-          file_name = File.basename(config_path)
-          if filesystem.exist?(file_name)
-            say "Skipping generating #{file_name} file (already exists)."
-          else
-            filesystem.write_path(file_name, File.read(config_path))
-            success "Config file #{file_name} successfully generated."
+        available_engine_configs.each do |(engine_name, config_paths)|
```
GordonDiggs: 💄 are the parentheses here required?

lib/cc/cli/init.rb
```diff
@@ -74,27 +74,40 @@ def create_codeclimate_checks
...
+          engine = engine_registry[engine_name]
+          config_mapping = Hash.new { |_, k| [k] }.merge(engine["config_files"] || {})
```
GordonDiggs: 💄 we prefer `fetch` here:

```ruby
engine.fetch("config_files", {})
```

lib/cc/cli/init.rb
```diff
@@ -74,27 +74,40 @@ def create_codeclimate_checks
...
         end
       end
 
+      def generate_config(config_path, possible_names)
+        file_name = File.basename(config_path)
+        existing_files = possible_names.select do |f|
+          filesystem.exist? f
+        end
+        if existing_files.any?
+          say "Skipping generating #{file_name} file (already exists: #{existing_files.join(", ")})."
```
GordonDiggs: I would reformat this a bit:

`Skipping generating #{file_name}, existing file(s) found: #{existing_files.join(", ")}).`

lib/cc/cli/init.rb
pointlessone: Not strictly in the latest Rubies. I personally prefer it that way as in this case it's destructuring. `available_engine_configs` is an array of arrays so block is technically passed a single argument, an array of two values.

GordonDiggs: Ah got it - makes sense

PR #578, opened 2017-03-02
Don't error if an unknown engine is encountered in `init`.

GordonDiggs:
This can error when in dev mode or when using engines not available in
the CLI

### DIFF
--- a/lib/cc/cli/init.rb
+++ b/lib/cc/cli/init.rb
@@ -75,11 +75,12 @@ module CC
       def create_default_engine_configs
         say "Generating default configuration for engines..."
         available_engine_configs.each do |(engine_name, config_paths)|
-          engine = engine_registry[engine_name]
-          config_mapping = Hash.new { |_, k| [k] }.merge(engine.fetch("config_files", {}))
+          if (engine = engine_registry[engine_name])
+            config_mapping = Hash.new { |_, k| [k] }.merge(engine.fetch("config_files", {}))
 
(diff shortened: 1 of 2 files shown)


lib/cc/cli/init.rb
```diff
@@ -75,11 +75,12 @@ def create_codeclimate_checks
       def create_default_engine_configs
         say "Generating default configuration for engines..."
         available_engine_configs.each do |(engine_name, config_paths)|
-          engine = engine_registry[engine_name]
-          config_mapping = Hash.new { |_, k| [k] }.merge(engine.fetch("config_files", {}))
+          if (engine = engine_registry[engine_name])
```
zenspider: This isn't ruby idiomatic and is an antipattern. Leave the assignment above. Depending on settings rubocop will probably bitch about it.

spec/cc/cli/init_spec.rb
```diff
@@ -126,6 +126,24 @@ module CC::CLI
...
+          describe "when an invalid engine is specified" do
+            it "does not error" do
+              config = <<-EOF
+                engines:
+                  hal9000:
+                    enabled: true
+              EOF
+              File.write(".codeclimate.yml", config)
+
+              _, _, exit_code = capture_io_and_exit_code do
```
zenspider: ooh. that seems useful. May I pull that into minitest (probably separate from capture_io)?

PR #582, opened 2017-03-03
Cache the other_location check by path.

zenspider:
This speeds up anything that makes use of other_locations extensively.
If one location is valid, all of them (with the same path) will be.
This took codeclimate-duplication on big_files from 603s to 108s.

More could probably be done to speed up the driver, but this was
definitely at the top of my profile.


zenspider:
Goes from:

blah

to:

blah3

noahd1:
👏

lib/cc/analyzer/issue_validations/other_locations_format_validation.rb
```diff
@@ -9,6 +9,11 @@ class OtherLocationsFormatValidation < Validation
           RelativePathValidation,
         ].freeze
 
+        def initialize(object)
+          super
+          @location = {}
+        end
```
pbrisbin: Instead doing this, WDYT of defining:

```rb
def other_locations
  @other_locations ||= {}
end
```

And using the reader at L34?

maxjacobson:
How did you generate those diagrams btw?

zenspider:
@maxjacobson integrated stackprof into `bin/codeclimate`. That'll be a separate PR.

@wfleming it won't go green as-is. There's a test that thwarts my change (and is right). I'm assuming that if one thing built an `other_location` for `path` properly, that they all will. The test shows that is false. Either the caching needs to be pushed down for JUST the path value, or some other assumptions need to be made about how the sub-validators are run. If I push down to the real path existance validator, the output is wrong because it bypasses the other 999 issues somewhere. (It lists 1 issue with 1 other location instead of 999 other locations)

@pbrisbin no benefit in using a lazy accessor and it makes the code less clear because you have to use `self.` or rename the arg where it is used.

pbrisbin:
@zenspider 

>  because you have to use self. or rename the arg where it is used.

I don't think so. You'd be calling `#[]` and `#[]=` on (basically) a reader, so none of the `self.` or scoping issues of a writer would come into play.

zenspider:
@maxjacobson https://github.com/codeclimate/codeclimate/pull/584

lib/cc/analyzer/issue_validations/other_locations_format_validation.rb
```diff
@@ -25,7 +30,8 @@ def message
         private
 
         def other_location_valid?(location)
-          CHECKS.all? do |klass|
+          path = location && location["path"]
+          @location[path] ||= CHECKS.all? do |klass|
```
pbrisbin: Can you rename `@location` to `@other_locations`? This would differentiate it from the local `location` variable, and indicate it's a Hash of many location values, not just one.

pbrisbin: Sorry, on second thought maybe `other_location_validities` would be better? I didn't realize this was caching the resulting boolean, not the locations themselves.

zenspider:
@pbrisbin done

zenspider:
@max any thoughts on that failing test? It's a valid test given the requirements... but maybe not a valid requirement. If an engine produces one `other_location` for a given path correctly, we can presume it will for all other values with the same path.

pbrisbin:
@zenspider I agree with this statement:

> If an engine produces one other_location for a given path correctly, we can presume it will for all other values with the same path.

But I also agree with the spec's docstring as written:

> CC::Analyzer::IssueValidations::OtherLocationsFormatValidation#valid? returns false if the other locations are not valid

Can the test be changed to produce _no_ valid other_locations, and such still be a good test?

zenspider:
@pbrisbin I don't understand the question as worded. Do you mean "change the test to have zero other_locations" or "change the test to have non-zero but entirely invalid other_locations"? 

Either way should be fine. The former wouldn't hit this code at all and the latter would cache the failure. If neither of those interpretations, can you please reword?

pbrisbin:
@zenspider "change the test to have non-zero but entirely invalid other_locations"

zenspider:
kk. fixed. This should be good to go and changes some of my runs from 9+ minutes to 3-4.

zenspider:
@GordonDiggs or @pbrisbin can you merge please?

GordonDiggs:
@zenspider Yep! I want to make sure we can monitor it after we ship it but I will merge it soon!

PR #584, opened 2017-03-07
Add support for top-level profiling via  stackprof.

zenspider:
Activated via CODECLIMATE_PROFILE=1.

Currently spits out both a plain text analysis and a graphviz... but
to stdout, since docker setup is strange and I don't know how to get
it back via a more normal route.


zenspider:
@pbrisbin done

zenspider:
I wrote that under contract with you... So you already have full rights to it.

PR #585, opened 2017-03-07
Dockerignore some common swap files

maxjacobson:
I noticed in our latest CLI release, we accidentally included a file
config/eslint/.eslintrc.yml.swp, and so this gets included when people
run `codeclimate init --upgrade` in their javascript repos.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,5 @@ bin/release
 circle.yml
 pkg
 tags
+*.sw*
+*~


.dockerignore
```diff
@@ -7,3 +7,6 @@ bin/release
 circle.yml
 pkg
 tags
+*.swp
```
GordonDiggs: I believe these can have any final character. Would `*.sw*` work here?

maxjacobson: SGTM - I know `.swf` is used for flash animations, so as long as we don't intend to include those in the future, we should be good

GordonDiggs: That's a limitation I'm willing to accept hahaha

If you're concerned, we could match `.*.sw*`

pointlessone:
This sort of things should be in everyone's personal global gitignore. People use all sorts of editors and tools. It's their responsibility to keep temp files created by their setup out of repo.

For instance, I use vim and ctags. It's my setup. It's not specific to any particular project I'm working on. It's my responsibility to keep files generated by vim and ctags out of git. Those files are not generated by anything in the repo. Someone who works on the same projects I do but uses Atom will never encounter swap files created by vim so they don't need to ever deal with them and there's no need for this piece of setup to ever be shared.

Just my 2 cents…

maxjacobson:
Re: your update - yep, exactly. Generally I agree with that philosophy though 😄

GordonDiggs:
Yeah I have these in my global gitignore (I think that's how I missed them and they ended up in the docker image). A global dockerignore would be awesome. Until then, we probably need to do this :/

maxjacobson:
I haven't shipped the CLI before, but after reviewing the release script it looks pretty straight-forward. Going for it

PR #586, opened 2017-03-08
Version checker

pointlessone:
A basic version checker triggered by a CLI parameter.

```bash
codeclimate --check-version version
```

It must be the first argument since it's a global change and not specific to any subcommand.

It still need a change to Versions API.

*Caveat*: Since there's no persistence UUID is generated on each check. This is probably not very useful.


lib/cc/cli/output.rb
```diff
@@ -0,0 +1,34 @@
...
+      def success(message)
+        terminal.say colorize(message, :green)
+      end
+
+      def say(message)
+        terminal.say message
+      end
+
+      def warn(message)
+        terminal.say(colorize("WARNING: #{message}", :yellow))
```
pbrisbin: :lipstick: I realize this was existing, but can you drop the outer parens so all invocations of `#say` are consistent here?

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        print_new_version_message if outdated?
+      end
+
+      private
+
+      def check_version?
+        first_arg = ARGV.first
+        if %w[--check-version].include? first_arg
+          ARGV.shift
```
pbrisbin: I think it's mixing responsibilities to have this class in charge of argument parsing. I would prefer this class just check the version and have the caller (`Runner`) invoke it or not based on the arguments (it's already got that responsibility).

```rb
class Runner
  def run
    VersionChecker.run if check_version?

```

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+        warn "A new version (v#{latest_version}) is available"
+      end
+
+      def api_response
+        @api_response ||=
+          begin
+            JSON.parse(api_response_body)
+          rescue JSON::ParserError => error
+            CLI.debug(error)
+            # We don't know so pretend all is peachy. We'll try again next time.
```
pbrisbin: WDYT about putting (only) a bare `rescue => ex`  in `#check`. IMO any failure in this class at all should be logged at DEBUG and ignored.

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+            {
+              "latest" => version,
+              "outdated" => false,
+            }
+          end
+      end
+
+      def api_response_body
+        Timeout.timeout(5) do
```
pbrisbin: Instead of doing this, setting `#open_timeout` and `#read_timeout` on the HTTP client should be functionally equivalent and more robust, right?

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+          if http_response.is_a? Net::HTTPSuccess
+            http_response.body
+          else
+            raise Net::HTTPFatalError.new("HTTP Error", http_response)
+          end
+        end
+      rescue Net::HTTPFatalError, Timeout::Error, Errno::EINVAL, Errno::ECONNRESET, EOFError, Errno::ECONNREFUSED => error
```
pbrisbin: This can also be removed if we put go with a "rescue everything in `#check`" approach.

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+          begin
+            uri = URI.parse(ENV.fetch("CODECLIMATE_VERSIONS_URL", DEFAULT_VERSIONS_URL))
+            uri.query = { version: version, uid: UUID.new.generate }.to_query
+            Net::HTTP.get_response(uri)
+          end
+      end
+
+      def version
+        @version ||=
+          File.read(File.expand_path("../../../VERSION", __dir__))
```
pbrisbin: Might not be worth it, but WDYT of DRYing this with https://github.com/codeclimate/codeclimate/blob/master/lib/cc/cli/version.rb somehow? I'm slightly concerned if we re-org this repo, we'll have to be careful to update multiple path references.

spec/cc/cli/version_checker_spec.rb
```diff
@@ -0,0 +1,52 @@
+require "spec_helper"
+
+RSpec.describe CC::CLI::VersionChecker do
```
pbrisbin: :lipstick: we don't use `RSpec.` in this codebase.

lib/cc/cli/version_checker.rb
pointlessone: Generally I try hard to avoid Pokemon exception handling (catch all exceptions). This is an expected one and this class knows how to handle it. There are other exceptions though. For example, Ctrl-C is an exception and this class has no business handling it but it may be handled somewhere else.

pbrisbin: `Ctrl-C` and similar wouldn't be handled by `rescue => ex` because they don't descend from `StandardError`, specifically for that reason.

Because there is a UX constraint that this object should be best-effort and can never cause the CLI itself to error, I believe you will need a `rescue => ex` in `#check`. If it makes you feel better to still rescue the parse error or http error internally, that's fine with me; I would just find it to be unnecessary clutter at that point.

(Sorry I didn't clarify the UX constraint in my original comment, I only just now realized that are hands are a bit tied in that way.)

pointlessone: Updated.
I left this one here because we can gracefully fallback here. And with upcoming cache we can even use real (albeit slightly outdated) values.

lib/cc/cli/runner.rb
```diff
@@ -56,6 +59,18 @@ def command
           command_name
         end
       end
+
+      private
+
+      def check_version?
+        first_arg = ARGV.first
+        if %w[--check-version].include? first_arg
```
pbrisbin: Just curious, is there a reason for the array inclusion instead of `if ARGV.first == "--check-version"`?

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,69 @@
...
+        api_response["latest"]
+      end
+
+      def print_new_version_message
+        warn "A new version (v#{latest_version}) is available"
+      end
+
+      def api_response
+        @api_response ||=
+          begin
```
pbrisbin: :lipstick: use implicit `begin`

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,69 @@
...
+      def api_response_body
+        if http_response.is_a? Net::HTTPSuccess
+          http_response.body
+        else
+          raise Net::HTTPFatalError.new("HTTP Error", http_response)
+        end
+      end
+
+      def http_response
+        @http_response ||=
```
pbrisbin: Since you're memoizing this as the parsed `api_response` already, I don't think you need to memoize here. I'd prefer you didn't, to cut down on some syntactic noise.

lib/cc/cli/version_checker.rb
```diff
@@ -0,0 +1,69 @@
...
+          begin
+            uri = URI.parse(ENV.fetch("CODECLIMATE_VERSIONS_URL", DEFAULT_VERSIONS_URL))
+            uri.query = { version: version, uid: UUID.new.generate }.to_query
```
pbrisbin: Is `uuid` a required param (or will it be soon)? If not, I'd suggest you remove that from this PR. We're probably going to have to do something different (which we can discuss in #587) so I think it'd be a cleaner eventual diff if this wasn't already present.

lib/cc/cli/runner.rb
pointlessone: At first I wanted to include a short version as well but failed to come up with a good one. Can change but This seem to be a short-lived code so…

lib/cc/cli/version_checker.rb
pointlessone: `http_response` is called twice in `api_response_body`. Momoization saves us one extra request here. I can save the resposnse in local var there if you wish.

lib/cc/cli/version_checker.rb
pointlessone: I'm not sure what you mean. Here `rescue` block returns a value we want to memoize so it doesn't work with method-lever rescue.

lib/cc/cli/version_checker.rb
pbrisbin: Oh I see. Sorry I misread that, maybe that's an indication that the methods could be broken up a bit differently and be easier to follow? I'm not sure, but FWIW I might've written this this way:

```rb
# Not in love with the names, just trying to show division of responsibility...

def remote_version_info
  # only memoize one thing, the final result
  @remote_version_info ||= get_remote_version_info
end

def get_remote_version_info
  # response parsing, not memoized
  response = http_response

  if response.is_a?(...)  
    JSON.parse(response.body)
  else
    raise ...
  end
rescue JSON::ParserError
  { ... }
end

def http_response
  # Net::HTTP stuff, not memoized
end
```

This is really similar to what you have now, it just avoids memoizing multiple intermediate results which can be confusing and hide bugs (in my opinion and experience). This also would address two explicit `begin`s, which I do consider smells.

lib/cc/cli/version_checker.rb
pbrisbin: Sorry I missed this, my comment [here](https://github.com/codeclimate/codeclimate/pull/586#discussion_r104961767) applies here too. If you decide not to go that way then this comment can be ignored.

pbrisbin:
@GordonDiggs should we merge this now, or wait for #587 to merge into it?

GordonDiggs:
@pbrisbin can't hurt to merge this first!

pbrisbin:
@pointlessone headsup I've rebased this branch and will merge on Green.

PR #587, opened 2017-03-08
XDG config and cache

pointlessone:
⚠️ DO NOT MERGE ⚠️ 

It apparently lack some Docker-specific parts.

---

Adds global config for persistent UUID and version checks switch.

Also adds some caching as well to prevent hitting Versions API too often.

All implemented to [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).


pbrisbin:
So, here's the deal :)

Since the CLI runs as a Docker container, anything you put in `~/` will only exist in that container. Nothing would persist run to run. There are some hacks we could do in the wrapper script to mount the user's local `~/` into the container on each run, but that can get complicated (especially with Docker For Mac, which is always finicky).

I would suggest two things:

1. Reduce the things you need persisted as much as possible

   I wouldn't worry about persisting the "don't check version", if users want to opt out they should pass `--no-check-version` on each invocation (and make their own shell `alias`es to "persist" it). That means you really just need to store a timestamp in `cc-version-last-checked-at.txt`.

1. Put that in `/tmp`, which we do [mount through][mount] and make [available in code][code]

[mount]: https://github.com/codeclimate/codeclimate/blob/master/codeclimate-wrapper#L54
[code]: https://github.com/codeclimate/codeclimate/blob/becc31c1c8292fd214d0b04b79cfd82117296269/lib/cc/analyzer/mounted_path.rb#L19

pbrisbin:
In my first suggestion, I forgot about storing the UUID so we can identify repeated checks from the same version. I'll have to think about this some more, since `/tmp` doesn't really work. Not really sure yet how to mean the product requirements there, given the docker constraints. /cc @GordonDiggs

maxjacobson:
Are we able to move some of that logic into https://github.com/codeclimate/codeclimate/blob/master/codeclimate-wrapper which runs on the host and has access to the full filesystem? It could read the file and pass it into the container. We can't assume that everyone uses the wrapper, but I think most people do. Just a thought 😄

GordonDiggs:
Is there some kind of machine identifier we could hash up and use? Looks like `docker info` has an ID that may be unique per docker engine, which may be good enough

pointlessone:
Since CC CLI can be ran both in container and out of it, we have to make sure both cases work identically.

> Is there some kind of machine identifier we could hash up and use? Looks like docker info has an ID that may be unique per docker engine, which may be good enough

UUID is a candidate. It's build of machine's MAC address (which is supposed to be unique, but might be spoofed), time and some other data for extra uniqueness.

ID from `docker info` is likely to be unique but I don't know if docker updates (or anything else) reset it.

A persisted UUID in user's home directory is likely to even survive OS reinstallation as homedir is the first thing people back up (if at all).

> Are we able to move some of that logic into codeclimate-wrapper

Probably not since I believe containerized and bare CLI should behave the same but it definitely will need to do some setup to get local config/cache into container and back.

pbrisbin:
> Since CC CLI can be ran both in container and out of it, we have to make sure both cases work identically.

This is not true. We do not endorse or support users running the CLI outside of docker.

> It's build of machine's MAC address

Can we access a stable machine MAC address, without the added uniqueness of a UUID? That sounds like what we want.

Barring the answer to this question, I see the following options:

- Don't do anything (my favorite option, cutting scope is _always_ an option)
- Identify by IP server-side (may be _close enough_ for requirements, who knows?)
- Use a generated, persisted UUID handled in the wrapper script
  - Pro: no docker complexities
  - Con: doesn't work for folks running `docker run` directly
- Use `docker info`
  - Pro: readily available, works with or without wrapper script
  - Con: changes at unknown times, best case: docker reinstall, medium case: docker upgrade, worst case: docker restarts

We need to take these tradeoffs to the feature-owner and come back with a direction. @GordonDiggs can you facilitate that?

That said, if you move to my suggestion (don't send a `uuid`, store `last-checked-at` in the mounted `tmp`), we could ship this as a meaningful user-facing increment and defer the user-identification until we have that direction.

GordonDiggs:
> changes at unknown times, best case: docker reinstall, medium case: docker upgrade, worst case: docker restarts

@pbrisbin are you sure it changes in these scenarios? The documentation I read yesterday suggested it is unique per docker engine installation

pbrisbin:
@GordonDiggs isn't that the _best case_ I described?

GordonDiggs:
@pbrisbin yes - my question is whether or not you have any evidence it happens in the other cases too

pbrisbin:
@GordonDiggs none. I was _guessing_ it could be as good as the _best_ or as bad as the _worst_ but had no evidence either way.

GordonDiggs:
@pbrisbin Just tried restarting docker and the ID changed, so that's a no-go.

Here's my prioritized list of options:

1. MAC Address
2. UUID if that is somehow possible but MAC address isn't
3. IP Address. This isn't great but may work for now and we can always revisit in the future

pointlessone:
I wonder why MAC address is higher priority than UUID.

Persisted UUID:
* Pros
  * Semantic and de-facto standard way of getting an unique identifier
  * Includes MAC address
* Cons
  * Somewhat more complex in generation
  * Requires wrapper to pass in

MAC address:
* Pros
  * Supposed to be unique
  * Less code than UUID
* Cons
  * Somewhat arbitrary piece of information
  * Can be spoofed
  * Change when hardware changes (sometimes even when default network interface changes, e.g. from Wifi to Ethernet/LTE/etc).
  * Inaccessible from container, still needs to be passed by a wrapper

> We do not endorse or support users running the CLI outside of docker.

OK. Noted.

> Con: doesn't work for folks running docker run directly

Is that a supported way of running CC?

pbrisbin:
> Is that a supported way of running CC?

It's "supported" yes, we document how to do it in the README. But this issue isn't about a user-facing feature, so it's up to us if "wrapper-only" is OK or not.

> Change when hardware changes (sometimes even when default network interface changes, e.g. from Wifi to Ethernet/LTE/etc).

This sounds like deal breaker.

> Inaccessible from container, still needs to be passed by a wrapper

This suprises me. If we can build a UUID from MAC in the container, it's obviously accessing the MAC somehow right? Anyway, doesn't really matter if MAC is out because it changes with network interface changes, IMO.

pointlessone:
> If we can build a UUID from MAC in the container, it's obviously accessing the MAC somehow right?

It does. Container's MAC. Which might be a real mac (if network is bridged) or some fake MAC (which might me the same for all containers on the planet using this particular VM). With UUID it doesn't really matter because there are other sources of entropy providing global uniqueness.

So as long as we persist generated UUID all is good.

> It's "supported" yes, we document how to do it in the README.

Documented Docker command is 7 lines long and not something I would instantly recall before my first coffee. Would adding 2-4 lines to it be a deal breaker?

> But this issue isn't about a user-facing feature, so it's up to us if "wrapper-only" is OK or not.

Is it OK or not? Wrapper definitely improves ergonomics but Docker-only doesn't appear to be impossible either.

pointlessone:
> If we can build a UUID from MAC in the container, it's obviously accessing the MAC somehow right?

It does. Container's MAC. Which might be a real mac (if network is bridged) or some fake MAC (which might me the same for all containers on the planet using this particular VM). With UUID it doesn't really matter because there are other sources of entropy providing global uniqueness.

So as long as we persist generated UUID all is good.

> It's "supported" yes, we document how to do it in the README.

Documented Docker command is 7 lines long and not something I would instantly recall before my first coffee. Would adding 2-4 lines to it be a deal breaker?

> But this issue isn't about a user-facing feature, so it's up to us if "wrapper-only" is OK or not.

Is it OK or not? Wrapper definitely improves ergonomics but Docker-only doesn't appear to be impossible either.

pointlessone:
I've updated the wrapper to properly mount XDG dirs.

Here's how a Docker version would look like:
```
docker run \
    --interactive --rm \
    --env CODECLIMATE_CODE \
    --env CODECLIMATE_TMP \
    --env CODECLIMATE_DEBUG \
    --env CODECLIMATE_VERSIONS_URL \
    --env CONTAINER_MAXIMUM_OUTPUT_BYTES \
    --env CONTAINER_TIMEOUT_SECONDS \
    --env ENGINE_MEMORY_LIMIT_BYTES \
    --env XDG_CONFIG_HOME="/xdg/config" \
    --env XDG_CACHE_HOME="/xdg/cache" \
    --volume "$CODECLIMATE_CODE":/code \
    --volume "$CODECLIMATE_TMP":/tmp/cc \
    --volume ~/.config/codeclimate:/xdg/config \
    --volume ~/.cache/codeclimate:/xdg/cache \
    analyze
```

pbrisbin:
Sorry @pointlessone I believe that is actually a sub-optimal solution. It means the files written by the CLI will be root-owned. I think it's surprising (and slightly alarming) from an end-user perspective to have a tool writing root-owned files into my user dotfiles.

@GordonDiggs WDYT? Does that matter / is it important? Given that MAC address is out, your next preference was for UUID, but you didn't clarify if it was OK to be a "wrapper-only" thing.

GordonDiggs:
I'd be ok with having a wrapper-only solution for now and then being able to see in the analytics events if there's a high percentage of uid's missing (which would correlate to people not using the wrapper)

pointlessone:
@pbrisbin Indeed. Do you think creating empty files before mounting would be acceptable? CLI never deletes those files. Only creates/writes to them. Once those files are there their permissions and ownership shouldn't change.

pbrisbin:
@pointlessone that's pretty good hack actually! I think I'm good with that.

codeclimate-wrapper
```diff
@@ -41,17 +41,34 @@ else
     && invalid_docker_host "$DOCKER_HOST"
 fi
 
+CC_XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-~/.config/}
```
pbrisbin: We shouldn't make up `CC_` versions of the XDG variables. We use them directly and default them if missing.

```sh
: "${XDG_CONFIG_HOME:=$HOME/.config}"
: "${XDG_CACHE_HOME:=$HOME/.cache}"
```

codeclimate-wrapper
```diff
@@ -41,17 +41,34 @@ else
...
+CC_XDG_CONFIG="${CC_XDG_CONFIG_HOME}/codeclimate/config.yml"
```
pbrisbin: These variables aren't themselves related to XDG: it's our config file, it could live anywhere, and we're nice enough to follow XDG in our default paths. Also, why not let the users set alternate values if they want?

```sh
: "${CC_CONFIG:=$XDG_CONFIG_HOME/codeclimate/config.yml}"
: "${CC_CACHE:=$XDG_CACHE_HOME/codeclimate/cache.yml}"
```

codeclimate-wrapper
```diff
@@ -41,17 +41,34 @@ else
...
+CC_XDG_CONFIG_DIR=$(dirname "${CC_XDG_CONFIG}")
+[ -d "$CC_XDG_CONFIG_DIR" ] || mkdir -p "$CC_XDG_CONFIG_DIR"
+[ -f "$CC_XDG_CONFIG" ] || ( echo "---" > "$CC_XDG_CONFIG" )
```
pbrisbin: These actions are idempotent (if you replace `echo` with `touch`), so I would just always do it. You can also DRY it up with below via a loop:

```sh
for file in "$CC_CONFIG" "$CC_CACHE"; do
  mkdir -p "$(dirname "$file")"
  touch "$file"
done
```

codeclimate-wrapper
```diff
@@ -41,17 +41,34 @@ else
...
+
+CC_XDG_CACHE_HOME=${XDG_CACHE_HOME:-~/.cache/}
+CC_XDG_CACHE="${CC_XDG_CACHE_HOME}/codeclimate/cache.yml"
+CC_XDG_CACHE_DIR=$(dirname "${CC_XDG_CACHE}")
+[ -d "$CC_XDG_CACHE_DIR" ] || mkdir -p "$CC_XDG_CACHE_DIR"
+[ -f "$CC_XDG_CACHE" ] || ( echo "---" > "$CC_XDG_CACHE" )
+
 docker_run() {
-  exec docker run \
+  docker run \
```
pbrisbin: What this intentional? It drastically changes the process tree semantics, signal handling, etc.

codeclimate-wrapper
```diff
@@ -41,17 +41,34 @@ else
...
+    --env CODECLIMATE_VERSIONS_URL \
     --env CONTAINER_MAXIMUM_OUTPUT_BYTES \
     --env CONTAINER_TIMEOUT_SECONDS \
     --env ENGINE_MEMORY_LIMIT_BYTES \
+    --env XDG_CONFIG_HOME="/xdg/config" \
+    --env XDG_CACHE_HOME="/xdg/cache" \
     --volume "$CODECLIMATE_CODE":/code \
     --volume "$CODECLIMATE_TMP":/tmp/cc \
+    --volume "${CC_XDG_CONFIG}":/xdg/config/codeclimate/config.yml \
+    --volume "${CC_XDG_CACHE}":/xdg/cache/codeclimate/cache.yml \
```
pbrisbin: :lipstick: we don't use `${ }` unless it's needed (it's not needed here).

codeclimate-wrapper
```diff
@@ -41,17 +41,34 @@ else
...
     --interactive --rm \
     --env CODECLIMATE_CODE \
     --env CODECLIMATE_TMP \
     --env CODECLIMATE_DEBUG \
+    --env CODECLIMATE_VERSIONS_URL \
     --env CONTAINER_MAXIMUM_OUTPUT_BYTES \
     --env CONTAINER_TIMEOUT_SECONDS \
     --env ENGINE_MEMORY_LIMIT_BYTES \
+    --env XDG_CONFIG_HOME="/xdg/config" \
+    --env XDG_CACHE_HOME="/xdg/cache" \
```
pbrisbin: My personal preference would be not to worry about XDG within the container. We're respecting it in the wrapper for the user (and huge :+1: to that!), but inside the container I'd be inclined to just look for something like a hard-coded `/config.yml` and `/cache.yml` for simplicity. WDYT?

lib/cc/cli/runner.rb
```diff
@@ -18,7 +18,7 @@ def initialize(args)
       end
 
       def run
-        VersionChecker.new.check if check_version?
+        VersionChecker.new.check
```
pbrisbin: We still want this, to support `--no-check-version`. We're just going to flip the default to `--check-version` now that the cache is in place.

lib/cc/cli/xdg_file.rb
```diff
@@ -0,0 +1,59 @@
+require "fileutils"
+require "yaml"
+
+module CC
+  module CLI
+    class XDGFile
```
pbrisbin: This class (and some of the others) seem to be mixing two very distinct concerns:

- Where is the file (XDG, etc)?
- How do I read/write a YAML file (which I'm surprised we have to deal with so directly)

I had recommended in another comment not worrying about XDG or namespaces _inside_ the container. We can just look for `/config.yml` and `/cache.yml`, defined as hard-coded constants (similar to how we handle `/code`). If you did that, I think a ton of this code goes away.

codeclimate-wrapper
pbrisbin: Allowing the override does introduce and error case if users set a relative path. Something to consider.

codeclimate-wrapper
pointlessone: It's a leftover of an intentional change. I started with a directory having only XDG files in it that was mounted into a container and used `trap EXIT` to clean up that dir. With exec script process got replaced by `exec` and there was no one to actually clean up.

I can change it back.

Is there any particular reason why we want docker to become the main process?

pbrisbin: > Is there any particular reason why we want docker to become the main process?

We don't necessarily rely on it now, but it's less surprising and error-prone IMO if the `docker` process sees any signals directly.

Clearly that was preventing _intentional_ intercept logic you had at one time, so the change would've been OK if we needed that, but given no other trade-offs I'd keep the direct signal handling via `exec` as the default behavior still.

lib/cc/cli/file_store.rb
```diff
@@ -0,0 +1,45 @@
...
+      # FILE_NAME is the name of the file this class wraps.
+
+      def initialize
+        load_data
+      end
+
+      def save
+        dir = File.dirname(self.class::FILE_NAME)
+        unless Dir.exist? dir
+          FileUtils.mkdir_p dir
```
pbrisbin: We shouldn't do this. If the files don't exist we should not attempt any version checking; and I wouldn't even have the code to do it, since it's a risk to write root-owned data.

codeclimate-wrapper
```diff
@@ -41,17 +41,31 @@ else
     && invalid_docker_host "$DOCKER_HOST"
 fi
 
+: ${XDG_CONFIG_HOME:=$HOME/.config/}
+CC_CONFIG=$(realpath "${CODECLIMATE_CONFIG:-$XDG_CONFIG_HOME/codeclimate/config.yml}")
```
pbrisbin: What's the portability of `realpath`? We claim the wrapper should work on any POSIX system.

lib/cc/cli/global_config.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def uuid
+        data["uuid"] ||= UUID.new.generate
+      end
+
+      private
+
+      def load_data
+        @data = DEFAULT_CONFIG.merge(super)
```
pbrisbin: Should this be `||=`?

spec/cc/cli/file_store_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "spec_helper"
+
+RSpec.describe CC::CLI::FileStore do
```
pbrisbin: :lipstick: Please drop the `RSpec.` part

codeclimate-wrapper
pointlessone: `realpath` is defined as a function in POSIX. Also present as a CLI tool in GNU coreutils. It's main use case is expansion of symlinks but it also gives a full path for a relative paths.

But on a second though we don't need it here. Docker mounts relative paths just fine.

lib/cc/cli/global_config.rb
pointlessone: No, it shouldn't. We still want the `check-version` in even if it's missing from the config file. `super` sets this ivar so we need to unconditionally reassign it.

codeclimate-wrapper
pbrisbin: > Docker mounts relative paths just fine.

Ah, that's news to be, but cool.

lib/cc/cli/global_config.rb
pbrisbin: I see; that's pretty confusing, but we don't have to hold up the PR further about it.

pbrisbin:
@pointlessone I've merged the PR on which this was based. Please let me know if you're finished here, then I can take over to clean up history and merge.

pointlessone:
I think it's all good. Just squash corresponding fixup commits and it'll be fine.

pbrisbin:
Re-opened as #591

PR #591, opened 2017-03-14
Support config and cache file

pbrisbin:
If using codeclimate-wrapper, manage config.yml and cache.yml files in the
user's home directory (following the XDG spec). These files are currently used
to manage the recently introduced version check feature.

Just re-opening #587 against master.

### DIFF
--- a/lib/cc/cli/version_checker.rb
+++ b/lib/cc/cli/version_checker.rb
@@ -1,3 +1,6 @@
+require "cc/cli/global_config"
+require "cc/cli/global_cache"
+
 module CC
   module CLI
     class VersionChecker
@@ -7,6 +10,8 @@ module CC
       DEFAULT_VERSIONS_URL = "https://versions.codeclimate.com".freeze
 
       def check
+        return unless global_config.check_version?
+
         print_new_version_message if outdated?
       rescue => error
         CLI.debug(error)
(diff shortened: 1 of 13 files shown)


pbrisbin:
Will merge and release tomorrow. Confirmed locally this doesn't break anything, but there is a JSON parse error visible in DEBUG, so it's just skipping the check at the moment.

/cc @pointlessone

pointlessone:
JSON? Is Verssions API broken?

pbrisbin:
Must've been temporary, it's working now.

Only thing left here now is the test failures, I guess.

pbrisbin:
I don't see a persistent UUID written anywhere though,

```
% cat ~/.{config,cache}/codeclimate/*.yml
---
latest-version: 0.52.0
outdated: false
last-version-check: 2017-03-15 14:03:52.801191019 +00:00
```

Is that expected?

pointlessone:
No, the config doesn't appear to be saved.

pbrisbin:
@pointlessone it appears to me nothing ever calls `GlobalCache#save`. The other file is saves as part of all the write methods.

lib/cc/cli/version_checker.rb
```diff
@@ -7,19 +10,33 @@ class VersionChecker
       DEFAULT_VERSIONS_URL = "https://versions.codeclimate.com".freeze
 
       def check
+        return unless global_config.check_version?
```
pointlessone: We need to add `global_config.save` after this line.

pointlessone: Sorry, after the next one.

pointlessone:
@pbrisbin Fixed specs and added config persistence. Separate commit for ease of review. Safe to squash before merging.

pbrisbin:
Thanks! Will try to do a release today.

PR #596, opened 2017-03-16
Fix error when invoking CLI without command

maxjacobson:


maxjacobson:
@pbrisbin updated - does that fix look OK? (context in commit message)

dblandin:
@maxjacobson What do you think of resetting the resolver stubs after each spec?

```ruby
RSpec.configure do |config|
  config.after do
    ::Resolv::DefaultResolver.replace_resolvers([])
  end
end
```

spec/support/resolv_helpers.rb
```diff
@@ -1,6 +1,6 @@
 module ResolvHelpers
   def stub_resolv(name, address)
-    dns = double
+    dns = double("dns")
```
maxjacobson: oo sure. happy to update. This change was helpful for confirming that this was the double that was leaking, btw (just mentioning in case it wasn't clear why this change was included)

PR #601, opened 2017-03-23
Mount `/code` as `:rw` if `CLI.debug?`.

zenspider:
This allows you to run with extra tooling enabled that might leave
file artifacts that are later used to analyze or report (like
stackprof and other profilers).

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -68,6 +68,8 @@ module CC
       end
 
       def container_options
+        mount = CLI.debug? ? "rw" : "ro"
+
         [
           "--cap-drop", "all",
           "--label", "com.codeclimate.label=#{@label}",
@@ -75,7 +77,7 @@ module CC
           "--memory-swap", "-1",
           "--net", "none",
           "--rm",
-          "--volume", "#{@code_path}:/code:ro",
+          "--volume", "#{@code_path}:/code:#{mount}",
           "--volume", "#{config_file.host_path}:/config.json:ro",
           "--user", "9000:9000"
         ]


dblandin:
I think this is great! I wonder though if the working directory is the best place to allow written artifacts. Keeping the working directory read-only is one way that we ensure engines behave properly. Would volume-mounting the configured tmp directory work for this case?

```ruby
# if CLI.debug?
 "--volume", "#{MountedPath.tmp.host_path}:#{MountedPath.tmp.join(SecureRandom.uuid)container_path}",
```

Then, within an engine, you can write file artifacts to `/tmp/cc`.

wfleming:
So far, the primary intent I've seen behind changing this seem to be the profiling, is that accurate? If that's the case, would it be better to change the perms based on the `PROFILE` setting instead of debug?

zenspider:
@wfleming That's my current reason, yes... But I see this as useful in all sorts of debugging tasks. That's why I toggled it on debug.

@dblandin "Keeping the working directory read-only is one way that we ensure engines behave properly"

Not sure why that's a concern when it is only in debug mode. It is still obviously enforced when it isn't enabled.

wfleming:
> That's my current reason, yes... But I see this as useful in all sorts of debugging tasks. That's why I toggled it on debug.

This is kinda YAGNI IMHO. If the current reason is for profiling, we should turn it on when doing profiling. If you're debugging normally, you don't necessarily want this on (and, in fact, having it on may be counterproductive since now you're "debugging" in an environment different from "production"). And, from the other side, doing profiling work does not mean you want to see a ton of debug output (and in fact may be counterproductive since the CLI may spend a bunch of time in I/O for debug output it wouldn't normally.)

All of that having been said, Devon's suggestion also sounds attractive to me. If it were reasonable to configure the profiling code you have in mind to write to `/tmp`, writing to a different mounted directory seems a little bit preferable long term since it doesn't mean special-casing how we treat the code-under-analysis.

dblandin:
Going to go ahead and close this PR for now.

PR #602, opened 2017-03-24
Assign minor severity by default

ABaldwinHunter:
According to the Code Climate
[SPEC](https://github.com/codeclimate/spec/blob/master/SPEC.md#issues)
today, issue severities are optional.

The SPEC was recently updated to include a list of five severity types:
info, minor, major, critical, blocker.

When an engine emits no severity, by default we'll report a severity of "minor".

### DIFF
--- a/spec/cc/analyzer/issue_spec.rb
+++ b/spec/cc/analyzer/issue_spec.rb
@@ -52,10 +52,24 @@ module CC::Analyzer
       it "merges in defaulted attributes" do
         expected_additions = {
           "fingerprint" => "433fae1189b03bcd9153dc8dce209fa5",
+          "severity" => Issue::DEFAULT_SEVERITY,
         }
         issue = Issue.new(output.to_json)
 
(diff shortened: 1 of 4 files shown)


spec/cc/analyzer/issue_spec.rb
```diff
@@ -52,11 +52,25 @@ module CC::Analyzer
...
         expect(issue.as_json).to eq(output.merge(expected_additions))
       end
+
+      it "doesn't overwrite defaulted attrs when present" do
+        optional_attrs = {
+          "fingerprint" => "433fae1189b03bcd9153dc8dce209fa5",
+          "severity" => "major",
+        }
+
+        unchanged  = output.merge(optional_attrs)
```
wfleming: 💄 you have two spaces before the `=` here.

ABaldwinHunter: nice catch! thanks

PR #607, opened 2017-03-26
Validate severity on issues

ABaldwinHunter:
Although severity continues to be an optional attribute for an engine to
emit, we now default to MINOR severity, and only permit 6 options: info,
minor, major, critical, blocker and normal (deprecated).

This validation enforces that restriction to ensure compliance with the
[CC
SPEC](https://github.com/codeclimate/spec/blob/master/SPEC.md#issues).

### DIFF
--- a/lib/cc/analyzer/issue_validations/severity_validation.rb
+++ b/lib/cc/analyzer/issue_validations/severity_validation.rb
@@ -1,0 +1,39 @@
+module CC
+  module Analyzer
+    module IssueValidations
+      class SeverityValidation < Validation
+        INFO = "info".freeze
+        MINOR = "minor".freeze
+        MAJOR = "major".freeze
+        CRITICAL = "critical".freeze
+        BLOCKER = "blocker".freeze
+
+        DEPRECATED_SEVERITIES = [
+          NORMAL = "normal".freeze,
+        ].freeze
+
+        VALID_SEVERITIES = ([
+          INFO,
+          MINOR,
+          MAJOR,
+          CRITICAL,
+          BLOCKER,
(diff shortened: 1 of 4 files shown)


lib/cc/analyzer/issue_validations/severity_validation.rb
```diff
@@ -0,0 +1,40 @@
...
+        VALID_SEVERITIES = ([
+          INFO,
+          MINOR,
+          MAJOR,
+          CRITICAL,
+          BLOCKER,
+        ] + DEPRECATED_SEVERITIES).freeze
+
+        def valid?
+          severity.nil? || VALID_SEVERITIES.include?(severity)
```
ABaldwinHunter: I considered checking for `.present?` here, but I don't think we need or want that. The overwrite to default `'minor` only checks for `nil` [here](https://github.com/codeclimate/codeclimate/commit/0938c70ab78b9075966bee3ae006265dff8aa5fc#diff-e013abc9492360e8274e92f76b9c6071R52). Open to other ideas though.

maxjacobson:
I'm curious if you've checked in the database whether any engines are already emitting issues that wouldn't pass this validation, and if so, if those engines would start errorring when this change is shipped.

ABaldwinHunter:
@maxjacobson fair question! the only engines of which I'm aware that emit `severity` are Brakeman and Bundler Audit: normal, info, critical.

A database query for distinct severity values on the relevant collection is unindexed and prohibitively expensive. Looking at https://github.com/codeclimate-community/submissions/blob/master/manifest.yml I don't believe there are any other engines about which to be concerned. If that's inaccurate we're most likely better off finding out and addressing - since those severities will not be supported.

PR #614, opened 2017-03-30
Replace cc checks machinery with a new prepare step

pbrisbin:
Rather than reading an ENV var and modifying the config in minor ways
shotgun-style as part of interacting with the registry, create an entirely
separate sub-command that re-writes .codeclimate.yml wholesale, such that it
runs our maintainability checks.

The benefits here are:

- Removes the shot-gun changes that we didn't like at the time
- Isolates this new feature WIP to a separate build step, that we can run or
  not (on hosted) based on rollout
- Makes a meaningful increment towards server-side configuration (ping me if
  you want details on this)

In general, this replaces something we knew was a temporary hack, with
something that's in the general direction of where we want to go.


lib/cc/cli/prepare/quality.rb
```diff
@@ -0,0 +1,59 @@
...
+    class Prepare
+      class Quality < Command
+        ENGINES_CONFIG = {
+          "complexity-ruby" => {
+            "enabled" => true,
+            "channel" => "beta",
+          },
+          "duplication" => {
+            "enabled" => true,
+            "channel" => "cronopio",
```
ABaldwinHunter: could this be extracted into a constant?

pbrisbin: Anything is possible!

Can you explain why you'd prefer that?

lib/cc/cli/prepare/quality.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        def modify(content)
+          content.delete("ratings")
+          content["engines"] ||= {}
+          content["engines"].merge!(ENGINES_CONFIG)
+          content
+        end
+
+        def read_codeclimate_yml
+          SafeYAML.load_file(CODECLIMATE_YAML)
```
ABaldwinHunter: any reason we're using safe_yaml over cc::yaml? simpler?

spec/cc/cli/config_generator_spec.rb
```diff
@@ -105,26 +105,6 @@ module CC::CLI
...
-          and_return("false")
-
-        expect(generator.codeclimate_checks.keys).to be_empty
-      end
-
-      it "returns an empty hash when the feature is disabled" do
-        expect(generator.codeclimate_checks.keys).to be_empty
-      end
-    end
-
```
ABaldwinHunter: 👏  enjoying the red

spec/cc/cli/prepare/quality_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+describe CC::CLI::Prepare::Quality do
+  include FileSystemHelpers
+  include ProcHelpers
+
+  around do |spec|
+    within_temp_dir { spec.run }
+  end
+
+  # Avoid changing to /code when executing
+  before { allow(Dir).to receive(:chdir).and_yield }
```
ABaldwinHunter: nice rspec `and_yield`. TIL :)

lib/cc/cli/prepare/quality.rb
pbrisbin: Yeah, in this case I only need to know that it's valid as YAML to work with it in the way I need to, all the cc-yaml validation machinery will still happen downstream (as desired).

As this feature grows, we may start to move some functionality around between here and cc-yaml, but I'm deferring on that for now.

lib/cc/cli/prepare/quality.rb
```diff
@@ -0,0 +1,59 @@
+require "safe_yaml/load"
+
+module CC
+  module CLI
+    class Prepare
+      class Quality < Command
```
chrishulton: so will this only execute if the cc yaml looks something like
```
prepare:
  quality: true
```
?  is that something to cover in a spec?

lib/cc/cli/prepare/quality.rb
ABaldwinHunter: Since `cronopio` was intentionally random, branch name will probably change in the future. Using a constant ref avoids us having to change the channel name here, and provides an opportunity to use a clearer name - like `QUALITY_CHANNEL` or `DUPLICATION_QUALITY_CHANNEL`?

Don't feel strongly though. up to you!

lib/cc/cli/prepare/quality.rb
```diff
@@ -0,0 +1,59 @@
...
+            write_codeclimate_yml(content)
+          end
+        end
+
+        private
+
+        def modify(content)
+          content.delete("ratings")
+          content["engines"] ||= {}
+          content["engines"].merge!(ENGINES_CONFIG)
```
ABaldwinHunter: I think this is fine. One thing to note that we might be losing for the Duplication engine is some additional configuration, like engine-specific exclude paths and a reduced language set.

lib/cc/cli/prepare/quality.rb
```diff
@@ -0,0 +1,59 @@
...
+                { "engines" => ENGINES_CONFIG }
+              end
+
+            write_codeclimate_yml(content)
+          end
+        end
+
+        private
+
+        def modify(content)
```
chrishulton: super minor, is `modify!(content)` more appropriate?

lib/cc/cli/prepare/quality.rb
pbrisbin: Yup, that's intentional for now, but we can revisit later.

lib/cc/cli/prepare/quality.rb
pbrisbin: This'll only execute if `codeclimate prepare:quality` is run. We'll run it (guarded by rollout) on hosted.

lib/cc/cli/prepare/quality.rb
pbrisbin: Seems reasonable.

pbrisbin: Actually, going to not do that for now. From the out-side the method is meant to take and return something. The fact this it uses mutation is kind of a quirk of the current set of changes we need to make right now. We can swing one way or the other in the future, but I'll keep the name more aligned with how it's used (non-mutating).

lib/cc/cli/prepare/quality.rb
```diff
@@ -0,0 +1,64 @@
...
+          "complexity-ruby" => {
+            "enabled" => true,
+            "channel" => "beta",
+          },
+          "duplication" => {
+            "enabled" => true,
+            "channel" => "cronopio",
+            "config" => {
+              "languages" => [
+                "ruby",
```
pbrisbin: There are a number of answers about what to do in this `languages` list. I'm just deferring entirely for now because it's a 2 second change to do something different later.

PR #625, opened 2017-04-12
Direct users to README for upgrade instructions

maxjacobson:
I think more people are going to upgrade if we tell them how.

### DIFF
--- a/lib/cc/cli/version_checker.rb
+++ b/lib/cc/cli/version_checker.rb
@@ -34,7 +34,7 @@ module CC
       end
 
       def print_new_version_message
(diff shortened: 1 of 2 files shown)


lib/cc/cli/version_checker.rb
```diff
@@ -42,7 +42,7 @@ def latest_version
       end
 
       def print_new_version_message
-        warn "A new version (v#{latest_version}) is available"
+        warn "A new version (v#{latest_version}) is available. Upgrade installations are available at: https://github.com/codeclimate/codeclimate#packages"
```
pbrisbin: Is this meant to be "Upgrade instructions..." ?

maxjacobson: @pbrisbin Yuuup. Updated. I think I must have heard someone say the word "installation" while I was typing that.

Issue #619, opened 2017-04-10
Outdated warning false positive


maxjacobson:
I'm seeing some unexpected behavior.

After coming back from vacation, I ran `codeclimate version` and was warned that I was out of date:

```console
$ codeclimate version
WARNING: A new version (v0.59.0) is available
0.58.0
```

So I upgraded my CLi (side note: IMO we should include upgrade instructions in this warning). Afterward, I tried again:

```console
$ codeclimate version
WARNING: A new version (v0.59.0) is available
0.59.1
```

Hm, weird, I am up-to-date, but it's telling me I'm not.

There are a few things going on here:

1. We didn't make a GitHub "release" for 0.59.1, so <https://versions.codeclimate.com/> thinks the latest version is 0.59 -- I think we should update that service to check tags instead of releases
1. the "outdated?" check is done server side and then cached locally. My `~/.cache/codeclimate/cache.yml` currently looks like:
    ```yaml
    ---
    latest-version: 0.59.0
    outdated: true
    last-version-check: 2017-04-10 16:42:47.691495101 +00:00
    ```

   So, even though my local version is 0.59.1, and the cached latest version is 0.59.0, we think we're outdated because that's what the cache says. In the common use-case (which I just followed) where I was notified that I was out of date, then upgraded, users will end up in this state.

cc @GordonDiggs

GordonDiggs:
@maxjacobson ah it's totally because we didn't make a tag - can you make one?

maxjacobson:
I can create a release for 0.59.1, which should resolve this issue for now, but I think there's some design changes that would help prevent it in the future (outlined in the issue). Curious if you have any thoughts on those.

Do you want me to create a new tagged release for the latest change on master (adding the grep engine) which isn't part of 0.59.1 as well?

PR #626, opened 2017-04-12
Don't warn about a new version after user upgrades

maxjacobson:
With this change, the server doesn't need to calculate the outdated
field anymore, but I think we should still leave it for people using an
older version of the CLI.

Fix https://github.com/codeclimate/codeclimate/issues/619

### DIFF
--- a/lib/cc/cli/global_cache.rb
+++ b/lib/cc/cli/global_cache.rb
@@ -31,17 +31,6 @@ module CC
         save
         value
       end
-
-      def outdated
-        data["outdated"] == true
-      end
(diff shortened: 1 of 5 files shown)


maxjacobson:
@GordonDiggs @filipesperandio  yesterday, after seeing the warning, upgrading, and then still seeing the warning I made this PR. WDYT of the approach of moving the checking of `outdated?` from the server to the client? I think the responsibility makes more sense here personally.

GordonDiggs:
Talked IRL - we want the server to do the outdated check by design. Closing

PR #627, opened 2017-04-13
Don't use the cached version if we are not checking for a new version

GordonDiggs:
Issue #619

Rather than using the cached version when we're not hitting the API, we
should skip the outdated check entirely.

### DIFF
--- a/spec/cc/cli/version_checker_spec.rb
+++ b/spec/cc/cli/version_checker_spec.rb
@@ -79,7 +79,7 @@ describe CC::CLI::VersionChecker do
     expect(stderr).to eq ""
   end
 
-  it "uses cached values when API is unavailable" do
+  it "does nothing when API is unavailable" do
     stub_resolv("versions.codeclimate.com", "255.255.255.255")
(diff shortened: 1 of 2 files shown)


spec/cc/cli/version_checker_spec.rb
```diff
@@ -79,7 +79,7 @@ def stub_version_request(versions_resp)
     expect(stderr).to eq ""
   end
 
-  it "uses cached values when API is unavailable" do
+  it "uses does nothing when API is unavailable" do
```
maxjacobson: 🤓 `uses does nothing` -> `does nothing`

spec/cc/cli/version_checker_spec.rb
```diff
@@ -91,10 +91,10 @@ def stub_version_request(versions_resp)
       checker.check
     end
 
-    expect(stderr).to include "A new version (v0.1.1) is available"
+    expect(stderr).to eq ""
   end
 
-  it "uses cached values if checked recently" do
+  it "uses does nothing if checked recently" do
```
maxjacobson: 🤓 `uses does nothing` -> `does nothing`

PR #628, opened 2017-04-13
Rewrite CC::CLI::Analyze

pbrisbin:
- Create new objects for configuration and registry logic
- Remove EnginesRunner, EnginesConfigBuilder, etc
- Use a static, default QM configuration

This begins the rewrite of configuration handling. It'll merge into the `quality-model` branch, which we'll pull in for dog-fooding and QA.


lib/cc/analyzer/engine.rb
```diff
@@ -52,11 +51,6 @@ def run(stdout_io, container_listener)
         container.run(container_options).tap do |result|
           CLI.debug("#{qualified_name} engine stderr: #{result.stderr}")
         end
-      rescue Container::ImageRequired
```
pbrisbin: A different exception is thrown by the `Registry` itself, and it has the context there so there's no need to catch and re-raise.

pbrisbin:
There's definitely more work on this track, but I'd like to pause at points where things are at least functional and I can get a deployment out, and avoid monster PRs.

lib/cc/config/default.rb
```diff
@@ -0,0 +1,44 @@
+module CC
+  module Config
+    class Default
+      attr_reader :engines, :exclude_patterns
+
+      def initialize
+        @engines = [structure_engine, duplication_engine]
+        @exclude_patterns = %w[
```
chrishulton: should this be a constant?

lib/cc/cli/analyze.rb
```diff
@@ -20,23 +20,26 @@ def initialize(_args = [])
...
+              run_engine(engine)
+            end
+          end
         end
-      rescue EnginesRunner::NoEnabledEngines
-        fatal("No enabled engines. Add some to your .codeclimate.yml file!")
+
+        formatter.finished
+      ensure
+        formatter.close if formatter.respond_to?(:close)
```
chrishulton: this could also just be defined as a no-op for the formatters that don't need to close, if it'd be better to have a uniform "formatter" interface.

PR #630, opened 2017-04-14
More rewrites to support Builder usage

pbrisbin:


pbrisbin:
Just opening for visibility, will generally merge these on green. We actually have really good tests in this area!

PR #636, opened 2017-04-18
Bring back CLI::Prepare

pbrisbin:
- Make some changes to assume working directory
- Add support for prepare stanza to CC::Config
- Bring back a modified CLI::Prepare command
  - Updated to interact with CC::Config, not CC::Yaml
  - No overall success/failure messaging

Verified as working locally.


lib/cc/config/prepare.rb
```diff
@@ -0,0 +1,68 @@
...
+          NoPrepareNeeded.new
+        end
+      end
+
+      def initialize(fetch:)
+        @fetch = fetch
+      end
+
+      class Fetch
+        def self.from_yaml(data)
```
pbrisbin: I can't decide if each object should own how to get itself from YAML, of it these should be reduced to value objects with the YAML "parsing" left to `Config::YAML`.

wfleming: Based on the mess we got ourselves into with `cc-yaml`, I'm inclined to say the "parse" stage should effectively produce a plain YAML object (maybe with some very light type-checking/coercing), and extracting semantic meaning from that YAML is a separate stage. Which it looks to me like what you have here, so I think this is a good direction.

wfleming: Although having said that I'm wondering how we'll end up structuring "validation" of config to give user's reasonable errors.

pbrisbin: My current approach is to write all of this consuming code with the assumption that the config has been validated. We'll maintain `codeclimate validate-config` (and run that on as a step on hosted still); it will go through `cc-yaml` *for now*, but be encapsulated in a class that we can rewrite later (see #638)

lib/cc/config/prepare.rb
```diff
@@ -0,0 +1,68 @@
...
+          def self.from_yaml(data)
+            case data
+            when String then new(data)
+            when Hash then new(data.fetch("url"), data["path"])
+            end
+          end
+
+          def initialize(url, path = nil)
+            @url = url
+            @path = path || url.split("/").last
```
wfleming: We did more strenuous validation of the path being safe for security reasons in the old impl. I think we want to keep that.

https://github.com/codeclimate/codeclimate-yaml/blob/master/lib/cc/yaml/nodes/fetch.rb#L51

Less critical, but the new impl here also doesn't prevent path ending up as `""` in several ways. I assume that will blow up at `File.write`, but it may be worth validating earlier.

lib/cc/config/prepare.rb
wfleming: Yeah, that's where I figured we'd end up. I was more thinking that we don't want to totally divorce that validate logic from the data munging in these ``CC::Config::*` classes. So maybe they'll raise exceptions or set error flags & the top-level config building runner will catch/use those and `validate-config` just builds the config & echos errors?

lib/cc/config/prepare.rb
pbrisbin: Nice catch. I'd like to do a few things to solve this (only one of which will be in this PR):

- Merge https://github.com/codeclimate/codeclimate/pull/638 first

  This'll ensure `validate-config` would catch this with a nice user-facing error.

- Don't bring `prepare` back into Builder without `validate-config`

  This'll ensure we don't get to this step without having validated already.

- Duplicate that check here, for *layered security*, just with a less user-friendly outcome because we don't need to be super graceful in an unexpected case where `validate-config` was not run.

pbrisbin: Updated

lib/cc/config/prepare.rb
pbrisbin: I could see use ending up there. Another option is to make it so you can only get a built config _through_ a validator. In any event, I'm not boxing us into any corners and am open to ideas as this matures.

PR #638, opened 2017-04-19
Fixes to validate-config

pbrisbin:
UX/UI:

- Exit non-zero if ERRORs are found, so this works as a fail-able Build step
- Don't color whole lines, makes them harder to read
- Don't color success messages, it was also hard to read
- Don't color WARNINGs red, I think this was a bug
- Filter validation issues that are expected in QM-land

Functionality:

- Allow injecting registry (undocumented) so we can run this "normally" via
  Builder instead of having to require and invoke classes ourselves (and fake
  our usual container logging, snapshot-events, etc, etc).

Implementation:

- Encapsulate through a class under CC::Config, so we have only a single place
  to change as we move from CC::Yaml to the new CC::Config-based logic to
  support Quality Model.

### DIFF
--- a/lib/cc/cli/validate_config.rb
+++ b/lib/cc/cli/validate_config.rb
@@ -4,94 +4,62 @@ module CC
   module CLI
     class ValidateConfig < Command
       SHORT_HELP = "Validate your .codeclimate.yml.".freeze
-
-      include CC::Analyzer
-      include CC::Yaml
-
       VALID_CONFIG_MESSAGE = "No errors or warnings found in .codeclimate.yml file.".freeze
 
       def run
         require_codeclimate_yml
-        verify_yaml
+        process_args
+
+        if any_issues?
+          display_issues
+        else
+          puts VALID_CONFIG_MESSAGE
+        end
+
+        exit 1 unless validator.valid?
       end
 
       private
 
-      def verify_yaml
-        if any_issues?
-          display_issues
-        else
-          puts colorize(VALID_CONFIG_MESSAGE, :green)
+      attr_reader :config, :registry, :validator
+
+      def process_args
+        registry_path = EngineRegistry::DEFAULT_MANIFEST_PATH
+        registry_prefix = ""
+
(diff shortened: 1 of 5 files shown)


pbrisbin:
TODO: fix and add specs.

pbrisbin:
Uncovering a thicket, removed the review-request for now.

pbrisbin:
OK, green and ready for review.

spec/cc/cli/validate_config_spec.rb
```diff
@@ -2,145 +2,115 @@
 
 module CC::CLI
   describe ValidateConfig do
```
pbrisbin: This may be best reviewed by just viewing it in the *To* state.

lib/cc/cli/validate_config.rb
```diff
@@ -4,95 +4,63 @@ module CC
...
           display_issues
         else
-          puts colorize(VALID_CONFIG_MESSAGE, :green)
+          puts VALID_CONFIG_MESSAGE
         end
-      end
 
-      def any_issues?
-        parsed_yaml.errors? || parsed_yaml.nested_warnings.any? || parsed_yaml.warnings? || invalid_engines.any?
+        exit 1 unless validator.valid?
```
wfleming: Oh, man, another case where we've just been always exiting 0 no matter what? Good catch to fix.

lib/cc/cli/validate_config.rb
```diff
@@ -4,95 +4,63 @@ module CC
...
       end
 
-      def display_nested_warnings
-        nested_warnings.each do |nested_warning|
-          if nested_warning[0][0]
-            puts colorize("WARNING in #{nested_warning[0][0]}: #{nested_warning[1]}", :red)
-          end
+      def display_issues
+        validator.errors.each do |error|
+          puts "#{colorize("ERROR", :red)}: #{error}"
```
wfleming: I like the change to only coloring the "type".

PR #645, opened 2017-04-26
Add an option to analyze a file from STDIN

pointlessone:
The invocation is simple:
```shell
echo "randome code" | codeclimate analyze lib/useful.rb
```
Key parts:
* This works only with `analyze` subcommand
* A filename must be specified. This is to properly chose engines, apply inclusion/exclusion rules, and let engines pick up the right configs. Only one filename must be specified.
* Pipe source to the commands STDIN.


pointlessone:
@ale7714 Thanks. I'll push a commit for that in this PR.

PR #658, opened 2017-05-15
Ensure Module#remove_method is defined

maxjacobson:
I accidentally updated activesupport in my last CLI release:
https://github.com/codeclimate/codeclimate/pull/657

And this version has a bug. Context:
https://github.com/rails/rails/issues/28918#issuecomment-298084204

This change patches over the bug. I could also downgrade activesupport,
but I think it's likely to happen again via our `bin/prep-release`
script, so this feels more stable to me.

### DIFF
--- a/lib/cc/cli.rb
+++ b/lib/cc/cli.rb
@@ -1,4 +1,5 @@
 require "active_support"
+require "active_support/core_ext/module/remove_method"
 require "active_support/core_ext"
 require "yaml"
 require "cc/analyzer"
--- a/lib/cc/cli.rb
+++ b/lib/cc/cli.rb
@@ -1,5 +1,5 @@
 require "active_support"
-require "active_support/core_ext/module/remove_method"
+require "active_support/core_ext/module/remove_method" # Temporary, see https://github.com/codeclimate/codeclimate/pull/658
 require "active_support/core_ext"
 require "yaml"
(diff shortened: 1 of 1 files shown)


maxjacobson:
Cool, added

wfleming:
> but I think it's likely to happen again via our bin/prep-release

I don't understand what `prep-release`'s flaw is here. If that script can ever change release version accidentally, that's a real problem.

maxjacobson:
Good point. I'll look into that

maxjacobson:
@wfleming here's the minimal reproduction case:

```
git checkout b3b89f18520e1165052bb94dfbae154c2a88ecdc
echo "0.63.4" > VERSION
make Gemfile.lock
```

You wouldn't expect that to update activesupport, but it does.

We can dig into that more, but wdyt so far?

wfleming:
Oh, I think I know what's up. The way bundler works, because the version of `codeclimate` changed, any of its dependencies are also up for grabs if there are new versions in range. If we wanted to be more careful I think we'd want to `prep-release` to pass flags to `bundler` to limit what it will do. Either `bundle update codeclimate` or `bundle update --source=codeclimate` would probably do what we want.

maxjacobson:
I tried those, and a few other things, but it seems to always update activesupport 😞

PR #664, opened 2017-05-22
More-thoroughly validate location, avoid error when inferring fingerprint

maxjacobson:
See the commit messages for more context.

### DIFF
--- a/lib/cc/analyzer/engine_output_filter.rb
+++ b/lib/cc/analyzer/engine_output_filter.rb
@@ -23,6 +23,8 @@ module CC
 
       def ignore_fingerprint?(issue)
         @config.fetch("exclude_fingerprints", []).include?(issue.fingerprint)
+      rescue SourceExtractor::InvalidLocation
+        false
       end
 
       def check_config(check_name)
--- a/lib/cc/analyzer/issue_validations/location_format_validation.rb
+++ b/lib/cc/analyzer/issue_validations/location_format_validation.rb
@@ -32,7 +32,7 @@ module CC
               (
                 [position["line"], position["column"]].all? { |value| value.is_a?(Integer) } ||
                 position["offset"].is_a?(Integer)
-            )
+              )
           end
 
           def valid_lines?(lines)
--- a/lib/cc/analyzer/issue_validations/location_format_validation.rb
+++ b/lib/cc/analyzer/issue_validations/location_format_validation.rb
@@ -2,42 +2,50 @@ module CC
   module Analyzer
     module IssueValidations
       class LocationFormatValidation < Validation
-        def valid?
-          if location["lines"]
-            valid_lines?(location["lines"])
-          elsif location["positions"]
-            valid_positions?(location["positions"])
-          else
-            false
+        class Validator
+          def initialize(location)
+            @location = location
           end
+
+          def valid?
+            if location["lines"]
+              valid_lines?(location["lines"])
+            elsif location["positions"]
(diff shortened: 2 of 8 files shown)


maxjacobson:
(@ABaldwinHunter @chrishulton, this is the exception we were talking about)

lib/cc/analyzer/engine_output_filter.rb
```diff
@@ -22,7 +22,7 @@ def check_disabled?(issue)
       end
 
       def ignore_fingerprint?(issue)
-        @config.fetch("exclude_fingerprints", []).include?(issue.fingerprint)
+        @config.fetch("exclude_fingerprints", []).include?(issue.fingerprint(expect_valid_locations: false))
```
chrishulton: `expect_valid_locations ` seems like kind of a weird argument to me.  since this is the only place where it is false, could you just add a rescue block here for the `SourceExtractor::InvalidLocation` instead?  i guess it could just return false in the rescue and error further downstream?

lib/cc/analyzer/source_extractor.rb
```diff
@@ -17,7 +19,14 @@ def extract(location)
 
       private
 
-      attr_reader :location, :source
+      attr_reader :source
+
+      def validate_location(location)
+        validator = IssueValidations::LocationFormatValidation::Validator.new(location)
+        unless validator.valid?
+          raise InvalidLocation
```
chrishulton: will this exception have information about why its invalid?  is it just `Location is not formatted correctly`?  i think the more detail the better.

lib/cc/analyzer/issue_validations/location_format_validation.rb
```diff
@@ -2,42 +2,50 @@ module CC
...
     module IssueValidations
       class LocationFormatValidation < Validation
-        def valid?
-          if location["lines"]
-            valid_lines?(location["lines"])
-          elsif location["positions"]
-            valid_positions?(location["positions"])
-          else
-            false
+        class Validator
```
ABaldwinHunter: Is this class mostly just a refactor? I don't see a lot of value in it and might avoid muddying the diff if not directly related to the change.

maxjacobson: The reason for extracting the class is so we can use the validator in a context where we have access to the location, but not the issue (like in `SourceExtractor`). It would've been cool to just use the existing validator as-is, but it requires having access to the issue... agreed this feels a little weird

lib/cc/analyzer/source_extractor.rb
maxjacobson: currently it doesn't have any level of detail. could add it, I suppose 🤔

lib/cc/analyzer/engine_output_filter.rb
maxjacobson: that sounds reasonable. The reason it felt a little weird was that the exception is pretty far away from this call site. For this call site to know that that exception might be raised requires it to know that `Issue#fingerprint` eventually, sometimes uses `SourceExtractor`. I guess that's OK, though.. and does make a smaller diff... I think I'm convinced, will update :)

maxjacobson:
@chrishulton updated this - removed the weird argument, and added details about why the location is invalid. The latter is ... a lot of conditionals. WDYT about the approach / the copy?

maxjacobson:
@chrishulton I like that. updated. Mind taking one more look?

PR #665, opened 2017-05-22
Add support for .codeclimate.json config

nporteschaikin:
This also passes the "checks" key from the JSON config to default
engines (complexity-ruby and duplication).

### DIFF
--- a/lib/cc/config/default.rb
+++ b/lib/cc/config/default.rb
@@ -42,6 +42,11 @@ module CC
           "complexity-ruby",
           enabled: true,
           channel: "beta",
+          config: {
+            "config" => {
+              "checks" => json.checks,
+            },
+          },
         )
       end
 
@@ -53,9 +58,14 @@ module CC
           config: {
             "config" => {
               "languages" => %w[ruby],
-            }
+              "checks" => json.checks,
+            },
           },
         )
+      end
+
+      def json
+        @json ||= JSON.new
       end
     end
   end
--- a/lib/cc/config/json.rb
+++ b/lib/cc/config/json.rb
@@ -1,0 +1,27 @@
+module CC
+  module Config
+    class JSON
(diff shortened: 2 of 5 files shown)


nporteschaikin:
One note: I make the deliberate decision not to decide which check configurations to send to the structure engine. I think leaving this determination to the engine saves us from writing this logic in two places.

lib/cc/config/default.rb
```diff
@@ -53,10 +58,15 @@ def duplication_engine
...
               "languages" => %w[ruby],
-            }
+              "checks" => json.checks,
+            },
           },
         )
       end
+
+      def json
+        @json ||= JSON.new
```
pbrisbin: This doesn't make sense to me. `Config::X` is basically a source of configuration:

- `Default` is our standard no-config default values
- `YAML::*` come from `.codeclimate.yml`
- `JSON::*` would come from the new `.codeclimate.json`

In no case would one source reference the either, they'd all be loaded individually and possibly merged by the caller.

In the case of the default configuration, there would be no `checks` key, since the default configuration would run the engine with, well, it's own default configuration.

lib/cc/config/json.rb
```diff
@@ -0,0 +1,27 @@
+module CC
+  module Config
+    class JSON
+      DEFAULT_PATH = ".codeclimate.json".freeze
+
+      def initialize(path = DEFAULT_PATH)
+        @path = path
+      end
+
+      def checks
```
pbrisbin: Any `Config::*` source object must have the same interface -- otherwise they can't be merged and used uniformly. So this will have to have all the same public methods as `Config::Default` and `Config::YAML`, but loaded from the JSON file.

nporteschaikin:
Re-thinking this in a new PR.

PR #666, opened 2017-05-22
Make configurations mergeable

nporteschaikin:


nporteschaikin:
Sending this to you while it's WIP, @pbrisbin, to get your early review.

lib/cc/config.rb
```diff
@@ -5,12 +5,95 @@
...
+
+      def engines
+        Engines.new(left.engines, right.engines).merge
+      end
+
+      def exclude_patterns
+        left.exclude_patterns | right.exclude_patterns
+      end
+
+      # This could also be moved inright the Prepare class itself.
```
pbrisbin: I like that idea, `Prepare#merge`. I also wonder if `Engine#merge` would be better than the `Engines` class you have here. Merging the engines would be a concatenation and calling `left.merge(right)` for any collisions.

lib/cc/config/json.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      private
+
+      attr_reader \
+        :path,
+        :json
+
+      def configured_default_engines
+        default.engines.map do |engine|
```
pbrisbin: Discussed a change to this IRL :)

lib/cc/config/default.rb
```diff
@@ -1,6 +1,6 @@
 module CC
-  module Config
-    class Default
+  class Config
+    class Default < Config
```
nporteschaikin: I think it makes sense as-is; the data _is_ the configuration. I do agree that using the class as a namespace is a bit awkward, but it makes the most semantic sense (and is better than, say, nesting another namespace).

PR #674, opened 2017-05-25
Bundle if necessary in bin/release

maxjacobson:
I recently installed ruby 2.4.1 and hadn't yet installed everything I
need to run this, and it led to a kind of awkward LoadError. This should
make it automatically bundle install when necessary.

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -15,6 +15,7 @@ git pull
 version=$(< VERSION)
 
 printf "RELEASE %s\n" "$version"
+bundle check || bundle install
 rake release
 
 docker build --rm -t codeclimate/codeclimate .


wfleming:
Would you consider making this script run `rake` from inside a docker container instead? That would be in keeping with our approach elsewhere and avoid the need to bundle install locally as well as in the image.

maxjacobson:
Hm interesting. If we mount `~/.gem/credentials` into the container, it might work. I can try that.

maxjacobson:
@wfleming there are a few hurdles here:

- the release process tries to tag the version, which requires access to the current user's git credentials. This is do-able if we expose some environment variables
- it of course also requires access to the `.git` directory, which is currently docker-ignored

were you thinking we'd use a different image than `codeclimate/codeclimate`? I played briefly with using the `codeclimate/alpine-ruby` one but that doesn't have git installed...

I'm leaning toward leaving this PR as-is but curious if you have any additional thoughts

maxjacobson:
:) no worries. I hope it wasn't too passive-aggressive to reach for hermes instead of pinging you again. I like that it picked you again though haha

PR #677, opened 2017-05-30
Partial scan support

pointlessone:
Notable changes:

* Workspace doesn't assume we include all files by default. Now we have to explicitly do that.
* Fixed a bug when removing all files in a directory it became fully included.
* Simplified inclusion/exclusion state tracking. Main features:

  * It's about as performant as the old one
  * It may consume a bit more memory on big code bases
  * It prevents project root escape
  * It may produce a more compact resulting path set

* Added a partial scan mode

### Partial Scan mode

This is a new way to construct a workspace for analysis. It's similar to target file mode when one can specify files for analysis except it takes into account `exclude_paths`.

This is useful in particular for editor integration where one would want to get edited files automatically analyzed but not get reports for excluded files. The results of a partial scan are always a subset of a full scan.

### DIFF
--- a/lib/cc/analyzer/engines_runner.rb
+++ b/lib/cc/analyzer/engines_runner.rb
@@ -5,13 +5,14 @@ module CC
     class EnginesRunner
       NoEnabledEngines = Class.new(StandardError)
 
-      def initialize(registry, formatter, source_dir, config, requested_paths = [], container_label = nil)
+      def initialize(registry, formatter, source_dir, config, requested_paths = [], container_label = nil, partial = false)
         @registry = registry
         @formatter = formatter
         @source_dir = source_dir
         @config = config
         @requested_paths = requested_paths
         @container_label = container_label
+        @partial = partial
       end
 
       def run(container_listener = ContainerListener.new)
@@ -28,7 +29,7 @@ module CC
 
       private
 
-      attr_reader :requested_paths
+      attr_reader :partial, :requested_paths
 
       def build_engine(built_config)
         Engine.new(
@@ -47,6 +48,7 @@ module CC
           container_label: @container_label,
           source_dir: @source_dir,
           requested_paths: requested_paths,
+          partial: partial,
         ).run
       end
 
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -7,6 +7,7 @@ module CC
       SHORT_HELP = "Run analysis with the given arguments".freeze
       HELP = "#{SHORT_HELP}\n" \
         "\n" \
+        "    -p, --partial                    Perform partial analysis." \
         "    -f <format>, --format <format>   Format of output. Possible values: #{CC::Analyzer::Formatters::FORMATTERS.keys.join ", "}\n" \
         "    -e <engine[:channel]>            Engine to run. Can be specified multiple times.\n" \
         "    --dev                            Run in development mode. Engines installed locally that are not in the manifest will be run.\n" \
@@ -18,6 +19,7 @@ module CC
         super
         @engine_options = []
         @path_options = []
+        @partial = false
 
         process_args
         apply_config_options
@@ -27,7 +29,7 @@ module CC
         require_codeclimate_yml
 
         Dir.chdir(MountedPath.code.container_path) do
-          runner = EnginesRunner.new(registry, formatter, source_dir, config, path_options)
+          runner = EnginesRunner.new(registry, formatter, source_dir, config, path_options, nil, partial)
           runner.run
         end
       rescue EnginesRunner::NoEnabledEngines
@@ -37,7 +39,7 @@ module CC
       private
 
       attr_accessor :config
-      attr_reader :engine_options, :path_options
+      attr_reader :engine_options, :path_options, :partial
 
       def process_args
         while (arg = @args.shift)
@@ -48,6 +50,8 @@ module CC
             @engine_options << @args.shift
           when "--dev"
             @dev_mode = true
+          when "-p", "--partial"
+            @partial = true
           else
             @path_options << arg
           end
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -24,16 +24,17 @@ module CC
         :container_label,
       )
 
-      def initialize(registry:, config:, container_label:, source_dir:, requested_paths:)
+      def initialize(registry:, config:, container_label:, source_dir:, requested_paths:, partial:)
         @registry = RegistryAdapter.new(registry)
         @config = config
         @container_label = container_label
-        @requested_paths = requested_paths
+        @requested_paths = Array(requested_paths)
         @source_dir = source_dir
+        @partial = partial
       end
 
       def run
-        names_and_raw_engine_configs.map do |name, raw_engine_config|
+        enabled_engine_configs.map do |name, raw_engine_config|
           label = @container_label || SecureRandom.uuid
           engine_config = engine_config(raw_engine_config)
           engine_metadata = @registry.fetch(name, raw_engine_config.channel)
@@ -55,7 +56,7 @@ module CC
       end
 
       def engine_workspace(raw_engine_config)
-        if raw_engine_config.key?("exclude_paths") && !@requested_paths.present?
+        if apply_excludes? && raw_engine_config.key?("exclude_paths")
           base_workspace.clone.tap do |workspace|
             workspace.remove(raw_engine_config["exclude_paths"])
           end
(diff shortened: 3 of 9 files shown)


lib/cc/workspace/fs_node.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      protected
+
+      attr_reader :path, :children
+
+      def complete?
+        (
+        !directory? ||
+          children.keys.sort == path.children.map { |path| path.basename.to_s }.sort &&
+            children.values.all? { |child| child.complete? }
```
pointlessone: CC suggests using a short version (`all?(&:complete?)`) but this is not equivalent in this case as the method is protected.

lib/cc/analyzer/engines_runner.rb
```diff
@@ -5,13 +5,14 @@ module Analyzer
     class EnginesRunner
       NoEnabledEngines = Class.new(StandardError)
 
-      def initialize(registry, formatter, source_dir, config, requested_paths = [], container_label = nil)
+      def initialize(registry, formatter, source_dir, config, requested_paths = [], container_label = nil, partial = false)
```
pointlessone: `container_label` is not used. I can remove it but chosen to keep it as the change is not related to this PR.

lib/cc/cli/analyze.rb
```diff
@@ -48,6 +50,8 @@ def process_args
             @engine_options << @args.shift
           when "--dev"
             @dev_mode = true
+          when "-p", "--partial"
```
pointlessone: This branch bumps complexity for this method over the threshold. While it's true I don't see how to make it simpler and keep readability.

wfleming:
When I run `spec:benchmark` on `master`, I got:

> `PathTree over 49890 files, with 50 excludes took: 0.27004753300025186s`

Running on this branch, I got:

> `PathTree over 49897 files, with 50 excludes took: 7.1820523669994145s`

While still within that benchmarks's max time of 10s, that's a very significant slowdown, and I find it concerning, so I don't agree that this is nearly as fast as the the previous implementation. Keep in mind that this is the amount of time a CLI user on a large repo might have to wait before any analysis starts running. IMHO, 7s is longer than we should consider acceptable. (The 10s limit was picked out of a hat to catch catastrophic performance regression, but since we've had sub-second performance for so long I'd really like to keep it there.)

If the goal here is to allow "partial scan mode", the changes here seem far larger than required to achieve that. I would actually support making exclude paths apply even with provided paths the *default* behavior, and allowing a flag to change it for cases where the user really does want to analyze excluded files. This is a very quick draft of a patch that would do that (with some important details not addressed):
```
diff --git a/lib/cc/analyzer/engines_config_builder.rb b/lib/cc/analyzer/engines_config_builder.rb
index 7e688e0..53a4b5d 100644
--- a/lib/cc/analyzer/engines_config_builder.rb
+++ b/lib/cc/analyzer/engines_config_builder.rb
@@ -24,12 +24,15 @@ module CC
         :container_label,
       )
 
-      def initialize(registry:, config:, container_label:, source_dir:, requested_paths:)
+      def initialize(registry:, config:, container_label:, source_dir:, requested_paths:, apply_excludes: true)
         @registry = RegistryAdapter.new(registry)
         @config = config
         @container_label = container_label
         @requested_paths = requested_paths
         @source_dir = source_dir
+        @apply_excludes = apply_excludes
+
+        CC::CLI.debug("in EngineConfigBuilder apply_excludes=#{@apply_excludes}")
       end
 
       def run
@@ -55,7 +58,7 @@ module CC
       end
 
       def engine_workspace(raw_engine_config)
-        if raw_engine_config.key?("exclude_paths") && !@requested_paths.present?
+        if raw_engine_config.key?("exclude_paths") && @apply_excludes
           base_workspace.clone.tap do |workspace|
             workspace.remove(raw_engine_config["exclude_paths"])
           end
@@ -76,8 +79,11 @@ module CC
 
       def base_workspace
         @base_workspace ||= Workspace.new.tap do |workspace|
+          CC::CLI.debug("WORKSPACE requested_paths=#{@requested_paths}")
           workspace.add(@requested_paths)
-          unless @requested_paths.present?
+          CC::CLI.debug("WORKSPACE workspace=#{workspace.inspect}")
+          exit 1
+          if @apply_excludes
             workspace.remove([".git"])
             workspace.remove(@config.exclude_paths)
           end
diff --git a/lib/cc/analyzer/engines_runner.rb b/lib/cc/analyzer/engines_runner.rb
index 803cf5c..88a6b7c 100644
--- a/lib/cc/analyzer/engines_runner.rb
+++ b/lib/cc/analyzer/engines_runner.rb
@@ -5,12 +5,13 @@ module CC
     class EnginesRunner
       NoEnabledEngines = Class.new(StandardError)
 
-      def initialize(registry, formatter, source_dir, config, requested_paths = [], container_label = nil)
+      def initialize(registry, formatter, source_dir, config, requested_paths = [], apply_excludes = false, container_label = nil)
         @registry = registry
         @formatter = formatter
         @source_dir = source_dir
         @config = config
         @requested_paths = requested_paths
+        @apply_excludes = apply_excludes
         @container_label = container_label
       end
 
@@ -47,6 +48,7 @@ module CC
           container_label: @container_label,
           source_dir: @source_dir,
           requested_paths: requested_paths,
+          apply_excludes: @apply_excludes,
         ).run
       end
 
diff --git a/lib/cc/cli/analyze.rb b/lib/cc/cli/analyze.rb
index 8e435a0..e0e506f 100644
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -18,6 +18,7 @@ module CC
         super
         @engine_options = []
         @path_options = []
+        @apply_excludes = true
 
         process_args
         apply_config_options
@@ -27,7 +28,7 @@ module CC
         require_codeclimate_yml
 
         Dir.chdir(MountedPath.code.container_path) do
-          runner = EnginesRunner.new(registry, formatter, source_dir, config, path_options)
+          runner = EnginesRunner.new(registry, formatter, source_dir, config, path_options, @apply_excludes)
           runner.run
         end
       rescue EnginesRunner::NoEnabledEngines
@@ -48,6 +49,8 @@ module CC
             @engine_options << @args.shift
           when "--dev"
             @dev_mode = true
+          when "--no-excludes"
+            @apply_excludes = false
           else
             @path_options << arg
           end
diff --git a/lib/cc/workspace/path_tree/dir_node.rb b/lib/cc/workspace/path_tree/dir_node.rb
index 36bef6d..3d33d60 100644
--- a/lib/cc/workspace/path_tree/dir_node.rb
+++ b/lib/cc/workspace/path_tree/dir_node.rb
@@ -14,7 +14,7 @@ module CC
         end
 
         def all_paths
-          if populated?
+          if populated? || @was_expanded
             children.values.flat_map(&:all_paths)
           else
             [File.join(root_path, File::SEPARATOR)]
@@ -39,6 +39,7 @@ module CC
           return if head.nil? && tail.empty?
 
           if (entry = find_direct_child(head))
+            @was_expanded = true
             children[entry.basename.to_s.dup.freeze] ||= PathTree.node_for_pathname(entry)
             children[entry.basename.to_s.dup.freeze].add(*tail)
           else
@@ -52,6 +53,7 @@ module CC
 
         def populate_direct_children
           return if populated?
+          @was_expanded = true
 
           Pathname.new(root_path).each_child do |child_path|
             children[child_path.basename.to_s] = PathTree.node_for_pathname(child_path)
diff --git a/spec/cc/workspace/path_tree_spec.rb b/spec/cc/workspace/path_tree_spec.rb
index 8935233..dba4648 100644
--- a/spec/cc/workspace/path_tree_spec.rb
+++ b/spec/cc/workspace/path_tree_spec.rb
@@ -44,6 +44,17 @@ class CC::Workspace
       end
     end
 
+    it "excludes everything after explicit include" do
+      within_temp_dir do
+        make_fixture_tree
+
+        tree = PathTree.for_path(".")
+        tree.include_paths(["code/a/bar.rb"])
+        tree.exclude_paths(["code/"])
+        expect(tree.all_paths.sort).to eq []
+      end
+    end
+
     it "excludes directory after excluding only part of it" do
       within_temp_dir do
         make_fixture_tree
```

> * It prevents project root escape

Can you provide more details here? The current implementation [does guard against root escape](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/workspace/path_tree.rb#L44). A demonstration of a configuration that would allow root escape would be helpful if this is a problem. I just tried to do the naive thing of writing a spec that added an existing file, and it was not able to escape the project:

```
    it "handles dir escape" do
      within_temp_dir do
        make_fixture_tree

        $stderr.puts "DEBUG pwd=#{Dir.pwd} exist=#{File.exist?("../../linuxrc")}"
        tree = PathTree.for_path(".")
        tree.include_paths(["../../linuxrc"])
        expect(tree.all_paths.sort).to eq ["./"]
      end
    end
```

pointlessone:
@wfleming I redone this from scratch. This time without rewriting `DirNode`/`FileNode`. This time it is as fast as it used to be.

I rewritten FS tree code because I couldn't find the issue with reexpansion. Also default include all and narrow down after include is confusing as well. I wanted to make it easier to understand: give it consistent behaviour (default is include none, include only adds paths).  Your code gave me idea how to implement what I wanted (inclusion state tracking) within existing code. While doing that I found the initial bug that motivated the rewrite: empty dir node get repopulated after _second_ time its child gets removed. See [example](https://github.com/codeclimate/codeclimate/pull/677/files#diff-41f2bbe66a50a63e47806e47979f3840R80).

> The current implementation [does guard against root escape](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/workspace/path_tree.rb#L44).

You're right. Partially. Current implementation does indeed guard against root escape. But it's done not by the linked code but in the [child node detection code](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/workspace/path_tree/dir_node.rb#L62). `Pathname#children` doesn't include `..` so it never can be added to paths. It's all right even if somewhat explicit.

dblandin:
@pointlessone Is this PR still active?

wfleming:
@dblandin I should still get around to reviewing it, and I feel very guilty I haven't, but please leave it open for now.

pointlessone:
@dblandin Yes, this is a desired feature. May need a rebase/minor tweaks but otherwise I'd like to merge it.

brynary:
> 👋 Doing some New Year's cleaning and (with @GordonDiggs's OK) am closing out inactive PRs (more than 30 days w/ no activity), using the handy Pull Requests report in Velocity.
>
> This is not an editorial judgement on the value of the PR, I'm just acting as a human 🤖 . Please feel free to re-open in the future if you'd like to advance this further.

PR #678, opened 2017-05-30
Add ? to supported Exclusion glob symbols

pointlessone:
This gets us to `*` and `?` but still leaves out `[a-z]`, `{a,b}`, and `\` (escapes).

Maybe it'd be easier and more robust to always glob even when there are no special characters?

### DIFF
--- a/lib/cc/workspace/exclusion.rb
+++ b/lib/cc/workspace/exclusion.rb
@@ -15,7 +15,7 @@ module CC
       end
 
       def glob?
-        pattern.include?("*")
+        !(pattern =~ /[*?]/).nil?
       end
 
       def negated?
--- a/spec/cc/workspace/exclusion_spec.rb
+++ b/spec/cc/workspace/exclusion_spec.rb
@@ -14,5 +14,21 @@ class CC::Workspace
       exclusion = Exclusion.new("!foo")
       expect(exclusion.expand).to eq(["foo"])
     end
+
+    describe "#glob?" do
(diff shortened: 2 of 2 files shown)


wfleming:
> This gets us to * and ? but still leaves out [a-z], {a,b}, and \ (escapes).

If we're adding stuff, we may as well also add `[]` and `{}`, yeah?

> Maybe it'd be easier and more robust to always glob even when there are no special characters?

I don't remember exactly why we did this. It was probably a performance optimization (don't read the FS if you don't have to), and it may have been premature. I'd be personally ok with always doing the match as a glob if it doesn't change any behavior we can identify and doesn't move the needle of `spec:benchmark`.

dblandin:
@pointlessone Is this PR still active?

pointlessone:
@dblandin If there's any interest in this functionality I'd like to add other glob capabilities as well. Otherwise feel free to close it.

dblandin:
> I'd be personally ok with always doing the match as a glob if it doesn't change any behavior we can identify and doesn't move the needle of spec:benchmark.

Ditto. Sounds like a good path forward 👍

brynary:
> 👋 Doing some New Year's cleaning and (with @GordonDiggs's OK) am closing out inactive PRs (more than 30 days w/ no activity), using the handy Pull Requests report in Velocity.
>
> This is not an editorial judgement on the value of the PR, I'm just acting as a human 🤖 . Please feel free to re-open in the future if you'd like to advance this further.

PR #687, opened 2017-06-20
Map deprecated severity normal to minor

ABaldwinHunter:
In conformity with Code Climate
[SPEC](https://github.com/codeclimate/spec/blob/master/SPEC.md#issues)


ABaldwinHunter:
Basically we don't want to produce issue data that has `normal` severity any more. 

Note: this should not change the fingerprint because of the way the merge works.

ABaldwinHunter:
Primarily impacts `brakeman` and `bundler-audit`.

PR #689, opened 2017-06-22
Add issue overrides

pointlessone:
Issue override is a way to modify engine output. At the moment there's only one override: severity. It can be any of the values specified in [the spec](https://github.com/codeclimate/spec/blob/master/SPEC.md#issues).

Depends on codeclimate/codeclimate-yaml#41

### DIFF
--- a/lib/cc/analyzer/engine_output_overrider.rb
+++ b/lib/cc/analyzer/engine_output_overrider.rb
@@ -1,0 +1,31 @@
+module CC
+  module Analyzer
+    class EngineOutputOverrider
+      def initialize(config = {})
+        @config = config
+      end
+
+      def apply(output)
+        if output.issue?
+          override_severity(output.as_issue.as_json)
+        else
+          output
+        end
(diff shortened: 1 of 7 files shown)


lib/cc/analyzer/engine_output_overrider.rb
```diff
@@ -0,0 +1,33 @@
+module CC
+  module Analyzer
+    class EngineOutputOverrider
+      ISSUE_TYPE = "issue".freeze
```
pbrisbin: Is this used?

pointlessone:
@pbrisbin Since CC CLI uses cc-yaml for config parsing it would warn and completely ignore the relevant config piece. Effectively disabling the feature.

lib/cc/analyzer/engine_output_overrider.rb
pointlessone: Nope. Copy-paste leftovers.

PR #694, opened 2017-07-07
Support "plugins" in JSON config

wfleming:
This is the new key for specifying plugin engines.

Since we'll shortly be ignoring legacy yaml entirely when JSON is
present, we also have a practical need to be able to configure
structure/duplication in this file (e.g. to set channel), so this will
respect configuration for those engines under the "plugins" key, even
though they are not semantically plugins.

### DIFF
--- a/lib/cc/config/json.rb
+++ b/lib/cc/config/json.rb
@@ -20,7 +20,7 @@ module CC
         :json
 
       def structure_engine
-        base_engine = Engine.new(
+        Engine.new(
           "structure",
           enabled: true,
           channel: channel_overrides.fetch("structure", "stable"),
@@ -33,7 +33,7 @@ module CC
       end
 
       def duplication_engine
-        base_engine = Engine.new(
+        Engine.new(
           "duplication",
           enabled: true,
           channel: channel_overrides.fetch("duplication", "cronopio"),
--- a/lib/cc/config/json.rb
+++ b/lib/cc/config/json.rb
@@ -20,49 +20,47 @@ module CC
         :json
 
       def structure_engine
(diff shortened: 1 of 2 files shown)


wfleming:
Good points both, @pbrisbin @nporteschaikin: updated.

>  but I still wonder whether we need any override logic, 

Pat & I discussed this in Slack, but my feeling is that we at least need to be able to override channel for QM engines so we can test changes to them. Pat suggested a different strategy there, which the latest commit here takes into account.

lib/cc/config/json.rb
```diff
@@ -20,9 +20,10 @@ def initialize(path = DEFAULT_PATH)
         :json
 
       def structure_engine
-        Engine.new(
+        base_engine = Engine.new(
```
nporteschaikin: You can remove the assignment here.

lib/cc/config/json.rb
```diff
@@ -32,10 +33,10 @@ def structure_engine
       end
 
       def duplication_engine
-        Engine.new(
+        base_engine = Engine.new(
```
nporteschaikin: ... and here 😄

wfleming:
> I think a canary to a repo that tests duplication would result in slightly cleaner code and an always-semantic configuration. But this has a very small footprint/isn't the hill I'd die on, so LGTM 😄

I don't think that really works long term: the beta for structure has already been really valuable while rolling through checks, and I suspect we'll do something similar for duplication when switching it to use the parser server. Channels are super valuable for dogfooding large changes before rolling them out to all customers, and I don't think spot-check canaries allows for the kind of persistent QA that channels do.

PR #695, opened 2017-07-07
Either YAML or JSON, never both

wfleming:
With plugins, we now support all core config fields in the new schema
(excpect prepare, which I'll address soon). To simplify the
understanding of the system, we're moving towards only loading one of
these files: right now, the newer-schema JSON is the pre-eminent one, so
we won't load YAML if the JSON exists.

### DIFF
--- a/lib/cc/config.rb
+++ b/lib/cc/config.rb
@@ -19,8 +19,11 @@ module CC
 
     def self.load
       config = Default.new
-      config = config.merge(YAML.new) if File.exist?(YAML::DEFAULT_PATH)
-      config = config.merge(JSON.new) if File.exist?(JSON::DEFAULT_PATH)
+      if File.exist?(JSON::DEFAULT_PATH)
+        config = config.merge(JSON.new)
+      elsif File.exist?(YAML::DEFAULT_PATH)
+        config = config.merge(YAML.new)
+      end
       config
     end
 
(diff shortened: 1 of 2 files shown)


lib/cc/config.rb
```diff
@@ -19,7 +19,9 @@ class Config
 
     def self.load
       config = Default.new
-      config = config.merge(YAML.new) if File.exist?(YAML::DEFAULT_PATH)
+      if File.exist?(YAML::DEFAULT_PATH) && !File.exist?(JSON::DEFAULT_PATH)
```
nporteschaikin: You could simplify this, I think:

```
if File.exist?(JSON::DEFAULT_PATH)
  # merge JSON
elsif if File.exist?(YAML::DEFAULT_PATH)
  # merge YAML
end
```

lib/cc/config.rb
```diff
@@ -19,7 +19,9 @@ class Config
...
+        config = config.merge(YAML.new)
+      end
       config = config.merge(JSON.new) if File.exist?(JSON::DEFAULT_PATH)
```
pbrisbin: I think this would be more readable like:

```rb
if File.exists?(JSON::DEFAULT_PATH)
  config.merge(JSON.new)
elsif File.exists?(YAML::DEFAULT_PATH)
  config.merge(YAML.new)
end
```

nporteschaikin: image

wfleming: 2017-07-07-125246_480x754_scrot

wfleming:
You're both right, again. Updated.

PR #696, opened 2017-07-07
Enable PHP in duplication engine

pbrisbin:


lib/cc/config/default.rb
```diff
@@ -1,7 +1,14 @@
 module CC
   class Config
     class Default < Config
+      DUPLICATION_LANGUAGES = %[
```
wfleming: Pretty sure you want `%w`

pbrisbin: Whoops

pbrisbin:
@wfleming nice catch. Added a spec and fixed another bug.

PR #698, opened 2017-07-10
Convert old file config to newer config style

wfleming:
The [cc-yaml][] config has been doing the conversion in this direction
for a long time, so I believe pretty much all engines support the newer
config format now. We primarily keep the conversion to support legacy
customer configs, not legacy engines, AFAIK. If there are any lingering
engines, we should fix those instead of re-writing to an older config
style, since the old-style config is *not supported* and will crash on
some engines.

[cc-yaml]: https://github.com/codeclimate/codeclimate-yaml/blob/master/spec/cc/yaml/nodes/engine_config_spec.rb#L58-L70

### DIFF
--- a/lib/cc/config/yaml.rb
+++ b/lib/cc/config/yaml.rb
@@ -33,7 +33,7 @@ module CC
           Engine.new(name, enabled: data)
         else
           if data.key?("config")
-            data["config"] = convert_to_legacy_file_config(data["config"])
+            data["config"] = upconvert_legacy_file_config(data["config"])
           end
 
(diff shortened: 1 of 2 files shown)


wfleming:
> Did you run this on things?

I did not. I will do some local testing before merging since you mention RuboCop and that surprises me.

wfleming:
Urgh, previous code was correct, RuboCop does not like the new style yet.

OTOH, csslint (& maybe others?) crash on the *old* style. So everything is wonderful.

PR #699, opened 2017-07-10
Remove auto-enable plugins

wfleming:
Now that we have in-app config for plugins, this is incompatible & going
away. The replacement will be to eventually have UI to prompt users to
enable plugins themselves based on what's detected in their repo.


lib/cc/analyzer/plugins.rb
```diff
@@ -1,27 +0,0 @@
-module CC
-  module Analyzer
-    class Plugins
```
ABaldwinHunter: will we still want this class for filtering plugins that are configured via the UI or a user's .codeclimate.yml, or will that live elsewhere?

wfleming: @ABaldwinHunter that lives elsewhere, as part of `Config`.

PR #706, opened 2017-07-13
Update release scripts

dblandin:
- Enable shellcheck engine
- Use `id --user` for posix shell compatibility
- Wrap variables in quotes
- Use printf
- Extract pre-release checks
- Use git fetch / reset instead of pull

### DIFF
--- a/bin/prep-release
+++ b/bin/prep-release
@@ -12,24 +12,12 @@ if [ -z "$1" ]; then
   exit 64
 fi
 
-if ! [ "$(git symbolic-ref --short HEAD)" = "master" ]; then
-  echo "must be on master branch to run" >&2
-  exit 1
-fi
-
-if ! git diff-index --quiet HEAD --; then
-  echo "commit or stash your workspace changes" >&2
-  exit 1
-fi
+./bin/validate-release
 
 version="$1"
 old_version=$(< VERSION)
 branch="release-$version"
 
-if ! make test; then
-  echo "test suite failed. not releasing" >&2
-  exit 1
-fi
 
 printf "RELEASE %s => %s\n" "$old_version" "$version"
 git fetch origin master
--- a/bin/prep-release
+++ b/bin/prep-release
@@ -12,6 +12,16 @@ if [ -z "$1" ]; then
   exit 64
 fi
 
+if ! [ "$(git symbolic-ref --short HEAD)" = "master" ]; then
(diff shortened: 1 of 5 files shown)


bin/prep-release
```diff
@@ -8,40 +8,37 @@
 set -e
 
 if [ -z "$1" ]; then
-  echo "usage: bin/prep-release VERSION" >&2
+  printf "usage: bin/prep-release VERSION\n" >&2
```
pbrisbin: FWIW my own style is to use simple `echo` for simple strings (as all these are), and only go to the clunkier `printf` when needed (for escapes, variables, etc).

bin/prep-release
```diff
@@ -8,40 +8,37 @@
...
 
 printf "%s\n" "$version" > VERSION
 make Gemfile.lock
 git add VERSION Gemfile.lock
-git commit -m "Release v$version"
+git commit --message "Release v$version"
 git push origin "$branch"
 
-branch_head=$(git rev-parse --short $branch)
+branch_head="$(git rev-parse --short "$branch")"
```
pbrisbin: Quotes are not needed for the left-hand side of an assignment, and we typically omit them.

```sh
branch_head=$(git rev-parse --short "$branch")
```

bin/release
```diff
@@ -9,23 +9,23 @@
 ###
 set -e
 
-git checkout master
-git pull
+./bin/validate-release
 
-version=$(< VERSION)
+version="$(< VERSION)"
```
pbrisbin: These quotes aren't needed

bin/release
```diff
@@ -9,23 +9,23 @@
...
 docker push codeclimate/codeclimate:latest
 docker tag codeclimate/codeclimate "codeclimate/codeclimate:$version"
 docker push "codeclimate/codeclimate:$version"
 
 (cd ../homebrew-formulae/ && bin/release "$version")
 
-echo "Be sure to update release notes:"
-echo ""
-echo "  https://github.com/codeclimate/codeclimate/releases/new?tag=v$version"
-echo ""
```
pbrisbin: The extra whitespace here was intentional. I found I missed this message if it wasn't set off from all the other logging. To address the shellcheck issue but keep the formatting, you could move to this:

```sh
cat <<EOM
Be sure to...

  https://...

EOM
```

PR #717, opened 2017-08-29
EngineFailure: take engine_name argument

wfleming:
We pass two args to these exceptions now, but the CLI's internal
exception class wasn't updated to take one.


wfleming:
> Thanks! Sorry i didn't know EngineFailure was present here

It's kinda dumb it wasn't covered by a test: I think it *used to be* for the spec of the `analyze` command, but that got "temporarily" erased during QM spike. We'll want to restore it, I'm sure, but honestly it's not something I have time to do right now.

PR #718, opened 2017-09-01
Upgrade CircleCi config to v2

scottbartell:
https://circleci.com/docs/2.0/migrating-from-1-2/


scottbartell:
This appears to be more complicated than I thought it would be. Apparently Circle 2.0 doesn't support mounting volumes this way 
https://discuss.circleci.com/t/why-circleci-2-0-does-not-support-mounting-folders/11605/2
https://circleci.com/docs/2.0/building-docker-images/#mounting-folders

PR #719, opened 2017-09-05
Enable java duplication analysis by default for QM

ABaldwinHunter:
We are already analyzing complexity for java by default on QM.

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -52,10 +52,11 @@ duplication:
     beta: codeclimate/codeclimate-duplication:beta
     cronopio: codeclimate/codeclimate-duplication:cronopio
     stable: codeclimate/codeclimate-duplication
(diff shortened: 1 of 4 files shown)


wfleming:
Pretty sure this isn't actually used in QM and won't work. In fact I think this whole file might be defunct.

You want to modify `CC::Config::Default`.

ABaldwinHunter:
> You want to modify CC::Config::Default.

@wfleming thanks! I think you're right - that's what's used for analysis.

However, this file appears to still be used - even in QM - for the `codeclimate engines:enable duplication` command.

So I think we want this change as well, right? Adding the other

ABaldwinHunter:
@wfleming ready for another look!

wfleming:
> However, this file appears to still be used - even in QM - for the codeclimate engines:enable duplication command.

Interesting. Well, having this in so many places is gonna be terrible long term, but that's a problem to consider another day.

PR #720, opened 2017-09-05
Use cronopio channel by default for duplication

ABaldwinHunter:
Duplication Quality Model lives on the cronopio channel instead of
master.

The default of `stable` can produce confusing results when QM users
modify the duplication engine in .codeclimate.yml.

### DIFF
--- a/lib/cc/config/engine.rb
+++ b/lib/cc/config/engine.rb
@@ -2,6 +2,8 @@ module CC
   class Config
     class Engine
       DEFAULT_CHANNEL = "stable".freeze
+      DUPLICATION_CHANNEL = "cronopio".freeze
+      DUPLICATION = "duplication".freeze
 
       attr_reader :name, :channel, :config
       attr_writer :enabled
@@ -9,7 +11,7 @@ module CC
       def initialize(name, enabled: false, channel: nil, config: nil)
         @name = name
         @enabled = enabled
-        @channel = channel || DEFAULT_CHANNEL
+        @channel = channel || default_channel
         @config = config || {}
       end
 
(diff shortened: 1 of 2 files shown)


lib/cc/config/engine.rb
```diff
@@ -37,6 +39,14 @@ def merge(other)
       # name in the same list. This should be true except maybe if we want to
       # work with multiple channels at once, which is unlikely.
 
+      def default_channel
+        if name == DUPLICATION
+          DUPLICATION_CHANNEL
+        else
+          DEFAULT_CHANNEL
+        end
```
ABaldwinHunter: this check is a bit hacky but I think it's an improvement. once QM is on `stable` we can remove

spec/cc/config/merge_spec.rb
```diff
@@ -60,6 +60,30 @@
...
+        merged_config = described_class.new(config1, config2).run
+
+        expect(merged_config.engines.count).to eq(1)
+
+        merged_config.engines.to_a.first.tap do |engine|
+          expect(engine.name).to eq("duplication")
+          expect(engine.config).to eq(languages: "ruby")
+          expect(engine.channel).to eq(CC::Config::Engine::DUPLICATION_CHANNEL)
+        end
+      end
```
maxjacobson: I think it could be worth covering just the changed behavior: that for duplication it uses cronopio, and for something else, it uses "stable". That second part seems not to be explicitly covered (not that I think it won't work). And the merging behavior probably doesn't need to be covered in this spec, in the interest of keeping specs focused

lib/cc/config/engine.rb
maxjacobson: seems fine to me for now 👍

ABaldwinHunter:
> Specifically: what's happening is that Config::YAML gets parsed and sees a duplication engine, and defaults it to the stable channel when initializing an Engine object. Config later merges Config::Default, which specifies cronopio, but at the point the earlier YAML object wins.

Yeah. In my own testing, I ran into a confusing experience when I updated my `.codeclimate.yml` to use 
```
engines:
  duplication:
    config:
      languages:
      - java
```
and inadvertently turned off java duplication analysis. Suddenly, 0 duplication issues found.

@wfleming @nporteschaikin the points about the best place in the code for this change, and that many QM users probably won't configure the duplication engine in their .codeclimate.yml config are well made.

I'll close this PR. Thanks for review

PR #721, opened 2017-09-08
Continue respecting older style file config

wfleming:
Many configs in the wild use this style, and some plugins still document
it as the preferred style. We were crashing on these because the code
assumed this was a hash.


spec/cc/config/yaml_spec.rb
```diff
@@ -76,6 +76,19 @@
       expect(rubocop).to be_present
       expect(rubocop.config["config"]).to eq("foo.rb")
     end
+
+    it "respects legact file config values" do
```
ale7714: typo `legact`

ale7714 (approved):
LGTM

PR #723, opened 2017-09-18
Treat YAML config as an adapter, not a separate config

wfleming:
This is the first part of a larger effort to make config treatment more
consistent, and to address lingering incompatibilities and edge cases in
Quality Model.

The core idea is that the user-provided config will go through a series
of transformations: instead of the transformations being
largely parallel as they are today (where YAML & JSON are implemented
completely in parallel), they will instead be a *series* of
transformations that build on each other. In this particular PR, the way
that works is that legacy YAML configs have their raw contents modified
to match newer schemas, and then go through the new JSON config logic.


lib/cc/config.rb
```diff
@@ -20,9 +20,9 @@ class Config
     def self.load
       config = Default.new
       if File.exist?(JSON::DEFAULT_PATH)
-        config = config.merge(JSON.new)
-      elsif File.exist?(YAML::DEFAULT_PATH)
-        config = config.merge(YAML.new)
+        config = config.merge(JSON.load)
```
dblandin: It's a little confusing that we're using `JSON` as the class name. Would it make sense to rename this to `JSONAdapter`?

lib/cc/config.rb
```diff
@@ -20,9 +20,9 @@ class Config
...
+      elsif File.exist?(YAMLAdapter::DEFAULT_PATH)
+        config = config.merge(JSON.new(YAMLAdapter.load.config))
```
dblandin: Is there an interface difference here between `JSON.load` and `YAMLAdapter.load`? Wondering why we need the extra `#config` for the YAML adapater but not for the JSON adapter.

lib/cc/config/yaml/validator.rb
```diff
@@ -48,10 +48,19 @@ def invalid_engine_warnings
...
-            engine_exists?(engine)
-          end
+          config = CC::Config::YAMLAdapter.load(path)
+          config.config["plugins"]
+            .map do |name, data|
+              Engine.new(
+                name,
+                enabled: data.fetch("enabled", true),
+                channel: data["channel"],
+                config: data["config"],
```
dblandin: Do `channel` and `config` account for `nil` values correctly within `Engine`? Or do we need `stable` and `{}` fallbacks here?

lib/cc/config/yaml/validator.rb
```diff
@@ -48,10 +48,19 @@ def invalid_engine_warnings
...
+              )
+            end.
+            reject do |engine|
```
dblandin: Should we error in the case of an unknown engine instead of silently pruning the entry?

lib/cc/config/yaml_adapter.rb
```diff
@@ -0,0 +1,72 @@
...
+      #
+      #   { config: PATH }
+      #
+      # But we document, and hope to eventually move to:
+      #
+      #   { config: { file: PATH } }
+      #
+      # We need to munge from the latter to the former when/if we encounter it
+      def convert_to_legacy_file_config(config)
+        if config.is_a?(Hash) && config.keys.size == 1 && config.key?("file")
```
dblandin: Might be able to use `config.keys.one?` here.

lib/cc/config.rb
wfleming: That's coming! I'm trying to make smaller changes so the diffs don't get too crazy with just name changes. I'm happy to keep pushing on this branch in commits if you'd prefer to review that way.

lib/cc/config.rb
wfleming: Because the JSON hasn't been turned into an `Adapter` of the same style *yet*. It will be.

lib/cc/config/yaml/validator.rb
wfleming: Both of those are defaulted as expected within `Engine`.

lib/cc/config/yaml/validator.rb
wfleming: We're pruning entries that *do* exist. The validator constructs all engines, deletes ones that are valid, and reports on the remaining ones.

I am planning to rebuild the validator in depth after stabilizing the rest of this: the current changes here were to keep specs passing/behavior consistent in the short term.

lib/cc/config/yaml_adapter.rb
wfleming: Changed, good catch.

PR #724, opened 2017-09-19
Handle all config as a series of transformations

wfleming:
Takes the approach of #723 & applies it more broadly. IMHO, this is a simpler & more straightforward way of layering the several config transformations we do (i.e. upgrading legacy schema, applying defaults, and making QM check config available to QM engines), and potentially making it simpler to add other transformations if needed in the future.


wfleming:
@dblandin might be easier to go through this one commit-by-commiit

lib/cc/config/default.rb
```diff
@@ -1,57 +0,0 @@
-module CC
-  class Config
-    class Default < Config
-      DUPLICATION_LANGUAGES = %w[
```
wfleming: Mentioned in a commit message, but we don't need these anymore: `duplication` provides a default itself if one is not considered.

PR #726, opened 2017-09-19
Rewrite config validation

wfleming:
Our schema is sufficiently different under QM that the old validation checks
are not terribly helpful. We've intended to abandon `codeclimate-yaml` for some
time, since it's type-coercion behavior has been troublesome. Addtionally, we
now support the JSON format, so we need to validate that as well.

This implements new validation logic, some of which is entirely new & some of which is
effectively ported from `codeclimate-yaml`. It also adds validation support for JSON, and attempts to cover legacy vs QM differences by reporting on deprecated keys.


lib/cc/cli/validate_config.rb
```diff
@@ -56,11 +50,36 @@ def display_issues
...
 
-        validator.nested_warnings.each do |warning|
-          puts "#{colorize("WARNING in #{warning.field}", :yellow)}: #{warning.message}"
+      def require_json_or_yaml
+        if !filesystem.exist?(Config::YAMLAdapter::DEFAULT_PATH) && !filesystem.exist?(Config::JSONAdapter::DEFAULT_PATH)
+          puts NO_CONFIG_MESSAGE
+          exit 0
+        elsif filesystem.exist?(Config::YAMLAdapter::DEFAULT_PATH) && filesystem.exist?(Config::JSONAdapter::DEFAULT_PATH)
+          puts "#{colorize("ERROR", :red)}: #{TOO_MANY_CONFIG_MESSAGE}"
+          exit 1
```
dblandin: Do we want to exit hard here or emit a warning and continue validating with the config file with higher precedence?

From what I remember, we're going to document a config file precedence giving priority to `.codeclimate.json`.

wfleming: Fair enough: I had favored erroring hard since it's a state that really isn't supposed to exist & warnings aren't the most user-obvious feedback if there aren't errors. But you're right, so I've pushed a new commit addressing that.

dblandin: I think we also have an app UI change in the pipeline to show which config file (or in-app config) is used, so this should be more visible.

PR #728, opened 2017-09-20
Respect engine specific excludes when running the engine

wfleming:
Added a small spec for `Bridge` as well since we didn't have one.

### DIFF
--- a/lib/cc/config/yaml_adapter.rb
+++ b/lib/cc/config/yaml_adapter.rb
@@ -55,6 +55,9 @@ module CC
           plugins = config.fetch("plugins", {})
           plugins.each do |engine, data|
             plugins[engine] = coerce_engine(data)
+            if plugins.fetch(engine)["exclude_paths"]
+              plugins.fetch(engine)["exclude_patterns"] ||= plugins.fetch(engine).delete("exclude_paths")
+            end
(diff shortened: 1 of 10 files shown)


lib/cc/config/yaml_adapter.rb
```diff
@@ -55,6 +55,9 @@ def upconvert_legacy_yaml!
           plugins = config.fetch("plugins", {})
           plugins.each do |engine, data|
             plugins[engine] = coerce_engine(data)
+            if plugins.fetch(engine)["exclude_paths"]
+              plugins.fetch(engine)["exclude_patterns"] ||= plugins.fetch(engine).delete("exclude_paths")
```
maxjacobson: nice use of `||=` to express that it shouldn't override the `exclude_patterns` when present. Worth covering that precedence behavior with a test?

maxjacobson: ~~Feel free to punt on this, but I'm wondering if we've considered any kind of strategy where we log a warning that they're using a legacy key, and that they should use the new thing?~~ I see you have this! That's what I get for reviewing commit-by-commit 😜

wfleming: Test added, will merge on green.

PR #730, opened 2017-09-20
Rm EngineRegistry#each_engine

wfleming:
This method is unused (here as well as `Builder`).


wfleming:
Re-purposing this for #733 -- closing

PR #733, opened 2017-09-21
Use new Config & EngineRegistry for remaining engines commands

wfleming:
Trying to move away from the legacy `Analyzer::Config` and
`Analyzer::EngineRegistry` to get more things using the newer
equivalents.

This happens to improve some behavior of `engines:install` as well: previously
it would install every available channel for an enabled
engine, which is pretty excessive. This now only installs the image for
whatever channel the user has enabled.

### DIFF
--- a/lib/cc/cli/engines/install.rb
+++ b/lib/cc/cli/engines/install.rb
@@ -13,10 +13,6 @@ module CC
 
         private
 
-        def engine_registry
-          @engine_registry ||= EngineRegistry.new
-        end
-
         def config
           @config ||= CC::Config.load
         end
--- a/lib/cc/cli/engines/install.rb
+++ b/lib/cc/cli/engines/install.rb
@@ -2,38 +2,36 @@ module CC
   module CLI
     module Engines
       class Install < EngineCommand
-        SHORT_HELP = "Pull the latest images for enabled engines in your codeclimate.yml.".freeze
+        SHORT_HELP = "Pull the latest images for enabled engines in your configuration".freeze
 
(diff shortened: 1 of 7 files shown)


lib/cc/cli/engines/install.rb
```diff
@@ -2,38 +2,33 @@ module CC
...
+        def pull_docker_images
+          config.engines.each(&method(:pull_engine))
         end
 
-        def pull_engine_image(engine_image)
-          unless system("docker pull #{engine_image}")
-            raise ImagePullFailure, "unable to pull image #{engine_image}"
+        def pull_engine(engine)
+          metadata = engine_registry.fetch_engine_details(engine)
+          $stderr.puts "DEBUG DOING A DOCKER PULL"
```
toddmazierski: Is the all-caps formatting intentional?

lib/cc/cli/engines/install.rb
```diff
@@ -2,38 +2,33 @@ module CC
...
-            end
-          end
+        def config
+          @config ||= CC::Config.load
         end
 
-        def engine_names
-          @engine_names ||= parsed_yaml.engine_names
+        def pull_docker_images
+          config.engines.each(&method(:pull_engine))
```
toddmazierski: `&method` 🆒

lib/cc/cli/engines/install.rb
wfleming: Ack, this wasn't meant to be in here at all, thanks for catching. Fixed.

PR #734, opened 2017-09-21
Remove some unused code

wfleming:
Specifically, we no longer need the older `Analyzer::Config` or `Analyzer::EngineRegistry`. There were also a couple dangling `autoload` declarations for classes that had already been deleted.


spec/cc/analyzer/issue_spec.rb
```diff
@@ -30,7 +17,7 @@ module CC::Analyzer
       it "adds a fingerprint when it is missing" do
         issue = Issue.new("", output.to_json)
 
-        expect(issue.fingerprint).to eq "433fae1189b03bcd9153dc8dce209fa5"
```
wfleming: These fingerprints were all based on reading the source of `lib/cc/analyzer/config.rb`, which I've deleted, so they changed here because I moved the relevant source to a new fixture file.

I am so mad that anybody ever thought it was a good idea to write tests that relied on an unrelated source file not changing.

spec/cc/analyzer/issue_spec.rb
```diff
@@ -3,20 +3,7 @@
 module CC::Analyzer
   describe Issue do
     let(:output) do
-      {
```
wfleming: This data was already basically identical to the contents of `Factory#sample_issue`, may as well use it instead.

spec/support/factory.rb
```diff
@@ -42,16 +42,16 @@ def create_yaml_with_no_engines
   def sample_issue(options = {})
     {
       "type" => "issue",
-      "check" => "Rubocop/Style/Documentation",
```
wfleming: This factory has had the wrong key for a long, long time. Nice.

spec/cc/analyzer/issue_spec.rb
ale7714: ohh :(

PR #735, opened 2017-09-21
Config validation: warn about missing version key

wfleming:
This should always exist in JSON configs, but we still support v1 YAML,
so in YAML it only warns if it sees one of the v2 keys.

### DIFF
--- a/lib/cc/config/validation/json.rb
+++ b/lib/cc/config/validation/json.rb
@@ -10,6 +10,7 @@ module CC
 
           return unless validate_hash_data
 
+          validate_version
           validate_prepare
           validate_engines("plugins")
           validate_checks
@@ -18,6 +19,12 @@ module CC
           warn_unrecognized_keys(%w[prepare plugins exclude_patterns version])
         rescue ::JSON::ParserError => ex
           errors << "Unable to parse: #{ex.message}"
+        end
+
+        def validate_version
+          unless data.key?("version")
+            warnings << %{missing 'version' key. Please add `"version": "2"`}
+          end
         end
       end
     end
--- a/lib/cc/config/validation/yaml.rb
+++ b/lib/cc/config/validation/yaml.rb
@@ -10,6 +10,7 @@ module CC
 
           return unless validate_hash_data
 
+          validate_version
           validate_prepare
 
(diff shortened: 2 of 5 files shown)


lib/cc/config/validation/yaml.rb
```diff
@@ -29,6 +30,12 @@ def validate
           errors << "Unable to parse: #{ex.message}"
         end
 
+        def validate_version
+          if !data.key?("version") && (data.key?("plugins") || data.key?("exclude_paths"))
```
dblandin: Isn't `exclude_paths` a common key in version 1 configs?

spec/cc/config/validation/json_spec.rb
```diff
@@ -219,6 +220,17 @@
     expect(validator.warnings).to include("unrecognized key 'exclude_pattttttterns'")
   end
 
+  describe "version validation" do
+    it "warns about missing version" do
+      validator = validate_json(<<-EOJSON)
```
toddmazierski: TIL that the closing parenthesis can go here! I've always placed it after the closing heredoc.

lib/cc/config/validation/json.rb
```diff
@@ -19,6 +20,12 @@ def validate
         rescue ::JSON::ParserError => ex
           errors << "Unable to parse: #{ex.message}"
         end
+
+        def validate_version
+          unless data.key?("version")
+            warnings << %{missing 'version' key. Please add `"version": "2"`}
```
toddmazierski: Is 2 the current version? Would it make sense to make a constant out of it so this warning is correct when it's bumped, or do we not expect it to change that often?

lib/cc/config/validation/yaml.rb
wfleming: d'oh, I meant to type `exclude_patterns`. Fixing.

lib/cc/config/validation/json.rb
wfleming: Version should change really rarely. I'm inclined to leave it messy since any version bump would probably radically change the organization of this parsing/validation, so leaning on DRY now would probably just make that potential future action harder.

PR #736, opened 2017-09-21
Fix pending spec

wfleming:
This spec has been pending for a while, I updated the expectation to
match the changes that caused it to break way back when.

### DIFF
--- a/spec/cc/analyzer/engine_spec.rb
+++ b/spec/cc/analyzer/engine_spec.rb
@@ -50,11 +50,10 @@ module CC::Analyzer
           expect(Container).to receive(:new).and_return(container)
 
           stdout = TestFormatter.new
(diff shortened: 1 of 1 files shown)


spec/cc/analyzer/engine_spec.rb
```diff
@@ -50,11 +50,10 @@ module CC::Analyzer
           expect(Container).to receive(:new).and_return(container)
 
           stdout = TestFormatter.new
-          engine = Engine.new("", { "image" => "" }, {}, "")
+          engine = Engine.new("bar", { "image" => "" }, {}, "")
           engine.run(stdout)
 
-          pending # WTF is with this expectation? i just added an engine_name key, now i have to fight my way through this whole thing
```
dblandin: heh :joy:

dblandin (approved):

PR #737, opened 2017-09-21
Increase default engine memory to 1GB

wfleming:
512MB is pretty restrictive, and will become more restrictive in Quality
Model with us adding parsers for more memory-hungry languages like Java.

I think 1 GB is eminitely reasonable as a default for CLI usage. Our
defaults for hosted analysis are already higher than this.

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -20,7 +20,7 @@ module CC
     class Engine
       Error = Class.new(StandardError)
 
-      DEFAULT_MEMORY_LIMIT = 512_000_000
+      DEFAULT_MEMORY_LIMIT = 1_024_000_000
 
       def initialize(name, metadata, config, label)
         @name = name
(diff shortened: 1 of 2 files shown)


dblandin (approved):
LGTM pending green statues

pointlessone:
This limit is [mentioned in the Spec](https://github.com/codeclimate/spec/blob/master/SPEC.md#resource-restrictions). And it's there for at least 7 months.

wfleming:
Thanks @pointlessone, I'll address the spec after this. Since the limit has been higher in web analysis for over a year, the spec was already pretty inaccurate.

wfleming:
Actually I think the spec is fine right now: the two relevant lines I see are:

```
* The Docker image for an Engine must not exceed 512MB, including all layers.
* The combined total RSS memory usage by all processes within the Docker container must not exceed 1GB at any time.
```

I think Alex was noticing that the first line, but that's a restriction on image size, not container memory usage. The second line already said 1GB, so CLI was being more restrictive than the spec suggested it should have been!

pointlessone:
Sorry if I wasn't clear. I was pointing out that CLI was behind the spec for at least 7 month with regards to memory limit.

wfleming:
Oh, my bad, I totally misunderstood you, and the *image* memory limit being coincidentally the same as the old container memory limit increased my confusion.

PR #740, opened 2017-09-27
Fix config handling of `analyze -e`.

wfleming:
We had a regression in QM with respecting engine config when only
certain engines are selected for running with `-e` or `--engine`.

The change here makes it so that `-e` will not reset any
already-specified config for an engine, but will still add a new engine
with default config when the named engine wasn't already configured.

### DIFF
--- a/lib/cc/cli/analyze.rb
+++ b/lib/cc/cli/analyze.rb
@@ -38,7 +38,7 @@ module CC
 
       private
 
-      attr_reader :config, :listener, :registry
+      attr_reader :config, :engines_disabled, :listener, :registry
 
       def process_args
         while (arg = @args.shift)
@@ -46,17 +46,9 @@ module CC
           when "-f", "--format"
             @formatter = Formatters.resolve(@args.shift).new(filesystem)
           when "-e", "--engine"
-            # First time passed, clear any configured engines
-            unless @cleared
-              config.engines.clear
-              @cleared = true
-            end
+            disable_all_engines!
             name, channel = @args.shift.split(":", 2)
-            config.engines << Config::Engine.new(
-              name,
-              channel: channel,
-              enabled: true,
-            )
+            enable_engine(name, channel)
           when "--dev"
             config.development = true
           when "--no-plugins"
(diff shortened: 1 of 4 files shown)


lib/cc/cli/analyze.rb
```diff
@@ -72,6 +64,27 @@ def process_args
...
+      def disable_all_engines!
+        unless engines_disabled
+          config.engines.each { |e| e.enabled = false }
+          @engines_disabled = true
+        end
+      end
+
+      def enable_engine(name, channel)
+        existing_engine = config.engines.detect { |e| e.name == name }
+        if !existing_engine.present?
```
dblandin: What do you think of flipping this so that `if existing_engine.present?` is the first block?

lib/cc/cli/analyze.rb
```diff
@@ -72,6 +64,27 @@ def process_args
       def formatter
         @formatter ||= Formatters::PlainTextFormatter.new(filesystem)
       end
+
+      def disable_all_engines!
+        unless engines_disabled
```
dblandin: Do you expect this to be called more than once? Wondering why we need the `engines_disabled` check.

wfleming: If you pass `-e foo -e bar`, you only want to disable all the engines one time, otherwise you'll *only* end up with `bar` enabled.

This logic was inline in `process_args`: I extracted it to a method, and kept the once-only guard here since otherwise it didn't seem worth extracting.

lib/cc/cli/analyze.rb
wfleming: Done

wfleming:
I don't know why the spec in https://circleci.com/gh/codeclimate/codeclimate/2168 is failing: it passes for me locally, and it passed previously on the `quality-model` branch. No obvious reason it should be flickery. Looking into it.

lib/cc/cli/analyze.rb
dblandin: Ahh, that makes sense 👍  I didn't realize `-e` was accepted multiple times.

wfleming:
I identified the spec breakage and have that open as #742, since it's not strictly related to this change: I think this change just shifted GC of the suite reliably enough to trigger the issue. I'll rebase here once that's merged and it should be green (it was when I had the fix on this branch: https://circleci.com/gh/codeclimate/codeclimate/2177)

@dblandin Can you take another look here, or was your comment at https://github.com/codeclimate/codeclimate/pull/740#discussion_r141445325 an implied approval?

dblandin (approved):
LGTM!

PR #742, opened 2017-09-27
Skip engines that are missing, warn about them being skipped

wfleming:
In .com we "skip" invalid engines, and don't crash on them. Treating a
missing engine as an exit code `99` will make maintaining that behavior
in .com difficult due to some other details of error handling.

This seems preferable: we'll treat skipped results as exit code `0`
(since they're not "errors"), and mark them as `skipped`, similar to
other meta-states like timed out.

I've also made the CLI's handling of this state more similar to what
.com's behavior used to be, and will be again: we don't crash, we just
warn.


wfleming:
@maxjacobson Good catch, I have written more specs, all lines are now super green.

maxjacobson:
🍏 📗 💚 🥗 🇬🇱

PR #743, opened 2017-09-28
Config: update validation for some plugin config

wfleming:
We still support checks within plugins for filtering, and also support exclude_patterns (although it's deprecated). This updates our validation checks to say the appropriate things.

### DIFF
--- a/lib/cc/config/validation/engine_validator.rb
+++ b/lib/cc/config/validation/engine_validator.rb
@@ -32,7 +32,9 @@ module CC
           validate_key_type("config", [String, Hash])
           validate_key_type("exclude_patterns", [Array])
 
-          warn_unrecognized_keys(%w[enabled channel config exclude_patterns])
+          validate_checks
+
+          warn_unrecognized_keys(%w[enabled channel checks config exclude_patterns])
         end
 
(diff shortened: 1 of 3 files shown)


lib/cc/config/validation/engine_validator.rb
```diff
@@ -42,6 +45,23 @@ def validate_root
...
+            validator = CheckValidator.new(check_data)
+            errors.push(*validator.errors)
+            warnings.push(*validator.warnings)
+          end
+        end
+
+        def validate_exclude_fingerprints
+          validate_key_type("exclude_fingerprints", Array)
+          if data.key?("exclude_fingerprints")
+            warnings << "'exclude_fingerprints' is deprecated, and support may be removed in the future"
```
maxjacobson: Worth mentioning that they can use issue statuses, instead?

maxjacobson (approved):

lib/cc/config/validation/engine_validator.rb
wfleming: Well, if they're primarily CLI users, they can't yet, so I don't want to say that yet.

PR #754, opened 2017-10-26
Allow engines to tell a minimum required memory to run

filipesperandio:
Set SonarJava memory to 2G

Related to https://github.com/codeclimate/app/issues/6291

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -185,7 +185,7 @@ sonar-java:
     stable: codeclimate/codeclimate-sonar-java
     beta: codeclimate/codeclimate-sonar-java:beta
   description: SonarLint for Java.
-  required_memory: 2_048_000_000
+  minimum_memory_limit: 2_048_000_000
 structure:
   channels:
     beta: codeclimate/codeclimate-structure:beta
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -185,6 +185,7 @@ sonar-java:
     stable: codeclimate/codeclimate-sonar-java
     beta: codeclimate/codeclimate-sonar-java:beta
   description: SonarLint for Java.
+  memory: 2_048_000_000
 structure:
   channels:
     beta: codeclimate/codeclimate-structure:beta
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -185,7 +185,7 @@ sonar-java:
     stable: codeclimate/codeclimate-sonar-java
     beta: codeclimate/codeclimate-sonar-java:beta
   description: SonarLint for Java.
-  memory: 2_048_000_000
+  required_memory: 2_048_000_000
 structure:
   channels:
     beta: codeclimate/codeclimate-structure:beta
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -20,8 +20,6 @@ module CC
     class Engine
       Error = Class.new(StandardError)
 
-      DEFAULT_MEMORY_LIMIT = 1_024_000_000
-
       def initialize(name, metadata, config, label)
         @name = name
         @metadata = metadata
@@ -34,8 +32,8 @@ module CC
         write_config_file
 
         container = Container.new(
-          image: @metadata.fetch("image"),
-          command: @metadata["command"],
+          image: metadata.fetch("image"),
+          command: metadata["command"],
           name: container_name,
         )
 
@@ -52,7 +50,7 @@ module CC
 
       private
 
-      attr_reader :name
+      attr_reader :name, :metadata
       attr_accessor :error
 
       def handle_output(container, io, raw_output)
@@ -80,7 +78,7 @@ module CC
           "--cap-drop", "all",
           "--label", "com.codeclimate.label=#{@label}",
           "--log-driver", "none",
-          "--memory", memory_limit,
+          "--memory", metadata["memory"].to_s,
           "--memory-swap", "-1",
           "--net", "none",
           "--rm",
@@ -118,15 +116,6 @@ module CC
 
       def output_overrider
         @output_overrider ||= EngineOutputOverrider.new(@config)
-      end
-
-      # Memory limit for a running engine in bytes
-      def memory_limit
-        [@metadata["memory"].to_i, default_memory_limit.to_i].max.to_s
-      end
-
-      def default_memory_limit
-        ENV["ENGINE_MEMORY_LIMIT_BYTES"] || DEFAULT_MEMORY_LIMIT
       end
     end
   end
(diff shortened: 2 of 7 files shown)


config/engines.yml
```diff
@@ -184,6 +184,7 @@ sonar-java:
   channels:
     beta: codeclimate/codeclimate-sonar-java:beta
   description: SonarLint for Java.
+  memory: 2_048_000_000
```
dblandin: What do you think about changing this key to `memory_limit`?

lib/cc/engine_registry.rb
```diff
@@ -27,7 +27,7 @@ def each
 
     def fetch_engine_details(engine, development: false)
       if development
-        EngineDetails.new("codeclimate/codeclimate-#{engine.name}", nil, "")
+        EngineDetails.new("codeclimate/codeclimate-#{engine.name}", nil, "", nil)
```
dblandin: I don't think you have to explicitly pass in `nil` here. Any missing arguments will default to a `nil` value:

```ruby
EngineDetails.new("codeclimate/codeclimate-#{engine.name}", nil, "")
```

lib/cc/engine_registry.rb
```diff
@@ -36,6 +36,7 @@ def fetch_engine_details(engine, development: false)
           [prefix, channels.fetch(engine.channel)].join,
           metadata.fetch("command", DEFAULT_COMMAND),
           metadata.fetch("description", "(No description available)"),
+          metadata.fetch("memory", nil),
         )
```
dblandin: What do you think of moving ` def default_memory_limit` here and having that be the fallback value for this fetch?

lib/cc/analyzer/engine.rb
```diff
@@ -122,7 +122,11 @@ def output_overrider
 
       # Memory limit for a running engine in bytes
       def memory_limit
-        (ENV["ENGINE_MEMORY_LIMIT_BYTES"] || DEFAULT_MEMORY_LIMIT).to_s
+        [@metadata["memory"].to_i, default_memory_limit.to_i].max.to_s
```
dblandin: I wonder if this logic is something we want to introduce right now. What do you think of adding an item in the chain of precedence but keeping out the max logic?

So the precedence would be:
- ENV
- yaml registry (new)
- `DEFAULT_MEMORY_LIMIT` constant

filipesperandio: @dblandin The problem with this approach is that we always have `ENGINE_MEMORY_LIMITS_BYTES` set when invoked from the `builder`. The ENV is set in `scheduler` and cascades from there, so we will never get to use the yaml entry in this scenario, and that's the reason I was treating it like a `minimum_required_memory` instead.

dblandin: Ahh, right. That makes sense. The max behavior sounds good to me then.

pointlessone:
I wonder why memory requirement doesn't go into engine specification file but rather into a registry. This sort of defeats the _tell_ part.

We also have more registries. Do we need to update those as well? What if CLI registry runs a different version of engine than the one in the registry? Say, someone uses older CLI but fetches a new engine build that requires more memory than specified in the registry?

I'd suggest adding memory requirement to the engine spec file. Use some sensible default if it's not specified. Cap it in runner to prevent OOMs.

filipesperandio:
@pointlessone That's an interesting thought. Let me look into it.
The way it is now we definitely need to update other registries, like the one in `builder`.

I think the versioning problem will always be present and I think we are having decent defaults to cover that.

pointlessone:
> I think the versioning problem will always be present

We can try and lit the effects.

> and I think we are having decent defaults to cover that.

I'm not absolutely sure about this. Imagine the following scenario:

A dev uses CC CLI 0.69. It's default mem limit is 1GB. An engine in its registry has no memory requirement specified. A new build of said engine is pushed. It wants more than 1 GB of memory. It starts crashing randomly (because that's how most OOM problems start). To fix it we need to update at least two registries and cut a new CLI release.

Alternatively, we can have a reasonably big mem cap (say, 4GB), default to 1GB and pull the actual value from inside the container. In this case we don't need to update any registries or cut CLI releases. Users can just pull new build of an engine and the configuration will be applied.

The downside might be that we need to update the spec and bump its version. And maybe start more thoroughly validate spec compliance. Maybe.

filipesperandio:
@pointlessone Yeah, not something too simple. Also not sure we want to spin up a container to get that info before starting the analysis, but I might be overthinking this right now.

I agree that updating multiple registries is not ideal, but my understanding is that it is like this for other things as well like `channels`.

config/engines.yml
filipesperandio: From the explanation below, this is not the `limit`... maybe `minimum_required_memory`?

pointlessone:
I'm not very familiar with this particular part. I thought we use the spec for something (that is, it's available for the runner). Otherwise why do we need it?

Even if it requires starting a container we don't need lots of resources. We need maybe 2MB of RAM and maybe a second of CPU time to run `cat /engine.json`. That is unless ther's another way to get it.

dblandin:
Unfortunately, the state of affairs regarding engine/plugin management is not so great. While we have plenty of automation to ease the process it still relies on registry documents instead of an API-backed system. In general, I think the CLI should be more forgiving than production registries which target systems which known constraints rather than any system running the CLI.

I think proposing a spec addition for plugin system requirements (like memory) is a good one, but should probably be taken out of this first iteration, which is a good step forward.

Also unfortunately, the `/engine.json` file in the spec is not widely followed. In the future, I'd love a better plugin management interface that allows plugin authors to submit their own updates and I imagine the `/engine.json` file will be useful to infer information regarding the submission (the docker image).

config/engines.yml
dblandin: I think `memory_limit` is probably fine for the registry. The fact that we can give it more memory either via the CLI or in production can be a behind the scenes detail.

filipesperandio: @dblandin Made it `required_memory`. Would give it another shot?

dblandin: The option this affects it the upper limit of memory a container can use, not necessarily guaranteeing that amount of memory will be available. That's more of a clustering/scheduling issue.

What do you think of `minimum_memory_limit` or simply `memory_limit` as an override key to the default, but knowing that the _allowed_ memory limit might be increased through other means, like a repo-override or CLI option.

filipesperandio: `minimum_memory_limit` makes sense to me! Adjusted.

filipesperandio: @dblandin Do you mind giving another shot?

PR #758, opened 2017-10-26
Enable spec focus

filipesperandio:


spec/spec_helper.rb
```diff
@@ -9,3 +9,10 @@
 
 ENV.delete("CODECLIMATE_DOCKER")
 ENV["CODECLIMATE_TMP"] = Dir.mktmpdir
+
+RSpec.configure do |config|
+  config.before(:example, :focus) { raise "Should not commit focused specs" } if ENV["CIRCLECI"]
```
maxjacobson: Good calling including this 👍 

I think we should consider using the more generic `if ENV["CI"]` which I believe will be set in all CI environments, just in case we migrate to some other service (something we've been talking about a bit recently)

PR #764, opened 2017-11-06
Add sonar-python

filipesperandio:


config/engines.yml
```diff
@@ -191,6 +191,11 @@ sonar-php:
     stable: codeclimate/codeclimate-sonar-php
   description: SonarLint for PHP.
   minimum_memory_limit: 2_048_000_000
+sonar-python:
+  channels:
+    stable: codeclimate/codeclimate-sonar-python
```
chrishulton: curious, why is this going into the `stable` channel?  This engine is under development, right?  Is it because this is the `quality-model` branch of CLI?

filipesperandio: @chrishulton The engine is based in the https://github.com/codeclimate/sonar-wrapper lib which we use for `sonar-java` and `sonar-php` already, meaning the infra was well tested already. The repo is also a fork from `sonar-php`, which also give me some confidence.
Adding it to the `stable` channel is to actually skip the `beta` step.

We don't have to go straight to `stable` though. WDYT?

chrishulton: Thanks for the context.  Since the infra is well QA'd and we haven't announced this engine yet, I think it's OK to go to stable, especially if you followed that process for php.

PR #766, opened 2017-11-08
Include engine channels in metrics

wfleming:
We've had this on master for a long time, and I've found the lack of it
frustrating while monitoring QM. It's good to be able to distinguish
engine metrics by channel.

Shipping this will require adjusting a few charts/alerts, I'll do that
after shipping.


wfleming:
@GordonDiggs pinging you for review in case you were party to a discussion on dropping this with Pat that I missed.

PR #771, opened 2017-11-13
Measurement document support

wfleming:
See spec changes at https://github.com/codeclimate/spec/pull/59, first engine usage at https://github.com/codeclimate/codeclimate-structure/pull/289

### DIFF
--- a/lib/cc/analyzer/engine_output.rb
+++ b/lib/cc/analyzer/engine_output.rb
@@ -24,7 +24,7 @@ module CC
         if issue?
           as_issue.to_json
         elsif measurement?
-          raw_output
+          Measurement.new(name, raw_output).to_json
         end
       end
 
--- a/lib/cc/analyzer/engine_output.rb
+++ b/lib/cc/analyzer/engine_output.rb
@@ -13,13 +13,18 @@ module CC
       end
 
       def measurement?
-        valid_with_type?("issue")
+        valid_with_type?("measurement")
+      end
+
+      def as_issue
+        Issue.new(name, raw_output)
       end
 
       def to_json
         if issue?
-          Issue.new(name, raw_output).to_json
-        elsif valid_with_type?("measurement")
+          as_issue.to_json
+        elsif measurement?
+          raw_output
         end
       end
 
@@ -60,6 +65,8 @@ module CC
       def validator
         if issue?
           IssueValidator.new(parsed_output)
+        elsif measurement?
+          MeasurementValidator.new(parsed_output)
         end
       end
     end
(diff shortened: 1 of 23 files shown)


lib/cc/analyzer/measurement_validations.rb
```diff
@@ -0,0 +1,16 @@
...
+    module MeasurementValidations
+      autoload :NameValidation, "cc/analyzer/measurement_validations/name_validation"
+      autoload :TypeValidation, "cc/analyzer/measurement_validations/type_validation"
+      autoload :ValueValidation, "cc/analyzer/measurement_validations/value_validation"
+      autoload :Validation, "cc/analyzer/measurement_validations/validation"
+
+      def self.validations
+        constants.sort.map(&method(:const_get)).select do |klass|
+          klass.is_a?(Class) && klass.superclass == Validation
+        end
```
toddmazierski: Do you think the literal array `[NameValidation, TypeValidation, ValueValidation]` might be preferable to iterating through `constants`? Besides being more straightforward, I wonder if there's a potential bug here if we added a helper module, for example.

lib/cc/analyzer/engine_output.rb
```diff
@@ -39,14 +50,24 @@ def valid_json?
...
       rescue JSON::ParserError
         nil
       end
 
       def validator
-        IssueValidator.new(parsed_output)
+        if issue?
+          IssueValidator.new(parsed_output)
+        elsif measurement?
+          MeasurementValidator.new(parsed_output)
```
toddmazierski: Perhaps we can try this if/when there's a third type, but both `#to_json` and `#validator` have knowledge of all possible document types. It might be nice to have an intermediate class to call `#to_json` and `#validator` _on_ instead.

lib/cc/analyzer/measurement_validations.rb
wfleming: I like that this allows us to add future validations with less boilerplate code around which validations to run: we have the same bit of code in `IssueValidations`, so I'm also in favor of treating the two things the same.

>  I wonder if there's a potential bug here if we added a helper module

We have specs asserting the validations that are run, so that seems unlikely.

lib/cc/analyzer/engine_output.rb
wfleming: I also feel like the current pattern is ugly, and definitely would want to do that extraction if we add a third type. I don't see it as absolutely necessary to do for this current change, and didn't do it since I didn't come into this with a very strong idea of what that new type would be called/look like. If you have a good idea there, I'm happy to do it now.

wfleming:
@toddmazierski commented, also added a new commit to catch a thing I realized I need on the builder side.

lib/cc/analyzer/engine_output.rb
toddmazierski: > If you have a good idea there, I'm happy to do it now.

I don't! So let's hold off on that.

PR #774, opened 2017-11-15
Cleanup plugin description copy

larkinscott:
- Update some description copy to match app counterpart 
- Ensure all plugin descriptions end with a period


wfleming:
@larkinscott text changes look good, but I think you want to make this change against `quality-model` instead of `master`: we're not really rebasing `quality-model` these days, and we're going to try and merge QM back to `master` real soon, so it'll be easier to just make these changes on QM right now.

PR #777, opened 2017-11-16
Release 10-point maintainability analysis

wfleming:
- New maintainability analysis for Ruby, JavaScript, PHP, Python, and Java
- Duplication support for Java
- Support for version 2 of our config file schema: see [documentation](https://docs.codeclimate.com/docs/advanced-configuration)
- A checked in `.codeclimate.yml` is no longer required: if the defaults work for you, you can just run `codeclimate analyze`.
- The new maintainability checks are *always* run by default. If you wish to turn off the maintainability analysis, you'll want to add this to your `.codeclimate.yml`:

    ```yaml
    checks:
      argument-count:
        enabled: false
      complex-logic:
        enabled: false
      file-lines:
        enabled: false
      method-complexity:
        enabled: false
      method-count:
        enabled: false
      method-lines:
        enabled: false
      nested-control-flow:
        enabled: false
      return-statements:
        enabled: false
      similar-code:
        enabled: false
      identical-code:
        enabled: false
    ```

    (That block also turns off duplication analysis: if you are already using duplication & want to keep it, leave off the last two checks listed above.)
- Several commands have been deprecated & removed: `init`, `engines:enable`, `engines:disable`, & `engines:remove` are no longer supported commands. Refer to the README or run `codeclimate help` for information about available commands.


wfleming:
I rebased `quality-model` on `master` to make the merge easier & resolved conflicts. To verify correctness, I:

- `git diff origin/maser -- config/engines.yml` to verify only expected changes were made to the engines registry.
- Sanity checked `git diff origin/quality-model` against this branch to make sure the differences of this rebased branch are what we expect.

Gemfile.lock
```diff
@@ -1,32 +1,27 @@
 PATH
   remote: .
   specs:
-    codeclimate (0.69.0)
-      activesupport (>= 4.2.1, < 5.1)
-      codeclimate-yaml (~> 0.13.0)
+    codeclimate (0.66.0)
```
chrishulton: hm, should this go back to `0.66` since we ended up getting up to `0.69` on master?  there is also the `VERSION` file on this branch which looks like it got updated to `0.69`.

GordonDiggs:
Version bumps should happen in their own PRs, but that raises an interesting question - are there breaking changes here? Should this be `1.0.0`?

dblandin:
> Several commands have been deprecated & removed: init, engines:enable, engines:disable, & engines:remove are no longer supported commands.

Many of our documentation pages refer to the `engines:enable` sub-command. We should make sure to clean up those references after this goes out.

wfleming:
> are there breaking changes here? Should this be 1.0.0?

There are breaking changes, but since we're currently sub-1.0, that doesn't matter: anything goes in semver land as long as you're less than 1.0.

I don't think we're planning to call this 1.0, for product/marketing reasons.

PR #785, opened 2017-11-17
Document a couple prerequisites to releasing

toddmazierski:


DEVELOPERS.md
```diff
@@ -19,6 +19,12 @@ Otherwise, invoke the `docker run` command found in the README.
 
 ## Releasing a new version
 
+### Prerequisites
+
+1. Be an owner of the [codeclimate gem](https://rubygems.org/gems/codeclimate)
+   and authenticate with `gem push`
```
wfleming: What do you mean by "authenticate with gem push"?

toddmazierski: This was recommended by the `gem` CLI when my first release failed. If you run `gem push` without specifying a `.gem`, you're prompted for your credentials and then they're saved for later.

I could flesh that out a little bit: "authenticate with rubygems.org by running `gem push`".

wfleming: Got it.  It's fine as-is, I did not realize that was how RubyGems actually suggested logging in.

PR #797, opened 2017-12-12
Validate issue remediation points

wfleming:
We introduced a bug in duplication that incorrectly emitted negative
remediation points in some instances. We're fixing that bug, but we
should also validate issues more thoroughly to catch issues like this
earlier & prevent downstream issues.

### DIFF
--- a/lib/cc/analyzer/issue_validations/remediation_points_validation.rb
+++ b/lib/cc/analyzer/issue_validations/remediation_points_validation.rb
@@ -1,0 +1,25 @@
+module CC
+  module Analyzer
+    module IssueValidations
+      class RemediationPointsValidation < Validation
+        def valid?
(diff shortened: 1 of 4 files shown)


lib/cc/analyzer/issue_validations/remediation_points_validation.rb
```diff
@@ -0,0 +1,25 @@
+module CC
+  module Analyzer
+    module IssueValidations
+      class RemediationPointsValidation < Validation
+        def valid?
+          remediation_points.nil? || positive_integer?(remediation_points)
+        end
+
+        def message
+          "Remediation points must be a positive integer"
```
chrishulton: should we use the terminology `non-negative` here rather than `positive` since it can be zero?

wfleming: Done.

PR #806, opened 2018-01-09
Rewrite `Net::HTTP` calls to support `http_proxy`

toddmazierski:
This is one piece of support for HTTP proxies with the `fetch` directive in `.codeclimate.yml`.

`Net::HTTP` supports an `http_proxy` environment variable. It's convenient in some respects, but requires that `Net::HTTP` be directly instantiated with `#new`, as opposed to one of the alternate constructors like `.start` or `.get_response`.

For more, please see: https://www.jethrocarr.com/2014/08/14/ruby-nethttp-proxies/

### DIFF
--- a/lib/cc/cli/version_checker.rb
+++ b/lib/cc/cli/version_checker.rb
@@ -66,11 +66,13 @@ module CC
             uri = URI.parse(ENV.fetch("CODECLIMATE_VERSIONS_URL", DEFAULT_VERSIONS_URL))
             uri.query = { version: version, uid: global_config.uuid }.to_query
 
-            request = Net::HTTP::Get.new(uri, "User-Agent" => user_agent)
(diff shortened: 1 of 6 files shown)


lib/cc/cli/version_checker.rb
```diff
@@ -66,11 +66,13 @@ def http_response
             uri = URI.parse(ENV.fetch("CODECLIMATE_VERSIONS_URL", DEFAULT_VERSIONS_URL))
             uri.query = { version: version, uid: global_config.uuid }.to_query
 
-            request = Net::HTTP::Get.new(uri, "User-Agent" => user_agent)
+            http = Net::HTTP.new(uri.host, uri.port)
+            http.open_timeout = 5
```
wfleming: What led to picking these timeout values? Are there not reasonable automatic defaults? Why are they necessary here but not in `prepare`?

PR #812, opened 2018-02-11
fail with error code on issues (text only)

ahmadnassri:
Fixes: #422


ahmadnassri:
sorry for the multi commit fixes, first time writing Ruby code! please squash before merging to maintain commit health!

ahmadnassri:
bump

chrishulton:
Hi, thanks for the PR.  I've been discussing this internally, and we think that we need a larger discussion for how this should work.

We are considering whether to distinguish the CLI exiting with an error (status 1) versus exiting with issues (which should perhaps use a separate error code) as we rely on the status code throughout our infrastructure as well.  There's also the issue of what formatters to apply this behavior to (plaintext, json, html).

There is some similar discussion in a [previous PR](https://github.com/codeclimate/codeclimate/pull/508).

I'd encourage continuing to use the workaround in https://github.com/codeclimate/codeclimate/issues/422 for now.  There's no ETA for when we will get to addressing this change but it's something on our radar.

ahmadnassri:
thanks @chrishulton I've seen both the PR and issue #422 which is what prompted me to make the change and open up this PR ...

happy to hear internal discussion is ongoing, would there be a venue for us on the outside to track and contribute to the discussion? _(for example, I'm representing corp organization with 700+ repos and 300+ members, might be able to give some context from an Enterprise technology perspective)_

wilson:
I think we'd be open to accepting a patch for this, but it does need to have the attributes described by @will over on issue #422 

I'll take a swing at this next week if I can fit it in.

will:
I described no such thing!

wilson:
Sorry, too much Slack; I mean a different Will. Sorry for the ping!

wfleming:
This PR has been inactive for a while, I'm going to go ahead and close it. I earlier tried to detail both why this approach doesn't work for us and also how I think a successful approach to implementing this would work [in this comment](https://github.com/codeclimate/codeclimate/issues/422#issuecomment-385122420). I'm afraid that this functionality is still not a priority for our team, so we're not planning to change this behavior ourselves anytime soon. I recognize that deferring to external contributors on a code base that we depend on for some internal usage can lead to frustrations in cases like this where a seemingly innocuous change in the open source code won't work because of non-obvious internal requirements. I hope the comment I linked to above can alleviate that in the future by more clearly laying out the boundaries of how the modules in this project are used, and am happy to discuss further questions about this feature on the original issue #422 if anybody is interested in implementing this.

PR #841, opened 2018-06-20
Release v0.74.0

wilson:


maxjacobson:
I think I stumbled on a tiny bug while QA-ing this

```shell
codeclimate analyze --dev -e reek
```

This works before the upgrade to docker, but not after. I think what's happening is that if the CLI is in development mode, it passes `--memory ""` down to the docker CLI, and maybe that used to be allowed, but now it's not.

https://github.com/codeclimate/codeclimate/blob/d63c59db1ba202cb6148645dfbe7a0df825125ae/lib/cc/engine_registry.rb#L30-L31

Worth fixing before releasing?

wilson:
Definitely worth fixing, I'd say.
Also, this image is noticeably larger than the previous release. I think I should close this one, and take another swing at it. Sound OK?

maxjacobson:
Sounds OK to me - I don't have context on the urgency of the other work.

Do you have the sizes offhand? I'm curious how large it is. It's not ideal, but OK if the CLI grows a little IMO

wilson:
I'm going to double-check in the morning to make sure I'm comparing apples to apples.. but it looks like 0.73.0 is 92MB, and this build is 180MB.
That being said, our 'structure' image is like 3.2GB, so it's still pretty small in the grand scheme.

wilson:
Fixed the dev-mode thing over here: https://github.com/codeclimate/codeclimate/pull/843

PR #842, opened 2018-06-21
Use `alpine` instead of `ruby` base image; 100MB vs. 180MB.

wilson:
This also updates the image to Ruby v2.4.4, from v2.2.10.

```
REPOSITORY                                      TAG                 IMAGE ID            CREATED             SIZE
codeclimate/codeclimate                         alpine              e22835391cc0        5 minutes ago       108MB
codeclimate/codeclimate                         latest              670535193ffc        20 hours ago        187MB
codeclimate/codeclimate                         0.73.0              7e8f7e46e8fc        13 days ago         98.3MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,10 +21,7 @@ RUN apk --no-cache upgrade && \
       wget && \
       bundle install -j 4 && \
       apk del build-base && \
-      rm -fr /usr/share/ri && \
-      printf 'export PATH=$PATH:/usr/lib/ruby/gems/2.4.0/bin' \
-      >> /etc/profile.d/rubygems.sh && \
-      chmod +x /etc/profile.d/rubygems.sh
+      rm -fr /usr/share/ri
 
 RUN wget -q -O /tmp/docker.tgz \
     https://download.docker.com/linux/static/stable/x86_64/docker-17.12.1-ce.tgz && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,14 +1,30 @@
...
+      openssl \
+      ruby \
+      ruby-bigdecimal \
+      ruby-bundler \
+      ruby-dev \
+      wget && \
+      bundle install -j 4 && \
+      apk del build-base && \
+      rm -fr /usr/share/ri && \
+      printf 'export PATH=$PATH:/usr/lib/ruby/gems/2.4.0/bin' \
```
maxjacobson: It seems like this will be a little more challenging to maintain, since we'll need to update this line whenever the apk registry switches over to 2.5 as the ruby that gets installed when someone runs `apk add ruby` :thinking:.

Users who use the CLI will pull the image, not build it themselves, so it's not likely that this occurrence will break things for users, but it could be a bit of a headache in the future for us. Do you agree? Your POV is that it's worth it for the sake of slimming the image, or are there other reasons as well?

wilson: Actually, that part may not be necessary in the first place.
I added it because it's what the `codeclimate/docker-alpine-ruby` Dockerfile we previously inherited from was doing. (Though there the path was set to 2.2.0)

Let me remove that and run the checks again; it shouldn't actually be necessary, because we're happy with the default paths built into the binary package.

PR #843, opened 2018-06-21
Only pass `--memory` option to Docker if a memory limit is set.

wilson:
This matters on newer versions of Docker, which no longer accept:
`docker run --memory "" foo`

### DIFF
--- a/lib/cc/analyzer/engine.rb
+++ b/lib/cc/analyzer/engine.rb
@@ -85,11 +85,8 @@ module CC
           "--volume", "#{config_file.host_path}:/config.json:ro",
           "--user", "9000:9000"
         ]
-        if metadata["memory"]
-          memory = metadata["memory"].to_s
-          unless memory.empty?
-            options.concat ["--memory", memory]
-          end
+        if (memory = metadata["memory"]).present?
+          options.concat(["--memory", memory.to_s])
         end
         options
       end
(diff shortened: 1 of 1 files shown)


lib/cc/analyzer/engine.rb
```diff
@@ -74,18 +74,24 @@ def qualified_name
...
           "--volume", "#{code.host_path}:/code:ro",
           "--volume", "#{config_file.host_path}:/config.json:ro",
           "--user", "9000:9000"
         ]
+        if metadata["memory"]
+          memory = metadata["memory"].to_s
+          unless memory.empty?
+            options.concat ["--memory", memory]
+          end
+        end
```
dblandin: You might be able to simplify this a bit using activesupport. Not sure you'd need to `to_s` with `#present?`.

```ruby
if (memory = metadata["memory"]).present?
  options.concat(["--memory", memory])
end
```

wilson: Aah, I wasn't sure if it was safe to rely on ActiveSupport in that part of the code.

dblandin: Yep, I'm pretty sure it's available.

wilson: Changed it, though a `to_s` is still necessary, because the specs check that it's a String.

PR #848, opened 2018-07-02
Use `/bin/bash` shebangs where appropriate.

wilson:
This corrects a few issues pointed out by running the gem on itself:
- There was an invalid config line in the `.codeclimate.yml`
- The Gemfile wasn't alphabetized
- Several `bash` scripts were running under `/bin/sh`
- Some variables needed quoting in the bin wrapper script

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -7,9 +7,10 @@ prepare:
 plugins:
   rubocop:
     enabled: true
+    channel: rubocop-0-57
   fixme:
     enabled: true
-    exclude_paths:
+    exclude_patterns:
       - config/engines.yml
   markdownlint:
     enabled: true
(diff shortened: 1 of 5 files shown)


maxjacobson:
Drive-by question: how did you determine we needed to change from `/bin/sh` to `/bin/bash`? I know in [our styleguide](https://github.com/codeclimate/styleguide/tree/master/shell#compatability) we say:

> `/bin/sh` should be preferred to `/bin/bash` for compatability unless Bash-only features are really needed

wilson:
1) shellcheck pointed out its use of features that /bin/sh lacks
2) The use of `local` is a giveaway, as that's a bash keyword; it's a no-op in POSIX shell.

I totally agree with that comment in the handbook; in this case I figured it made sense to call bash-using things bash, and then ideally turn them into `/bin/sh` scripts as a subsequent step.

The bin/prep-release and bin/release scripts use $(< FILE) which is a bash-only feature.

wilson:
Closing this in favor of separate branches for each of these unrelated changes.

PR #851, opened 2018-07-03
Correct `shellcheck` issues in the bin wrapper script.

wilson:
See https://github.com/codeclimate/codeclimate/pull/848 for previous discussion.


wilson:
This is ready for another look, at your convenience.

PR #870, opened 2018-09-13
Update plugins manifest

dblandin:
- Remove beta channels
- Add missing stable plugins


vovimayhem:
Why did the beta channels got removed?

How can we re-enable them?

dblandin:
@vovimayhem The `beta` channel for these engines were promoted to a `stable` channel. You may explicitly target the `stable` channel within your configuration or omit the channel and your builds with use the `stable` channel by default.

vovimayhem:
Got it! Thanks!

PR #893, opened 2019-02-12
Escape Other instances source code

madrian-es:
Fixes #804

PR #896, opened 2019-03-04
Resolve #888: Remove warning for exclude_fingerprints.

### LINKED ISSUE #888: Alternative to exclude_fingerprints?
I am currently able to exclude some false positives in the `duplication` plugin/engine using `exclude_fingerprints` like this:

```yaml
---
version: "2"
checks:
  argument-count:
    config:
      threshold: 6
  file-lines:
    enabled: false
  method-lines:
    config:
      threshold: 100
  return-statements:
    enabled: false
plugins:
  duplication:
    exclude_fingerprints:
    - "876e809d86a7ba31b169ff0ccd05a2a5"
    - "281b544dfb713a0120ebba8cf7a82770"
    - "2425b8596dcd296f319e31a788af5993"
    - "658fe21f6fc5f431c8f154323fb7c521"
  fixme:
    enabled: true
    config:
      strings:
      - FIXME
      - BUG
      - HACK
exclude_patterns:
- "artifacts/"
- "build/"
- "test/"
- "**/bin/"
- "**/obj/"
```

However, upon validating the configuration it tells me that `exclude_fingerprints` is deprecated. [I can find the code](https://github.com/codeclimate/codeclimate/blob/e2f7d7f7f75c0279351ddeab1f99fd50e335406b/lib/cc/config/validation/engine_validator.rb#L90) where the warning is issued. I can see that `exclude_paths` has been replaced with `exclude_patterns`... but what is the replacement for `exclude_fingerprints` if it's deprecated? I couldn't find that in the documentation and the [Duplication engine docs don't cover it, either](https://docs.codeclimate.com/docs/duplication).

tillig:


tillig:
It appears the `ci/circleci: build_test` check is not working or otherwise never completes. #893 was submitted 20 days ago and is still waiting for completion.

maxjacobson:
Good find. Looks like, after migrating to Circle CI 2.0, the "context" for the commit status changed. Looks like tests are running, but they're sending their status as `build_test`. I believe we should update our branch protection settings to expect/require that setting, instead of `ci/circleci: build_test`. cc @wfleming

wfleming:
Thanks, @tillig, sorry for the delayed merge.

PR #901, opened 2019-03-28
Check if CODECLIMATE_DEBUG is not blank

matteeyah:
This changes the CODECLIMATE_DEBUG check to check if the environment
variable has a non whitespace character in it instead of just checking
if its set or not.

Just checking if it's set or not prevents doing things like `--env "CODECLIMATE_DEBUG=${CODECLIMATE_DEBUG:-$DEBUG}"` (since this results in `CODECLIMATE_DEBUG being set to an empty string).

Context: https://gitlab.com/gitlab-org/security-products/codequality/merge_requests/22

https://github.com/codeclimate/codeclimate/pull/902 and https://github.com/codeclimate/codeclimate/pull/903 should both be merged before this.

### DIFF
--- a/lib/cc/cli.rb
+++ b/lib/cc/cli.rb
@@ -20,7 +20,7 @@ require "cc/cli/version"
 module CC
   module CLI
     def self.debug?
-      ENV["CODECLIMATE_DEBUG"]
+      ENV["CODECLIMATE_DEBUG"].present?
     end
 
     def self.logger


matteeyah:
@wfleming You're absolutely right. I didn't think that one through 😄 Somehow I didn't notice that `active_support` is required in the file and totally forgot about `#present?`. It's also a much better choice because it conveys what the method actually does more clearly.

wfleming:
Thanks for this!

PR #902, opened 2019-03-28
Bump rubocop ruby target

matteeyah:
I changed the rubocop target ruby version and made some changes to make `rubocop` happy.

I also disabled the `Style/FrozenStringLiteralComment` cop since there's ~300 offenses and adding the comment to the files isn't as straightforward since it either results in runtime errors or incorrect behavior (like fingerprint mismatches).

This should be merged together with https://github.com/codeclimate/codeclimate/pull/903

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,7 @@
 inherit_from: base_rubocop.yml
 
 AllCops:
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.3
 
 Lint/Debugger:
   Exclude:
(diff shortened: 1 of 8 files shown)


matteeyah:
The two failing specs are `spec/cc/analyzer/issue_spec.rb` and `spec/cc/analyzer/source_fingerprint_spec.rb` because the fingerprints don't seem to match.

This is most likely related to enabling frozen string literal comments.

matteeyah:
I removed the frozen string literal comment related changes and opened an issue to follow it up separately - https://github.com/codeclimate/codeclimate/issues/904

wfleming:
Thanks @matteeyah

PR #907, opened 2019-03-28
Move to ruby 2.5

wfleming:
The recent PRs around ruby version reminded me that 2.3 is actually very
near it's end-of-life. 2.3 will no longer be supported for security
patches after the end of this month, so as long as we're here this seems
like a good moment to upgrade to 2.5.

(Also 2.3 is old enough that it's not building easily on my system,
making it tricky to cut a release, and I'd rather upgrade ruby for the
CLI than figure out how to build an almost-unsupported old version of
Ruby on my machine.)


ale7714:
where are the CC issues coming from 👀 ?

wfleming:
it looks likeruby learned a `do ... rescue ... ensure` form in 2.5 so you don't need a `begin`. Similar to how you could do `def foo ... rescue .. ensure` for a long time. That's handy.

PR #922, opened 2019-08-12
Add rubocop-0-72, rubocop-0-73 and rubocop-0-74 channels

filipesperandio:


filipesperandio:
@ale7714 Will need a gem owner (or to become one) to release when this is merged.

PR #933, opened 2020-01-08
Re-add Pylint to the CLI and promote to stable

larkinscott:
We previously removed it in error. Promoting it to stable while I'm at it.


larkinscott:
@filipesperandio I think some people were trying to use the beta. They mentioned it and that's how we figured out we removed it by accident. Just keeping it for now to make sure we don't break anything and then remove it shortly after.

PR #980, opened 2020-09-18
update description for install:engines

efueger:
Adds description to include that `install:engines` also checks for (and downloads) any new images for engines.


filipesperandio (approved):
Nice!

PR #988, opened 2020-10-16
Freeze string literals

### LINKED ISSUE #904: Freeze all strings
Moving forward `codeclimate` should probably slowly and iteratively enable the `# frozen_string_literal: true` magic comment throughout the source files. There's currently several uses of strings in a mutable way.

After this is done, the `Style/FrozenStringLiteralComment` cop should be enabled.

artfuldodger:
fixes #904

### DIFF
--- a/lib/cc/analyzer/location_description.rb
+++ b/lib/cc/analyzer/location_description.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 module CC
   module Analyzer
     class LocationDescription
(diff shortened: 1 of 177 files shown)


artfuldodger:
This this is ready for review now - sorry for not marking it as a draft while I worked through the various violations.

One thing I'm unsure of is where to find `base_rubocop.yml` to make sure the rule can be removed entirely from `.rubocop.yml`

filipesperandio:
Have you tried `RUBYOPT="--enable-frozen-string-literal"` here by any chance? Wondering if our dependecies are safe for this or not.

To get the base rubcop: `codeclimate prepare`

artfuldodger:
When running `RUBYOPT="--enable-frozen-string-literal" rspec`, everything fails when trying to `require "uuid"`:

```
An error occurred while loading ./spec/cc/workspace/path_tree_spec.rb.
Failure/Error: require "uuid"

NoMethodError:
  undefined method `version' for nil:NilClass
# ./lib/cc/cli/global_config.rb:4:in `<top (required)>'
# ./lib/cc/cli/version_checker.rb:3:in `<top (required)>'
# ./lib/cc/cli/runner.rb:5:in `<top (required)>'
# ./lib/cc/cli.rb:18:in `<top (required)>'
# ./spec/spec_helper.rb:9:in `<top (required)>'
# ./spec/cc/workspace/path_tree_spec.rb:3:in `<top (required)>'
Invalid gemspec in [/Users/jon/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/uuid-2.3.9/uuid.gemspec]: can't modify frozen String
```

lib/cc/analyzer/location_description.rb
```diff
@@ -8,23 +10,30 @@ def initialize(source_buffer, location, suffix = "")
...
+          lines,
+          lines.blank? ? "" : suffix,
+        ].join
       end
 
       private
 
       attr_reader :location, :suffix
 
+      def lines
```
filipesperandio, line 23: This PR is already huge just by adding the _magic comment_, can we move any refactor to their own PR to mitigate risk?

artfuldodger: The only refactoring that was done was in order to stop modifying string literals - this change is necessary in order to add the magic comment.

filipesperandio:
Sorry for being quiet for too long. While we like marking strings with freeze, I think there is an unnecessary risk here this being such a big PR. I'd be ok doing this a few files at time as part of other work but the big change doesn't seem worth it, closing.

PR #1013, opened 2021-04-19
Set deploy pipeline

fede-moya:
Add pipeline on _circleci_ to publish new versions when `VERSION` has been modified on `master`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,6 @@ init: &init
     name: init
     command: |
       echo '. .circleci/shared.bash' >> "$BASH_ENV"
-      . .circleci/shared.bash
 
 jobs:
   build_test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,12 @@
 version: 2
+
+init: &init
+  run:
+    name: init
+    command: |
+      echo '. .circleci/shared.bash' >> "$BASH_ENV"
+      . .circleci/shared.bash
+
 jobs:
   build_test:
     environment:
@@ -14,11 +22,39 @@ jobs:
       - run: make image
       - run: make citest
 
+  publish:
+    machine: true
+    steps:
+      - checkout
+      - *init
+      - run:
+          name: Install Hub dependency
+          command: install_hub
+      - run:
+          name: Login on Dockerhub
+          command: login_to_dockerhub
+      - run:
+          name: Login on RubyGems
+          command: login_to_rubygems
+      - run:
+          name: Publish new version
+          command: |
+            if [ `git diff --quiet HEAD~ VERSION; echo $?` -eq 1 ]; then
+              publish_new_version
+            fi
+
 workflows:
   version: 2
-  build_test:
+  build_publish:
     jobs:
       - build_test
+      - publish:
+          requires:
+            - build_test
+          filters:
+            branches:
+              only:
+                - master
 
 notify:
   webhooks:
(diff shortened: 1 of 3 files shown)


fede-moya:
@filipesperandio you will notice that this differs a little bit from what we agreed on our latest conversation. The approach that. we discussed at that time consisted on releasing on each commit over `master` but it had the pitfall of having to commit version files like the following
```sh
0.85.x
```

What I ended up not liking about that is that considering that the `gemspec` file reads that file to infer the version I was kind of forced to inject some logic across different places. And also we would end with that file with an `x` committed on `master` which I kind of didn't like. 

Let me know, what do you think about this approach.

README.md
```diff
@@ -178,6 +178,29 @@ The following is a brief explanation of each available command.
...
+  * Push new image with latest version as tag.
+
+Ideally someone will open a pull request against master updating only
+[VERSION](https://github.com/codeclimate/codeclimate/blob/master/VERSION).
+
+There is script in place, which assumes [hub](https://hub.github.com/) is installed,
+to facilitate that. Just run;
+
+```sh
+./bin/prep-release VERSION
```
filipesperandio, line 201:
```suggestion
./bin/prep-release <VERSION>
```

README.md
```diff
@@ -178,6 +178,29 @@ The following is a brief explanation of each available command.
...
+1. Create a new release on Github and an associated tag.
+1. Update docker images:
+  * Push new `latest` image.
+  * Push new image with latest version as tag.
+
+Ideally someone will open a pull request against master updating only
+[VERSION](https://github.com/codeclimate/codeclimate/blob/master/VERSION).
+
+There is script in place, which assumes [hub](https://hub.github.com/) is installed,
+to facilitate that. Just run;
```
filipesperandio, line 198:
```suggestion
to facilitate that. Check the current VERSION (`cat VERSION`) and upgrade accordingly running:
```

.circleci/config.yml
```diff
@@ -1,4 +1,12 @@
 version: 2
+
+init: &init
+  run:
+    name: init
+    command: |
+      echo '. .circleci/shared.bash' >> "$BASH_ENV"
+      . .circleci/shared.bash
```
filipesperandio, line 8: Do we need both calls? I understand the first is how Circle loads those into its context, right?

fede-moya: > Do we need both calls? 
I don't think so. 

>  I understand the first is how Circle loads those into its context, right?
yeap.

Removing.

fede-moya:
Hey @filipesperandio ! 

I have just opened a PR creating a manual workflow in the `hombrew-formulae` repo for triggering the release. I have also added a step in this pipeline to dispatch that workflow. 

https://github.com/codeclimate/homebrew-formulae/pull/31

.circleci/config.yml
```diff
@@ -14,11 +21,39 @@ jobs:
...
+      - run:
+          name: Login on Dockerhub
+          command: login_to_dockerhub
+      - run:
+          name: Login on RubyGems
+          command: login_to_rubygems
+      - run:
+          name: Publish new version
+          command: |
+            if [ `git diff --quiet HEAD~ VERSION; echo $?` -eq 1 ]; then
```
filipesperandio, line 41: Can we extract this to something we can name? Trying to make it super easy to whoever is reading these steps.

filipesperandio (approved):

.circleci/config.yml
fede-moya: Believe me I tried. I tried creating a function for it, but it's wasn't useful it's not easy to pass the last returned code as a value. The only way is to store it in a variable but it's not an option at this step on circleci.

filipesperandio (approved):

PR #1015, opened 2021-04-20
Minor update on activesupport

fede-moya:
Address [CVE-2020-8165](https://groups.google.com/g/rubyonrails-security/c/bv6fW4S0Y1c)

### DIFF
--- a/codeclimate.gemspec
+++ b/codeclimate.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "~> 2.5"
 
-  s.add_dependency "activesupport", ">= 5.2"
+  s.add_dependency "activesupport", "~> 5.2.3"
   s.add_dependency "tty-spinner", "~> 0.1.0"
   s.add_dependency "highline", "~> 1.7", ">= 1.7.2"
   s.add_dependency "posix-spawn", "~> 0.3", ">= 0.3.11"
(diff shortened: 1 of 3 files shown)


codeclimate.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "~> 2.5"
 
-  s.add_dependency "activesupport", ">= 5.1.3", "< 5.2"
+  s.add_dependency "activesupport", ">= 5.2.4.3", ">= 5.2"
```
filipesperandio, line 21: Can this be simplified?
```suggestion
  s.add_dependency "activesupport", ">= 5.2"
```
Do we have any comments about why it was previously explicitly keeping it < 5.2?

fede-moya: yeap, done ! 👍🏼

codeclimate.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "~> 2.5"
 
-  s.add_dependency "activesupport", ">= 5.1.3", "< 5.2"
+  s.add_dependency "activesupport", ">= 5.2"
```
filipesperandio, line 21: sorry to bring this again, I think this should use the `Pessimistic Version Operator` (`~>`) like the others below.
Maybe something like `~> 5.2.4`, allowing it to apply new patches but without jumping minor versions, that might be problematic with activesupport I believe. WDYT?

fede-moya: Actually that sounds like a clever idea. Done.

> sorry to bring this again
No, thanks for always find a way to improve the work.

frnico33 (approved):
LGTM!

PR #1018, opened 2021-04-28
Rename workflow

fede-moya:
Open pull requests are getting hang waiting for required `build_test` workflow. The workflow has been recently renamed to `build_publish` since it includes a releasing step. We need someone with admin access over the repo to re-configure that required build in the repo settings to be `build_publish` until we get that done we can keep things moving by renaming the `build_publish` version to it's previous state.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


frnico33 (approved):
LGTM!

PR #1019, opened 2021-04-28
Add Reek channel version 6.0.4

frnico33:
Add Reek channel version 6.0.4


filipesperandio (commented):
Hey! Just made a comment here, I'd like us to not follow the somewhat pattern we setup in Rubocop supporting multiple versions for any other plugin. So, I'd suggest we update submissions with a new beta. That will mean we don't even have to touch this repo, only the builder.

https://codeclimate.atlassian.net/browse/QUA-217?focusedCommentId=28828

PR #1047, opened 2022-06-02
QUA-580: Don't analyze files fetched in the prepare step

dantevvp:
This PR tells the Analyzer to ignore all files fetched in the `prepare` step of the build.

It does so by going through the prepare config in the `.codeclimate.yml` file and adding the paths to the `exclude_patterns` setting programmatically. This is done so no analysis is run on these files, and they don't appear in the `Code` tab in the user's dashboard.

Example:

In a project with this configuration:

image

The fetched files are saved in the codebase:

image

And they don't appear in the Code tab:

image

https://codeclimate.atlassian.net/browse/QUA-580

### DIFF
--- a/lib/cc/config.rb
+++ b/lib/cc/config.rb
@@ -41,10 +41,13 @@ module CC
     end
 
     def self.build(data)
+      prepare_data = Prepare.from_data(data["prepare"])
+      prepare_paths = prepare_data.fetch.entries.map(&:path)
+      exclude_patterns_data = data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS) + prepare_paths
       new(
         engines: EngineSet.new(data.fetch("plugins", {})).engines,
-        exclude_patterns: data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS),
-        prepare: Prepare.from_data(data["prepare"]),
+        exclude_patterns: exclude_patterns_data,
+        prepare: prepare_data,
       )
     end
 
--- a/lib/cc/config.rb
+++ b/lib/cc/config.rb
@@ -40,13 +40,12 @@ module CC
     end
 
     def self.build(data)
-      prepare_data = Prepare.from_data(data["prepare"])
-      prepare_paths = prepare_data.fetch.map(&:path)
-      exclude_patterns_data = data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS) + prepare_paths
+      prepare = Prepare.from_data(data["prepare"])
+      base_excluded_patterns = data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS)
       new(
         engines: EngineSet.new(data.fetch("plugins", {})).engines,
-        exclude_patterns: exclude_patterns_data,
-        prepare: prepare_data,
+        exclude_patterns: base_excluded_patterns + prepare.fetch.paths,
+        prepare: prepare,
       )
     end
 
--- a/lib/cc/config.rb
+++ b/lib/cc/config.rb
@@ -42,7 +42,7 @@ module CC
 
     def self.build(data)
       prepare_data = Prepare.from_data(data["prepare"])
-      prepare_paths = prepare_data.fetch.entries.map(&:path)
+      prepare_paths = prepare_data.fetch.map(&:path)
       exclude_patterns_data = data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS) + prepare_paths
       new(
         engines: EngineSet.new(data.fetch("plugins", {})).engines,
--- a/lib/cc/config.rb
+++ b/lib/cc/config.rb
@@ -23,7 +23,8 @@ module CC
       :prepare
 
     attr_writer \
-      :development
+      :development,
+      :exclude_patterns
 
     def self.load
       config =
--- a/lib/cc/config.rb
+++ b/lib/cc/config.rb
@@ -23,8 +23,7 @@ module CC
       :prepare
 
     attr_writer \
-      :development,
-      :exclude_patterns
+      :development
 
     def self.load
       config =
--- a/lib/cc/config/prepare.rb
+++ b/lib/cc/config/prepare.rb
@@ -34,8 +34,8 @@ module CC
           entries.each(&block)
         end
 
-        def map(&block)
-          entries.map(&block)
+        def paths
+          entries.map(&:path)
         end
 
         def merge(other)
--- a/lib/cc/config/prepare.rb
+++ b/lib/cc/config/prepare.rb
@@ -34,6 +34,10 @@ module CC
           entries.each(&block)
         end
 
+        def map(&block)
+          entries.map(&block)
+        end
+
         def merge(other)
           Fetch.new(each.to_a | other.each.to_a)
         end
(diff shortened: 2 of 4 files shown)


fede-moya:
@dantevvp food for thought - what would happen with a config like this one ?
```
version: "2"
prepare:
  fetch:
    - url: "https://raw.githubusercontent.com/example/example/master/rubocop.yml"
```
Notice it's missing the `path` directive, which is used by your changes to determine where is it that files fetched by the prepare step reside.

fede-moya:
Update on,

> @dantevvp food for thought - what would happen with a config like this one ?
> ```
> version: "2"
> prepare:
>   fetch:
>     - url: "https://raw.githubusercontent.com/example/example/master/rubocop.yml"
> ```
> 
> 
> Notice it's missing the `path` directive, which is used by your changes to determine where is it that files fetched by the prepare step reside.

I tracked this down to https://github.com/codeclimate/codeclimate/blob/ef5757c94c35a10b12c91c8fec70ea262d21de19/lib/cc/cli/prepare.rb#L65, which ends in https://github.com/codeclimate/codeclimate/blob/ef5757c94c35a10b12c91c8fec70ea262d21de19/lib/cc/cli/prepare.rb#L74  which will blow if there is no file name given, so I guess we can't forget about this edge case.

fede-moya:
What I'm struggling to determine is who is responsible for this, the config itself or the entity that consumes the config ?

The config class has some clear attributes;
```ruby
module CC
  class Config
    attr_reader :analysis_paths, :engines, :exclude_patterns, :prepare
  end
end
```

Currently the `prepare` config represents a list of things to be fetch from the internet to be used by the some engine. Potentially the `prepare` step might involve some other activities besides merging stuff. 

So my question is, should the `Config` class be responsible for knowing that the stuff listed on the `prepare` should be ignore for analysis or is the runner who should determine that ? I believe in this case this responsibility goes down to `CC::Analyzer::Bridge` but I'm not seeing it crystal clear.

dantevvp:
> @dantevvp food for thought - what would happen with a config like this one ?
> 
> ```
> version: "2"
> prepare:
>   fetch:
>     - url: "https://raw.githubusercontent.com/example/example/master/rubocop.yml"
> ```
> 
> Notice it's missing the `path` directive, which is used by your changes to determine where is it that files fetched by the prepare step reside.

In this case, I believe that the Entry class inside the Fetch class would be initialized with `:path` being equal to the last component of the `:url`. So that would make the path equal to `rubocop.yml`. If somehow the path is still null after that operation, then an exception would be thrown. This can be seen in this part of the code:

https://github.com/codeclimate/codeclimate/blob/d976a103f000006574428266ba480a04b3a4678f/lib/cc/config/prepare.rb#L49-L81

dantevvp:
> Update on,
> 
> > @dantevvp food for thought - what would happen with a config like this one ?
> > ```
> > version: "2"
> > prepare:
> >   fetch:
> >     - url: "https://raw.githubusercontent.com/example/example/master/rubocop.yml"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Notice it's missing the `path` directive, which is used by your changes to determine where is it that files fetched by the prepare step reside.
> 
> I tracked this down to
> 
> https://github.com/codeclimate/codeclimate/blob/ef5757c94c35a10b12c91c8fec70ea262d21de19/lib/cc/cli/prepare.rb#L65
> 
> , which ends in
> https://github.com/codeclimate/codeclimate/blob/ef5757c94c35a10b12c91c8fec70ea262d21de19/lib/cc/cli/prepare.rb#L74
> 
> which will blow if there is no file name given, so I guess we can't forget about this edge case.

The config file in the gem throws an error if no path is given, but even if the path is null, a Fetch::Entry is initialized using the url that is passed to it, so in this case, the path attribute would be equal to rubocop.yml.

camillof (approved):
LGTM! 🧙‍♂️ !

lib/cc/config/prepare.rb
```diff
@@ -34,6 +34,10 @@ def each(&block)
           entries.each(&block)
         end
 
+        def map(&block)
+          entries.map(&block)
+        end
```
fede-moya, line 39: maybe;
 ```ruby
def paths
  entries.map(&:path)
end
```

File: lib/cc/config.rb [hunk 2]
```diff
@@ -40,10 +40,13 @@ def self.load
...
+      prepare_data = Prepare.from_data(data["prepare"])
+      prepare_paths = prepare_data.fetch.map(&:path)
+      exclude_patterns_data = data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS) + prepare_paths
       new(
         engines: EngineSet.new(data.fetch("plugins", {})).engines,
-        exclude_patterns: data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS),
-        prepare: Prepare.from_data(data["prepare"]),
+        exclude_patterns: exclude_patterns_data,
+        prepare: prepare_data,
       )
```
fede-moya, line 50: wdty about ?
```ruby
def self.build(data)
  prepare = Prepare.from_data(data["prepare"])
  new(
    engines: EngineSet.new(data.fetch("plugins", {})).engines,
    exclude_patterns: data.fetch("exclude_patterns", DefaultAdapter::EXCLUDE_PATTERNS) + prepare.fetch.paths,
    prepare: prepare,
  )
end
```

spec/cc/config_spec.rb
```diff
@@ -104,6 +104,32 @@
...
+      EOYAML
+
+      config = CC::Config.load
+
+      expect(config.exclude_patterns).to include("**/*.rb")
+      expect(config.exclude_patterns).to include("foo/")
+      expect(config.exclude_patterns).to include("rubocop.yml")
+      expect(config.exclude_patterns).to include("ignore/this.json")
+      expect(config.exclude_patterns).to include("nopath.yml")
+      expect(config.prepare.fetch.map(&:path)).to eq(["rubocop.yml", "ignore/this.json", "nopath.yml"])
```
fede-moya, line 130:
```suggestion
      expect(config.prepare.fetch.paths).to eq(["rubocop.yml", "ignore/this.json", "nopath.yml"])
```

spec/cc/config_spec.rb
```diff
@@ -104,6 +104,32 @@
...
+      - foo/
+      EOYAML
+
+      config = CC::Config.load
+
+      expect(config.exclude_patterns).to include("**/*.rb")
+      expect(config.exclude_patterns).to include("foo/")
+      expect(config.exclude_patterns).to include("rubocop.yml")
+      expect(config.exclude_patterns).to include("ignore/this.json")
+      expect(config.exclude_patterns).to include("nopath.yml")
```
fede-moya, line 129:
```suggestion
expect(config.exclude_patterns).to include(["**/*.rb", "foo/", "rubocop.yml", "ignore/this.json", "nopath.yml"])
```

spec/cc/config_spec.rb
```diff
@@ -104,6 +104,32 @@
       expect(config.prepare.fetch.each.to_a).to eq([CC::Config::Prepare::Fetch::Entry.new("rubocop.yml")])
     end
 
+    it "adds files fetched in prepare step to exclude_patterns" do
+      yaml = write_cc_yaml(<<-EOYAML)
+      prepare:
+        fetch:
```
fede-moya, line 110: can we add a similar spec where there is no prepare step give ?

fede-moya (approved):
8hcE

PR #1051, opened 2022-06-23
QUA-526: Upgrade base image and Ruby version

camillof:
### Context

We need to upgrade the docker base image, as `alpine:3.11.6` is no longer supported

### Changes
- Upgraded to alpine:3.16.0 which comes with Ruby 3.1.2
- Kept the compatibility with previous ruby version, so it can be consumed from systems that weren't upgraded yet.
- Using keywords arguments as required by Ruby 3, while keeping compatibility with older version

image

No vulnerabilities found after running `docker scan`

NOTE: it Is also possible to keep this running on ruby 2.6 if we don't want to upgrade to ruby 3. Please see [first commit](https://github.com/codeclimate/codeclimate/pull/1051/commits/d9aaafe999d9dd7bc9f24acdc6d6276b93abbaa1). But it won't fix all vulnerabilities

Linked issue: https://codeclimate.atlassian.net/browse/QUA-626


camillof:
> @camillof We might need to update the Rubocop config as well? https://codeclimate.com/github/codeclimate/codeclimate/pull/1051

Yep, missed that, updated rubocop channel as well, as the previous one didn't support Ruby 3.1. 

Thanks!

lib/cc/cli/file_store.rb
```diff
@@ -31,7 +31,7 @@ def load_data
         @data =
           if File.exist? self.class::FILE_NAME
             File.open(self.class::FILE_NAME, "r:bom|utf-8") do |f|
-              YAML.safe_load(f, [Time], [], false, self.class::FILE_NAME) || {}
+              YAML.safe_load(f, permitted_classes: [Time], permitted_symbols: [], aliases: false, filename: self.class::FILE_NAME)  || {}
```
fede-moya, line 34: curious seeing this `safe_load` twice in the same week...

.ruby-version
```diff
@@ -1 +1 @@
-2.6.6
+3.1.2
```
fede-moya, line 1: 👍🏼

.codeclimate.yml
```diff
@@ -9,7 +9,7 @@ plugins:
     enabled: true
   rubocop:
     enabled: true
-    channel: rubocop-0-92
+    channel: rubocop-1-30
```
fede-moya, line 12: 👍🏼

f-moya (commented):
I would strongly advise to run some tests using ruby 2.7 what might help us to identify some places that might be needing updates in order to support Ruby 3. I'm particularly concern about the dependencies on runtime. 

So I will be first confirming current dependencies support ruby 3;
```
  s.add_dependency "activesupport", "~> 5.2.3"
  s.add_dependency "tty-spinner", "~> 0.1.0"
  s.add_dependency "highline", "~> 1.7", ">= 1.7.2"
  s.add_dependency "posix-spawn", "~> 0.3", ">= 0.3.11"
  s.add_dependency "pry", "~> 0.10.1"
  s.add_dependency "rainbow", "~> 2.0", ">= 2.0.0"
  s.add_dependency "redcarpet", "~> 3.2"
  s.add_dependency "uuid", "~> 2.3"

```

camillof:
> I would strongly advise to run some tests using ruby 2.7 what might help us to identify some places that might be needing updates in order to support Ruby 3. I'm particularly concern about the dependencies on runtime.
> 
> So I will be first confirming current dependencies support ruby 3;
> 
> ```
>   s.add_dependency "activesupport", "~> 5.2.3"
>   s.add_dependency "tty-spinner", "~> 0.1.0"
>   s.add_dependency "highline", "~> 1.7", ">= 1.7.2"
>   s.add_dependency "posix-spawn", "~> 0.3", ">= 0.3.11"
>   s.add_dependency "pry", "~> 0.10.1"
>   s.add_dependency "rainbow", "~> 2.0", ">= 2.0.0"
>   s.add_dependency "redcarpet", "~> 3.2"
>   s.add_dependency "uuid", "~> 2.3"
> ```

Yep, I spent some time reading documentation of each dependency, and most of them are locked to Ruby > 2.x (so no constraint about ruby 3). I did update highline which didn't work on the previous version.

The only one I found an open issue is `posix-spawn`: https://github.com/rtomayko/posix-spawn/issues/90, and reading the comments on that nice discussion, you may find that we could even ditch this dependency. Wdyt? Should that be done on this PR?

I'll run tests on ruby 2.7 as well, thanks for that idea!

camillof:
Moving this to draft, doing some extra tests I noticed some errors that are probably related to the `posix-spawn` gem

fede-moya:
> The only one I found an open issue is posix-spawn: https://github.com/rtomayko/posix-spawn/issues/90, and reading the comments on that nice discussion, you may find that we could even ditch this dependency. Wdyt? Should that be done on this PR?

Letting go `posix-spawn` seems like a clever move yeah. It's great to say goodbye to another dependency. Regarding if it's a good call to do it know or for this upgrade ?, it's a small enough change I would say let's do it. We test and rollback fast if needed.

camillof:
> Letting go posix-spawn seems like a clever move yeah. It's great to say goodbye to another dependency. Regarding if it's a good call to do it now or for this upgrade ?, it's a small enough change I would say let's do it. We test and rollback fast if needed.

Yep, sorry I didn't updated my comment, but if we intend to upgrade the Ruby version to >3.X we _need_ to drop this dependency, as I faced the issue described on the link too. (I didn't on my first test, but I did later while testing on a larger codebase)

lib/cc/analyzer/container.rb
```diff
@@ -178,9 +176,11 @@ def kill_wait_thread
 
       def reap_running_container(message)
         Analyzer.logger.warn("killing container name=#{@name} message=#{message.inspect}")
-        POSIX::Spawn::Child.new("docker", "kill", @name, timeout: 2.minutes)
-        POSIX::Spawn::Child.new("docker", "wait", @name, timeout: 2.minutes)
-      rescue POSIX::Spawn::TimeoutExceeded
+        Timeout.timeout(2.minutes.to_i) do
+          Kernel.system("docker kill #{@name}")
+          Kernel.system("docker wait #{@name}")
```
fede-moya, line 181:
```suggestion
          system("docker", "kill", @name)
          system("docker", "wait", @name)
```
This way you avoid the shell which is generally preferred. See https://apidock.com/ruby/Kernel/system

f-moya (commented):
Looks good @camillof minor comment on passing args to `Kernel.system` that I think worth addressing

lib/cc/analyzer/container.rb
```diff
@@ -52,19 +52,14 @@ def run(options = [])
 
         command = docker_run_command(options)
         Analyzer.logger.debug("docker run: #{command.inspect}")
-        pid, _, out, err = POSIX::Spawn.popen4(*command)
+        _, out, err, @t_wait = Open3.popen3(*command)
```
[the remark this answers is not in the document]
camillof: You may be right, we just want the status. Not sure if that would allow us to kill that process tough: https://github.com/codeclimate/codeclimate/blob/974c82b95e820b7d4b58df02e46b30019746ca98/lib/cc/analyzer/container.rb#L174

lib/cc/analyzer/container.rb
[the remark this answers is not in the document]
camillof: I can imagine that originally this was because the `docker kill` command wasn't a blocking one? so they had to `wait` for the container to exit.

As today, those two line together don't make sense, as the `kill` command is blocking.

Again, it's a piece of code that I don't want to change in this already long PR

lib/cc/analyzer/container.rb
```diff
@@ -52,19 +52,14 @@ def run(options = [])
...
         @t_out = read_stdout(out)
         @t_err = read_stderr(err)
         t_timeout = timeout_thread
 
-        # blocks until the engine stops. this is put in a thread so that we can
-        # explicitly abort it as part of #stop. otherwise a run-away container
-        # could still block here forever if the docker-kill/wait is not
-        # successful. there may still be stdout in flight if it was being
-        # produced more quickly than consumed.
-        @t_wait = Thread.new { _, @status = Process.waitpid2(pid) }
```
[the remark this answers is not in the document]
camillof: Yep, you're right, we're now using the directly the thread that the `Open3.popen3` method returns. There is no need of wrapping it into a new thread like we did before, where we had a process pid

f-moya (approved):

larkinscott (approved):
This LGTM. Thanks for leaving those comments along the way!

PR #1059, opened 2022-08-03
QUA-567: Update Docker version from 17.12.1 to 20.10.17

### LINKED ISSUE #1053: Dockerimage - 0.86.0 "No such file or directory"
Hi, the :0.86.0 docker-image appears broke
 `error: (Errno::ENOENT) No such file or directory - docker`

### Used via Gitlab-CI so dont have full details,

Notably: this "docker not found" is NOT the 'docker run' call, but coming from inside the image 
(as seen on the "pull" 2 lines above working)
```
$ docker pull codeclimate/codeclimate
Using default tag: latest
latest: Pulling from codeclimate/codeclimate
2408cc74d12b: Pulling fs layer
4d8a1bad31d1: Pulling fs layer
...
$ docker run {{options omitted}} codeclimate/codeclimate analyze
error: (Errno::ENOENT) No such file or directory - docker
```

everything was working yesterday

camillof:
Updates used Docker version from 17.12.1 to ~~20.10.9~~ 20.10.17.  May fix #1053 

Related issue: https://codeclimate.atlassian.net/browse/QUA-657

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN apk --no-cache upgrade && \
       rm -fr /usr/share/ri
 
 RUN wget -q -O /tmp/docker.tgz \
-    https://download.docker.com/linux/static/stable/x86_64/docker-17.12.1-ce.tgz && \
+    https://download.docker.com/linux/static/stable/x86_64/docker-20.10.9.tgz && \
     tar -C /tmp -xzvf /tmp/docker.tgz && \
     mv /tmp/docker/docker /bin/docker && \
     chmod +x /bin/docker && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN apk --no-cache upgrade && \
       rm -fr /usr/share/ri
 
 RUN wget -q -O /tmp/docker.tgz \
-    https://download.docker.com/linux/static/stable/x86_64/docker-20.10.9.tgz && \
+    https://download.docker.com/linux/static/stable/x86_64/docker-20.10.17.tgz && \
     tar -C /tmp -xzvf /tmp/docker.tgz && \
     mv /tmp/docker/docker /bin/docker && \
     chmod +x /bin/docker && \


dantevvp:
I see that there's a more recent docker version, [Docker 20.10.17](https://download.docker.com/linux/static/stable/x86_64/docker-20.10.17.tgz).

Would it be a good idea to update to the latest version or is there a specific reason for this particular one?

Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN apk --no-cache upgrade && \
       rm -fr /usr/share/ri
 
 RUN wget -q -O /tmp/docker.tgz \
-    https://download.docker.com/linux/static/stable/x86_64/docker-17.12.1-ce.tgz && \
+    https://download.docker.com/linux/static/stable/x86_64/docker-20.10.9.tgz && \
```
fede-moya, line 27: why we missed the `-ce` (community edition) part ? not offered any more ?

f-moya (commented):
Hey @camillof , this project depends heavily on the docker api. Even though Docker's API it's stable over time, I'm surprised we haven't been affected by any change. Have you run some tests locally ?

I pull the branch and run some tests but everything looks normal.

camillof:
> I see that there's a more recent docker version, [Docker 20.10.17](https://download.docker.com/linux/static/stable/x86_64/docker-20.10.17.tgz).
> 
> Would it be a good idea to update to the latest version or is there a specific reason for this particular one?

Good catch @dantevvp , that weird ordering on the list got me. Will update to that version, thanks!

> Hey @camillof , this project depends heavily on the docker api. Even though Docker's API it's stable over time, I'm surprised we haven't been affected by any change. Have you run some tests locally ?
> 
> I pull the branch and run some tests but everything looks normal.

@fede-moya  To be honest, I was also surprised to not find any error on this large version bump. I did run some tests locally:
- Used the CLI directly
- Used it through the manual docker invocation
- Used it through the builder on the app

Mostly on our `app` project. But I think it would benefit all of us if you could run test locally as well.

One of the things that raised concerns, was the deprecation of `docker-machine` https://github.com/docker/machine , that we use here https://github.com/codeclimate/codeclimate/blob/b31699702ddc36594de44543d59dc9c8bd6c659f/codeclimate-wrapper#L167. 

The others commands used on the code look pretty standard (run, pull, kill, stop, etc.). But It also surprised me that even their output could have changed and for some specific cases we rely on it

Dockerfile
camillof: Yep, not offered anymore, here is the list: https://download.docker.com/linux/static/stable/x86_64/

I searched for a changelog explaining those naming change, but couldn't find a reason, just the dialog box here: https://docs.docker.com/engine/release-notes/18.09/

f-moya (approved):
I run `make install` then I tried the CLI on different repos and looks good 👍🏼

PR #1071, opened 2022-11-15
QUA-832: Do not use Ruby 3.1 hash syntax

camillof:
There are some clients of this gem, that don't use ruby 3.1 version, therefore they fail to interpret the hash syntax.


RubyBrewsday (approved):
sweet, yeah this was a silly bug to run into but glad we caught it before merging into builder :)

PR #1073, opened 2022-11-16
add brakeman beta channel

RubyBrewsday:
* Brakeman beta will point to version 5.3.1

Here's an example of what the output looked like when I ran it against app:
```
➜  app git:(master) codeclimate analyze -e brakeman:beta app/models/user.rb
Starting analysis
Running brakeman: Done!

== Gemfile (1 issue) ==
1: Support for Rails 5.2.4.5 ended on 2022-06-01 [brakeman]

Analysis complete! Found 1 issue.
```


fede-moya (commented):
Mike, here we reference images using names and not the build number. Don't take my work check `engines.yml`

RubyBrewsday:
@fede-moya ah good catch here, appreciate that/this makes sense especially as we introduce new versions to the various channels. This was me just being a copypaster from builder :)

fede-moya (approved):

PR #1074, opened 2022-11-18
Release v0.87.4

RubyBrewsday:
* Add beta channel for Brakeman that uses 5.3.1


Gemfile.lock
```diff
@@ -24,13 +24,13 @@ GEM
     diff-lcs (1.3)
     docile (1.1.5)
     highline (2.0.3)
-    i18n (1.11.0)
+    i18n (1.12.0)
```
camillof, line 27: Just in case, I double checked this and looks safe https://github.com/ruby-i18n/i18n/releases

camillof (approved):

PR #1067, opened 2022-11-02
QUA-832: update statsd to tag engine metrics

camillof:
~## ⚠️  DO NOT MERGE UNTIL https://github.com/codeclimate/builder/pull/1974 IS MERGED~

On https://github.com/codeclimate/builder/pull/1974 we updated our statsd "host" gem, which now allows us to make use of [custom tags](https://docs.datadoghq.com/metrics/custom_metrics/dogstatsd_metrics_submission/?tab=ruby#metric-tagging).

Having this gives us the opportunity of having more granularity over our metrics dashboards and controls.

PR #1085, opened 2022-12-20
QUA-870: metrics engines names cleanup and add repo_id

camillof:
### Context

On #1067 we started tagging the DD metrics, including the engines names on tags. There is no need anymore of including the engine name on the metric name itself.

Also, having the `repo_id` available as a tag, is useful for monitoring.

Screenshot 2022-12-20 at 14 29 28

For reference, these are the changes that would make use of this functionality -> https://github.com/codeclimate/builder/pull/1986

### DIFF
--- a/lib/cc/analyzer/statsd_container_listener.rb
+++ b/lib/cc/analyzer/statsd_container_listener.rb
@@ -31,27 +31,22 @@ module CC
 
       attr_reader :statsd
 
-      def increment(engine, metric_name)
+      def increment(engine, action)
         tags = engine_tags(engine)
-        # rubocop:disable Style/HashSyntax
-        metrics(engine, metric_name).each { |metric| statsd.increment(metric, tags: tags) }
-        # rubocop:enable Style/HashSyntax
+        metric = metric_name(engine, action)
+
+        statsd.increment(metric, tags: tags)
       end
 
-      def timing(engine, metric_name, millis)
+      def timing(engine, action, millis)
         tags = engine_tags(engine)
-        # rubocop:disable Style/HashSyntax
-        metrics(engine, metric_name).each { |metric| statsd.timing(metric, millis, tags: tags) }
-        # rubocop:enable Style/HashSyntax
+        metric = metric_name(engine, action)
+
+        statsd.timing(metric, millis, tags: tags)
       end
 
-      def metrics(engine, metric_name)
-        [
-          "engines.#{metric_name}",
-          "engines.names.#{engine.name}.#{metric_name}",
-        ].tap do |metrics|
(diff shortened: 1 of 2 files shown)


lib/cc/analyzer/statsd_container_listener.rb
```diff
@@ -1,9 +1,12 @@
 module CC
   module Analyzer
     class StatsdContainerListener < ContainerListener
-      def initialize(statsd)
+      # rubocop:disable Lint/MethodMissingSuper
```
dantevvp, line 4:
```suggestion
      # rubocop:disable Lint/MissingSuper
```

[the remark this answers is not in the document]
camillof: Gotcha, thanks!! 🙌

dantevvp (approved):
LGTM!

lib/cc/analyzer/statsd_container_listener.rb
```diff
@@ -1,9 +1,12 @@
 module CC
   module Analyzer
     class StatsdContainerListener < ContainerListener
-      def initialize(statsd)
+      # rubocop:disable Lint/MissingSuper
+      def initialize(statsd, repo_id: nil)
         @statsd = statsd
+        @repo_id = repo_id
       end
```
fede-moya, line 8: Millo this feels very forced, can't we just try to read that from the environment instead ?

lib/cc/analyzer/statsd_container_listener.rb
```diff
@@ -29,34 +32,34 @@ def finished(engine, _details, result)
 
       private
 
-      attr_reader :statsd
+      attr_reader :statsd, :repo_id
 
-      def increment(engine, metric_name)
+      def increment(engine, action)
```
fede-moya, line 37: don't you think `event` may be more correct rather than `action` ?

lib/cc/analyzer/statsd_container_listener.rb
```diff
@@ -29,34 +32,34 @@ def finished(engine, _details, result)
...
 
-      def metrics(engine, metric_name)
-        [
-          "engines.#{metric_name}",
-          "engines.names.#{engine.name}.#{metric_name}",
-        ].tap do |metrics|
-          metrics << "engines.names.#{engine.name}.#{engine.channel}.#{metric_name}" if engine_channel_present?(engine)
-        end
+      def metric_name(action)
+        "engines.#{action}"
```
fede-moya, line 56: are you aware this change will brake some dashboards we have in place ?

f-moya (commented):
Hey Millo, I left some comments.

PR #1086, opened 2022-12-20
Add beta and 1-7-4 channels to bandit

dantevvp:


camillof (approved):
Validated that both tags are published: https://hub.docker.com/r/codeclimate/codeclimate-bandit/tags ✅

PR #1108, opened 2023-03-30
forgot to add the org to the image reference for brakeman-5-4-1

RubyBrewsday:
When you try to use the new channel it will say it cannot find the image, forgot that we needed to add the prefix for the org in the image.

### DIFF
--- a/config/engines.yml
+++ b/config/engines.yml
@@ -21,7 +21,7 @@ brakeman:
   channels:
     stable: codeclimate/codeclimate-brakeman
     beta: codeclimate/codeclimate-brakeman:beta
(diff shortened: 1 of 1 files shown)


config/engines.yml
```diff
@@ -21,7 +21,7 @@ brakeman:
   channels:
     stable: codeclimate/codeclimate-brakeman
     beta: codeclimate/codeclimate-brakeman:beta
-    brakeman-5-4-1: codeclimate-brakeman:brakeman-5-4-1
+    brakeman-5-4-1: codeclimate/codeclimate-brakeman:brakeman-5-4-1
```
giordanoluzardo, line 24: 👍

giordanoluzardo (approved):

PR #1148, opened 2024-11-18
Simplify GHA workflow names

brynary:
The emojis were fun but have become a pain, simplify this for utility.


lsegal:
Been toying with this. Would love if this were done on the individual job names too!

brynary:
Agreed, the jobs could use a pass also

PR #1158, opened 2024-11-19
Make with prefixes default behaviour

marschattha:
Removes the `--with-prefixes` flag from `qlty init` and makes the behaviour for prefix during init default.

### DIFF
--- a/qlty-cli/src/commands/init.rs
+++ b/qlty-cli/src/commands/init.rs
@@ -40,10 +40,6 @@ pub struct Init {
     /// a path to a local directory(name=directory).
     #[arg(long, value_parser = SourceSpec::new)]
     pub source: Option<SourceSpec>,
-
-    /// Enable plugin prefix detection.
-    #[arg(long)]
-    pub with_prefixes: bool,
 }
 
 impl Init {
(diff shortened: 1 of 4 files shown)


brynary (commented):
If we remove the flag, any invocation will start erroring. To be more backwards compatible, can we please make the flag a no-op for now and print a `warn!` which tells the user `--with-prefixes is deprecated and will be removed in a future version`

marschattha:
This only dictates whether during `qlty init` the plugins generated can have a prefix option, so don't think it will cause any errors.

But yes a deprecation warning is probably a safer transition.

qlty-cli/src/commands/init.rs
```diff
@@ -41,13 +41,21 @@ pub struct Init {
     #[arg(long, value_parser = SourceSpec::new)]
     pub source: Option<SourceSpec>,
 
+    /// Warning: this option has been deprecated!
```
brynary, line 44: We can actually hide options from the auto-generated Clap output, which I think is better than a warning 

```rust
#[arg(hide = true, long)]
pub with_prefixes: bool,
```

brynary (approved):

PR #1160, opened 2024-11-19
Attempt to fix source issue on CircleCI

marschattha:


brynary (approved):
I agree this is better, but do you have a theory on why this may fix the bug report?

lsegal:
Why would the .git matter?

PR #1180, opened 2024-11-20
Fix cache hits for config batches

marschattha:
If there is a driver that is batched by config with the config file in `.qlty/configs` directory the cache hit was being missed because the path for the config file was changed during batch computations and not reverted back to the original path of the file.


brynary:
@marschattha I don't have any reason to believe this should not work, however modifying the path and then modifying it back strikes me as a bit messy.

How could we approach this in a way which doesn't require that?

brynary:
Tagged @lsegal in case he has any feedback. In the meantime, going to merge this now to cut a release, since this is a significant performance regression.

PR #1194, opened 2024-11-20
Do not require qlty.toml for coverage command

lsegal:
Use default config options if qlty.toml is not present.


marschattha (approved):
LGTM!

PR #1196, opened 2024-11-21
Do not run source upgrade (if upgrading CLI)

noahd1:
This is a bugfix that does change the behavior of `qlty upgrade` in some cases.

Previously, if you ran `qlty upgrade` within a Git repository, it would upgrade the CLI, but then also upgrade the sources.

The problem was if you ran it outside of a `.git` directory: it would upgrade the CLI and then raise the following error:

❌ This subcommand must be run within a Git repository.

Source upgrades can also be performed independently by running:

`qlty upgrade source`

We could also _optionally_ run upgrade sources if within a git directory, but that would require additional code.

### DIFF
--- a/qlty-cli/src/commands/upgrade.rs
+++ b/qlty-cli/src/commands/upgrade.rs
@@ -88,8 +88,6 @@ impl Upgrade {
             self.install(&executable_path)?;
         }
 
-        SourceUpgrade::new().run()?;
-
         self.install_completions().ok();
         self.print_result(&timer, &release);
         CommandSuccess::ok()


brynary:
@noahd1 the source upgrade behavior is valuable, so I think we need to account for that.

When run outside of of an initialized project, we should just update the CLI. When run inside of an initialized project, we should perform the current behavior to upgrade the CLI and then upgrade the sources

lsegal:
Would it be sufficient to just use the following? Not sure if this will generate garbage in logs or not.

```rust
let _ = SourceUpgrade::new().run();
```

noahd1:
Tried it. It works. 5 character addition FTW.

Updated the title and the description to more properly reflect the decided upon solution.

brynary (approved):

PR #1200, opened 2024-11-21
Refactor qlty upgrade to re-use the install script

lsegal:
The existing `qlty upgrade` command has a few platform compatibility and failure case issues, alongside unnecessary complexity:

1. Windows is currently unsupported. The current upgrade command assumes the download is a tar xz package and does not support other file types.
2. Non-glibc Linux distributions will fail in unknown ways due to incorrect handling of platform.
3. The detection of latest version performs a ListBucket operation against S3 which is not only slow, but does not handle pagination. Detecting the latest version will begin to fail after 1000 releases. For reference, we have already generated ~300ish releases.
4. The package locations are hardcoded within the upgrade command and thus testing this system is difficult.
5. The logic of downloading, unarchiving, and managing of binaries is duplicated from the install scripts, creating another point of failure.

Fixing all of these issues is possible, but requires a lot of extra complexity. Complexity that has already been tackled and solved by our existing installer scripts.

Instead this PR considers a different approach: when `qlty upgrade [--version X.Y.Z]` is called, simply re-run the installer scripts with an optional version to install (latest by default).

This also has the added benefit of separating installation logic from the qlty binary. Specifically, if the qlty binary had a logical bug causing issues with updating (for example the Rust code moved the binary to an invalid location and crashed), users would be locked into an upgrade loop. By instead relying on the remote installer script, issues with installation can be fixed independently reducing the failure case. This does not entirely eliminate the possibility of an upgrade loop, but rather minimizes the risk.

Version detection is also simplified by reading the latest version from the [cargo-dist manifest.json file](https://qlty-releases.s3.amazonaws.com/qlty/latest/dist-manifest.json) instead of using the ListBuckets approach.

### DIFF
--- a/qlty-cli/src/commands/upgrade.rs
+++ b/qlty-cli/src/commands/upgrade.rs
@@ -1,17 +1,11 @@
 use crate::QltyRelease;
-use crate::{upgrade::ReleaseSpec, Arguments, CommandError, CommandSuccess};
-use anyhow::{bail, Context, Result};
+use crate::{Arguments, CommandError, CommandSuccess};
+use anyhow::Result;
 use clap::{Args, Subcommand};
 use console::style;
-use qlty_analysis::utils::fs::path_to_string;
 use qlty_analysis::version::QLTY_VERSION;
 use qlty_config::sources::SourceUpgrade;
-use std::{
-    path::{Path, PathBuf},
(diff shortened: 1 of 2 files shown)


brynary (approved):
Nice simplification!

PR #1206, opened 2024-11-22
Support biome suggestions

marschattha:


brynary (approved):

brynary:
@marschattha were we able to make this work without adding a step to read source code to the formatter?

marschattha:
Yes, I found a field in their json output that had the contents of the file, used that. 

I think there is a small bug related to suggestions, so I moved it to draft. 
I will check in the morning.

PR #1207, opened 2024-11-22
 Support in-place upgrades and support non-curl systems

lsegal:
Since wget is supported in the install script, do not rely on curl to download the script. Instead download via ureq.


lsegal:
Just manually tested on Windows / Linux / macOS both in and outside of ~/.qlty and they all work.

PR #1213, opened 2024-11-25
Update trivy version

marschattha:
Update trivy version, the aws test does not give any issues in the newer versions of trivy so did not add that test for 0.57.1.


brynary (approved):
Hmm, does it seem like a bug that we aren't getting issues for the new verson?

marschattha:
I ran Trivy directly on the test which also did not show any issue for 0.57.1, while it does for older version 0.50.1 so it is probably a change than a bug. 
If it is a bug it is on Trivy side.

brynary:
OK, cool. Sounds good to merge when this builds green

Issue #1201, opened 2024-11-21
Bug: error with hadolint during initial install


joshknowles:
### What happened?

Error during the init for new repository related to the hadolint plugin
![CleanShot 2024-11-21 at 06 07 19@2x](https://github.com/user-attachments/assets/9570d017-74a6-4ca8-9b88-c18f4a6da683)

### What did you expect to happen?

Successful init

### Can you reproduce the problem?

Yes, consistently

### CLI Version

0.448.0

### Relevant log output

```shell
› Initializing qlty in the current directory...
✔ Created .qlty/ directory
✔ Fetched plugins definitions
✔ Created .gitignore in .qlty/
✔ Created qlty.toml config file with 18 plugins

Plugin         Version  Targets      Config
prettier       3.3.2    1,028 files  
hadolint       latest   7 files      .hadolint.yaml
trufflehog     latest   1,718 files  
brakeman       latest   0 files      
yamllint       latest   117 files    
eslint         8.57.0   563 files    
ruff           0.5.3    6 files      
bandit         1.7.9    6 files      
checkov        latest   483 files    
trivy          0.53.0   323 files    
actionlint     latest   8 files      
osv-scanner    latest   7 files      
golangci-lint  latest   0 files      
shfmt          latest   26 files     
dotenv-linter  latest   7 files      
ripgrep        latest   601 files    
markdownlint   0.41.0   177 files    
shellcheck     latest   26 files     

✔ Set up qlty in the current directory
✔ Would you like to run plugins to see a sample of issues? · yes

› Running qlty check --sample 5

     [0/1] 🤔  Planning...  0.28s
     [1/1] 🔍  Analyzing 5 samples... 
❌ Failed to get version for package "hadolint": (command sh -c "hadolint --version" exited with code 0)
❌ command ["/Users/jknowles/.qlty/bin/qlty", "check", "--sample", "5"] exited with code 99
```

marschattha:
Haven't been able to replicate this, also looking at hex data for hadolint there haven't been any failures reported. 
I am more worried about the first error up top in your screenshot could be causing this somehow.
Do you get a similar error on lets say an open source repo we can look at?

<img width="406" alt="image" src="https://github.com/user-attachments/assets/65df45fa-f7f7-4583-8be5-2c38c4f3aa7c">

joshknowles:
I'm able to consistently reproduce. Blocked on running `qlty install` due to
```
[...]

PR #1214, opened 2024-11-25
Don't automatically enable hadolint

brynary:
The latest release of Hadolint is ~2 years old and it segfaults on MacOS Sequoia. To avoid auto-enabling a plugin that will produce an error, this moved Hadolint to only be auto-enabled if a Hadolint configuration file is already in the repository.

#1201


marschattha:
Um wrong file added and committed?

brynary:
@marschattha thanks, fixed

PR #1216, opened 2024-11-25
Add wait on github 403

marschattha:


brynary:
We discussed this a bit more in Slack and the conclusion was that using sleeps to overcome wait limits is problematic in an interactive CLI context, especially as some rate limits can take 60 mins before they reset.

PR #1217, opened 2024-11-26
Add QLTY_GITHUB_TOKEN support

marschattha:


qlty-check/src/tool/github.rs
```diff
@@ -12,6 +12,7 @@ use tracing::{debug, info, trace};
 
 const GITHUB_API_VERSION: &str = "2022-11-28";
 const USER_AGENT_PREFIX: &str = "qlty-check";
+const QLTY_GITHUB_TOKEN: &str = "QLTY_GITHUB_TOKEN";
```
brynary, line 15: I think we can skip this constant variable. The constant is nice when it is assigning the name to a value but in this case it's an env var so we don't have extra info to convey. (And we have a naming convention so we can always search for QLTY_ to find env vars)

brynary (approved):
Approved with one small note

qlty-check/src/tool/github.rs
marschattha: Oh yea added this when I thought we might need it in multiple places.

PR #1219, opened 2024-11-26
Add plugins disable

marschattha:
Add plugins disable subcommand


brynary (approved):
This looks great, thanks

PR #1227, opened 2024-11-26
Bump clippy version

brynary:


.qlty/qlty.toml
```diff
@@ -69,6 +71,7 @@ version = "1.7.9"
 
 [[plugin]]
 name = "clippy"
+version = "1.82.0" # This should be kept in sync with rust runtime version above
```
marschattha, line 74: Actually this line doesn't matter beyond readability.

PR #1229, opened 2024-11-26
Update enable plugins and add tests

marschattha:
Slight cleanup and added some tests.

Also I was thinking of addressing the warning on enable [docs](https://docs.qlty.sh/cli/commands/plugins-enable) 

> Currently, the command does not consider whether the plugin may already be enabled.

But not sure if we should ignore plugins that are already activated


brynary (commented):
If the plugin is already enabled, what do you think about printing a warning to the terminal and not editing the file?

brynary (approved):

PR #1230, opened 2024-11-26
Update clippy version

marschattha:


.qlty/qlty.toml
```diff
@@ -71,7 +71,7 @@ version = "1.7.9"
 
 [[plugin]]
 name = "clippy"
-# Note: Clippy version is controlled by the rust version above
+version = "1.82.0" # Note: Clippy version is controlled by the rust version above
```
marschattha, line 74: Turns out we do need this in order to make sure the version check passes after tool install.

This is needed since the latest_version defined in the plugin definition was `1.77.2` so we are expecting that when version is not defined, but bumping the rust runtime above we are basically installing `1.82.0` so version check fails.

brynary (approved):

PR #1236, opened 2024-11-27
Fix trycmd tests

marschattha:
The trycmd tests haven't been running it seems since the workspace overhaul.


qlty-cli/tests/cmd/fmt/basic.in/.qlty/qlty.toml
```diff
@@ -2,7 +2,7 @@ config_version = "0"
 
 [sources.default]
 repository = "https://github.com/qltysh/qlty"
-tag = "v0.13.0-pre"
+tag = "v0.452.0"
```
lsegal, line 5: Shouldn't this use `branch = "main"` otherwise we risk testing against old code?

brynary (approved):
Nice catch

PR #1241, opened 2024-11-29
Avoid panic for unknown shell

brynary:
Downgrade this from a panic to a command error which will cause the CLI to exit non-zero.


lsegal:
Weird test failure 🤔

lsegal (approved):
Code looks fine but you may wanna retrigger the unit tests to make sure that failure is ephemeral.

brynary:
Thanks.

@marschattha has identified that we have inconsistent results for our timeout tests on CI. Since there's no reason to believe CI would behave differently than a laptop, I am assuming this is a bug.

However, not much value in chasing a ghost so we are going to disable the test to improve CI stability and keep an eye out for a bug report.

PR #1243, opened 2024-11-29
Pre-commit (fmt) and pre-push (check) git hooks

brynary:
Note: A `git push` operation can push one or more references, and the references that are pushed may not be a reference that is checked out in the local working copy.

Since the `check` command always works on the files that are checked out in the working directory, the files that we check may not match the actual files being pushed.

Also, it's unclear what the correct behavior should be if multiple references are being pushed simultaneously. Mt first gut reaction is probably to no-op, since we don't want to disrupt the user.

Git passes a pre-push hook data on stdin which contains the commit ID of the local and the remote, and this PR uses the remote commit ID as our "upstream", so that the `check` execution will target the files that are changed in the diff.

Also -- I am shelling out to `git add` and it would be nice to replace this with the equivalent behavior emulated by the git2 crate.


brynary:
@lsegal this is my experiment to font as a pre-commit hook. It's not working and I would value your input.

I can install the pre-commit hook and it will indeed fmt the files when I run git commit.

However, the reformatting edits do not end up in the commit. I believe the issue has something to do with the Git index and the need to do something special in the pre-commit in order for the commit to include the reformatting.

The `--index` option I added to this PR attempts to do it but throws an error that the index is locked (which makes sense -- I believe Git itself is locking it).

I'm hunting around StackOverflow to find a better lead

qlty-cli/src/commands/fmt.rs
```diff
@@ -73,6 +86,22 @@ impl Fmt {
...
+        let mut args = vec!["add"];
+
+        for path in paths {
+            if let Some(path_str) = path.to_str() {
+                args.push(path_str);
+            }
+        }
+
+        if args.len() > 1 {
+            cmd("git", &args).run()?;
```
brynary, line 99: I was unable to find a way to accomplish this same behavior with the git2 crate

brynary:
I was able to get this working seemingly reasonably at least on my machine.

.qlty/hooks/pre-commit.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Install with:
+#
+# ln -s ../../.qlty/hooks/pre-commit.sh .git/hooks/pre-commit
+# chmod +x .git/hooks/pre-commit
```
lsegal, line 6: A `qlty git-hooks` command would be useful to auto symlink this stuff for you -- as well as a `qlty git-hooks init` subcommand to setup a basic default. Just a thought.

qlty-cli/src/commands/fmt.rs
lsegal: git2 doesn't support add? What is the missing piece here?

.qlty/hooks/pre-commit.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
lsegal, line 1: This should probably be `#!/bin/sh` for compatibility.

.qlty/hooks/pre-push.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
lsegal, line 1: This should probably be `#!/bin/sh` for compatibility.

lsegal (approved):
Works on my Windows machine too.

.qlty/hooks/pre-commit.sh
brynary: Agreed, will work on that next

qlty-cli/src/commands/fmt.rs
brynary: It has an add capability on the Index struct, which I thought would provide what we need. Unfortunately in my (admittedly very limited) testing, it did not behave exactly the same as shelling out to git add.

I am fairly confident git2 has the raw functionality but it wasn't immediately obvious to me what sequence of git2 calls emulates the high level command

brynary:
Great thanks, will clean up and merge.

PR #1251, opened 2024-12-02
Add missing_output_as_error to eslint

marschattha:
This should fix https://github.com/qltysh/qlty/issues/1223


brynary:
To expand on that... The reason I think it's valuable to limit this behavior to the smallest version set is that essentially this flag introduces ambiguity when otherwise there would be clarity.

PR #1254, opened 2024-12-04
Disable semgrep lints on Windows

lsegal:
This is an initial quick fix to get Windows compatibility when qlty.toml specifies semgrep. Platform support should ideally be configurable in the plugin spec.


brynary (approved):
Looks good as a temporary patch. We have a feature to add plugin platform targeting in the near term backlog

PR #1255, opened 2024-12-05
Ignore unsafe rules from check --fix

lsegal:
Seed with some rules we do not "trust" to work reliably. Respects `--unsafe` flag.

Also fixes an eslint issue in converting char ranges to bytes for replacements.

### DIFF
--- a/qlty-check/src/parser/eslint.rs
+++ b/qlty-check/src/parser/eslint.rs
@@ -76,6 +76,19 @@ impl Parser for Eslint {
                 let end_column = message.end_column.unwrap_or(column) as u32;
 
                 let suggestions = if let Some(fix) = message.fix {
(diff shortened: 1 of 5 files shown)


brynary (approved):
PR looks good.

One quick clarification... Am I reading this PR and the code correctly that prior to this patch the `--unsafe` option did not have any impact on _non-LLM_ fixes?

(I think we added the concept of "unsafe" when we introduced LLM fixes, and before we thought about safe vs. unsafe _tool-generated_ fixes, so that would make sense given the evolution.)

lsegal:
> (I think we added the concept of "unsafe" when we introduced LLM fixes, and before we thought about safe vs. unsafe tool-generated fixes, so that would make sense given the evolution.)

Good catch I didn't realize this was already wired up through check. Updated with unsafe support.

PR #1259, opened 2024-12-06
Consolidate version computation into qlty-config

lsegal:
Also simplify build scripts


brynary (approved):
This looks great, thanks!

PR #1262, opened 2024-12-06
Bundle plugin definitions into compiled binary

brynary:
Opening a POC for feedback...

This approach uses rust-embed to build the TOML files directly into the crate. Therefore, we could support a third type of Source, in addition to Git-based and local sources: A "default" source which is compiled into the CLI.

So instead of generating this:

```toml
[[source]]
name = "default"
tag = "v0.450.0"
```

We could generate something like this:

```toml
[[source]]
name = "default"
default = true
```

This would mean that for default configurations the CLI version _is_ the plugins version.


qlty-cli/tests/cmd/init/network/basic.stdout
```diff
@@ -76,8 +74,8 @@ name = "shfmt"
 [[plugin]]
 name = "trivy"
 drivers = [
-  [..],
-  [..],
+  "config",
+  "fs-vuln",
```
brynary, line 78: We are now rendering these in deterministic, alphabetical order

brynary:
@lsegal I think this is cross-platform-compatible, but could you please give this a try on Windows when you have a moment?

lsegal (approved):

PR #1263, opened 2024-12-06
Add `githooks install` subcommand

brynary:
```sh
# Create hooks in .qlty/hooks and symlink them from .git/hooks
qlty githooks install

# Remote .qlty/hooks and deactivate any symlinks by making them non-executable
qlty githooks uninstall
```


qlty-cli/src/commands/githooks/install.rs
```diff
@@ -0,0 +1,93 @@
...
+        })?;
+    }
+
+    let hook_relative_path = Path::new("..")
+        .join("..")
+        .join(".qlty")
+        .join("hooks")
+        .join(script_filename);
+
+    symlink(&hook_relative_path, &symlink_path).with_context(|| {
```
brynary, line 65: Need to figure out what needs to be done to make this work on Windows

qlty-cli/src/commands/githooks/install.rs
```diff
@@ -0,0 +1,93 @@
...
+            "Failed to get metadata for {} symlink at {}",
+            hook_name,
+            symlink_path.display()
+        )
+    })?;
+
+    let mut perms = metadata.permissions();
+    perms.set_mode(0o755);
+
+    fs::set_permissions(&symlink_path, perms).with_context(|| {
```
brynary, line 84: Does `set_permissions` work on Windows?

qlty-cli/src/commands/githooks/install.rs
lsegal: This is how we do it elsewhere:

https://github.com/qltysh/qlty/blob/b5801ea27ca94c385b714d5ae93522da843863ed/qlty-check/src/executor/staging_area.rs#L288-L295

brynary: What do you think about extracting this into a utility function which we could put in either a qlty-tips crate (which sits at the bottom of the dependency graph) or just drop into qlty-config initially which is a dependency of everything?

qlty-cli/src/commands/githooks/install.rs
```diff
@@ -0,0 +1,109 @@
...
+use std::os::unix::fs::PermissionsExt as _;
+
+#[derive(Args, Debug)]
+pub struct Install {}
+
+const QLTY_HOOKS_DIR: &str = ".qlty/hooks";
+
+const PRE_COMMIT_HOOK: &str = r#"#!/bin/sh
+qlty fmt --trigger pre-commit --index-file="$GIT_INDEX_FILE"
+"#;
```
lsegal, line 18: @brynary FYI something to consider here is using the Rust includr_str embed approach from #1262 to separate this out into its own .sh file so that we can have shellcheck run against the file, more easily edit, etc.

PR #1265, opened 2024-12-06
Reek plugin

noahd1:
@marschattha Thank you for pairing with me on this.


plugins/linters/reek/plugin.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[plugins.definitions.reek]
+runtime = "ruby"
+package = "reek"
+file_types = ["ruby"]
+latest_version = "6.3.0"
+known_good_version = "6.3.0"
+version_command = "reek --version"
+config_files = [".reek.yml"]
+description = "Ruby linter"
```
brynary, line 11: Using the [official description](https://github.com/troessner/reek?tab=readme-ov-file#code-smell-detector-for-ruby) is preferred when possible:

```suggestion
description = "Code smell detector for Ruby"
```

qlty-check/src/parser/reek.rs
```diff
@@ -0,0 +1,98 @@
+use super::Parser;
+use anyhow::Result;
+use qlty_types::analysis::v1::{Category, Issue, Level, Location, Range};
+use serde::{Deserialize, Serialize};
+
+pub struct Reek {}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ReekJson {
```
brynary, line 9: The name `ReekJson` signals to me that this struct will represent the entire JSON structure, however this seems like this corresponds with a specific smell so I would call this `ReekSmell`. 

Also, does it need to be `pub`?

qlty-check/src/parser/reek.rs
```diff
@@ -0,0 +1,98 @@
...
+                tool: plugin_name.into(),
+                documentation_url: smell.documentation_link,
+                message: smell.message.trim().to_string(),
+                category: Category::Lint.into(),
+                level: Level::Medium.into(),
+                rule_key: smell.smell_type,
+                location: Some(Location {
+                    path: smell.source,
+                    range: Some(Range {
+                        start_line: smell.lines[0] as u32,
```
brynary, line 34: Are we throwing away data here? (`lines` after first element)

qlty-config/src/config/plugin.rs
```diff
@@ -551,6 +553,7 @@ impl std::fmt::Display for OutputFormat {
             OutputFormat::Ruff => write!(f, "ruff"),
             OutputFormat::GolangciLint => write!(f, "golangci_lint"),
             OutputFormat::Biome => write!(f, "biome"),
+            OutputFormat::Reek => write!(f, "reek"),
```
brynary, line 556: I would sort these alphabetically for consistency

qlty-config/src/config/plugin.rs
```diff
@@ -515,6 +515,8 @@ pub enum OutputFormat {
     Coffeelint,
     #[serde(rename = "ruff")]
     Ruff,
+    #[serde(rename = "reek")]
+    Reek,
```
brynary, line 519: I would sort these alphabetically for consistency

qlty-check/src/parser/reek.rs
```diff
@@ -0,0 +1,98 @@
+use super::Parser;
+use anyhow::Result;
+use qlty_types::analysis::v1::{Category, Issue, Level, Location, Range};
+use serde::{Deserialize, Serialize};
+
+pub struct Reek {}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
brynary, line 8: Do we need `Serialize`? I think only _De_serialize

qlty-check/src/executor/driver.rs
```diff
@@ -429,6 +430,7 @@ impl Driver {
             OutputFormat::Ruff => Box::new(Ruff {}),
             OutputFormat::GolangciLint => Box::new(GolangciLint {}),
             OutputFormat::Biome => Box::new(Biome {}),
+            OutputFormat::Reek => Box::new(Reek {}),
```
brynary, line 433: Would sort these alphabetically

plugins/linters/reek/plugin.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[plugins.definitions.reek.drivers.lint]
+script = "reek --format json ${target}"
+success_codes = [0, 2]
```
brynary, line 15: Is there documentation for these 0 and 2 exit codes being success?

I am not 100% sure but I am wondering if it would be a good idea to comment the plugin definition code with references to the relevant docs to make it easier on someone making changes later not needing to start from scratch.  

Thoughts?

qlty-check/src/parser/reek.rs
```diff
@@ -0,0 +1,98 @@
...
+pub struct ReekJson {
+  pub context: String,
+  pub lines: Vec<i32>,
+  pub message: String,
+  pub smell_type: String,
+  pub source: String,
+  pub documentation_link: String,
+}
```
brynary, line 16: Is there documentation on this JSON format available?

brynary (approved):
Nicely done! Added some comments.

brynary:
@noahd1 how can we take the information you learned from pairing with @marschattha and encode it into docs so that the next person who does a plugin can do it more easily?

qlty-check/src/parser/reek.rs
noahd1: I could not find an example of a multi-line smell and I tested the following scenarios:
- when the smell references a multi-line violation (too many statements e.g.)
- when the same smell type has more than a single violation in the same file

I did find it odd however that the JSON format calls the attribute `lines` and accepts an array, so maybe worth digging in more.

plugins/linters/reek/plugin.toml
noahd1: Unfortunately, the only reference to these exit codes is if you run `reek --help` which informs you what the default exit codes are and how to override them if you'd like.

qlty-check/src/parser/reek.rs
noahd1: None that I found, but there is this spec:

https://github.com/troessner/reek/blob/master/spec/reek/report/json_report_spec.rb

qlty-check/src/parser/reek.rs
brynary: OK. Aside from peeking at the Reek code it may be hard to tell.

That said, if lines is an array of line numbers, we could use the min(lines) as the start_line and max(lines) as the end line. That might be more compatible, but would be equivalent to this code if lines is a single element

This could reduce the risk of a future bug if they start using the field in a way they are not using it today. (It just makes an assumption that lines is contiguous -- however that is probably a better behavior than throwing away data).

A final option would be to adjust the parser so that if unexpected data occurs it returns an error.

Combining these, here is my off the cuff recommendation for long term reliability:

1. Add a condition to evaluate if the lines array is contiguous
2. If lines is contiguous, use the minimum as the start line and the maximum as the end line
3. If lines is more than one element and not contiguous, return an error from the parser

This may be overkill but it's probably about 10 extra lines of code and may save our butt someday.

plugins/linters/reek/plugin.toml
brynary: OK, we could capture that information into a comment for the benefit of a future developer

While I enjoy having short and sweet plugin definitions, I don't see a harm in commenting them

qlty-check/src/parser/reek.rs
brynary: I'd add a code comment to the parser linking to that

A main thing to watch out for is not accidentally depending on a field always being present when it is present only in some conditions.

But I also wouldn't recommend using defensive coding on every field as that is cumbersome and really reduces the information density of the parser

noahd1:
@brynary I just spent some time writing down some notes in a local file based on my experience. It looks like I should update https://github.com/qltysh/qlty/blob/main/plugins/plugin_guide.md ... maybe in a PR so we can review separately.

qlty-check/src/parser/reek.rs
noahd1: I did manage to look at the source code and generate examples where `lines` is an array. e.g.

```rb
# My documented class
class Foo
  def check_response
    if respond_to?(:foo)
      puts "hi"
    elsif respond_to?(:bar)
      puts "var"
    end
  end
end
```

This sees 2 instances of the `ManualDispatch` violation and generates the following (in json format):

```json
[{"context":"Foo#check_response","lines":[4,6],"message":"manually dispatches method call","smell_type":"ManualDispatch","source":"linters/reek/fixtures/basic.in.rb","documentation_link":"https://github.com/troessner/reek/blob/v6.3.0/docs/Manual-Dispatch.md"}]
```

brynary: An great. So now the question is, how should we translate that data to Qlty formatted data most accurately

noahd1: @brynary I updated the code to handle the above use case.

brynary (approved):
This looks fantastic! Yes, if you could please update the Plugin Guide markdown in another PR that would be helpful.

tim-gq:
@noahd1 @brynary Thanks for the quick work on the reek plugin!  💯 

I've just enabled it.  I noticed it installed the reek gem, despite us having it in our Gemfile. I think this is probably expected.

Just wanted to check in with you, does it observe the config in `.reek.yml`?

plugins/linters/reek/plugin.toml
kceb: @brynary @noahd1  I don't think status code 2 is considered success in reek 🤔 

I get status code of 2 when there _are_ lint errors with reek. Their source code also seems to signal this


Reek source: https://github.com/troessner/reek/pull/808/files#diff-d36785fb0b5b36a2a6fe691901b0cca5c9279a3ba9fd1478031169a850579856R67

Screenshot:
Annotation on 2025-06-23 at 13-58-58

kceb: As a result of this, my reek checks are always returning successfully via Github checks even when there are newly introduced reek issues

PR #1268, opened 2024-12-07
Kube-linter plugin

noahd1:
Support for `kube-linter`: https://github.com/stackrox/kube-linter

Things I'm unsure about:

### Installation
When there's [multiple ways to install](https://docs.kubelinter.io/#/?id=installing-kubelinter) (binaries, go packages, etc), what do we prefer?

### file type targeting

Any `yaml` file can be a kubernetes configuration file. It becomes linted when you point the tool at it specifically. So I've specified the file types as `yaml` ...

```toml
file_types = ["yaml"]
```

However, now it's considering _all_ yaml files to be kubernetes configuration files. How do we handle this? 

### configuration files
`kube-linter` allows users to [point to their own configuration files](https://docs.kubelinter.io/#/configuring-kubelinter?id=configuring-kubelinter) using a `--config` CLI argument. Is that something we can support or can support easily?


brynary:
> When there's [multiple ways to install](https://docs.kubelinter.io/#/?id=installing-kubelinter) (binaries, go packages, etc), what do we prefer?

Unless we run into issues, a binary download has the fewest moving pieces and will be the fastest

> Any `yaml` file can be a kubernetes configuration file. It becomes linted when you point the tool at it specifically. So I've specified the file types as `yaml` ...

> However, now it's considering _all_ yaml files to be kubernetes configuration files. How do we handle this? 

We don't have any mechanism for inspecting the content of files to determine targets.

It would be interesting to know... what does itself kubelint do?

> `kube-linter` allows users to [point to their own configuration files](https://docs.kubelinter.io/#/configuring-kubelinter?id=configuring-kubelinter) using a `--config` CLI argument. Is that something we can support or can support easily?

Yes, we support this if the user overrides the script in the plugin definition. At that point it basically becomes "user makes their own plugin definition"

However most common is for their plugin to have a known set of config file paths which it automatically uses and are declared in the `config_files` in the plugin definition. This covers the 80/20

noahd1:
> We don't have any mechanism for inspecting the content of files to determine targets.
> It would be interesting to know... what does itself kubelint do?

Well, according to documentation, it expects you to pass a kubernetes configuration file `kube-linter lint foo.yml` as an argument (or a folder which contains a bunch of them).

I ran a test though with the following directory structure:

```
[testing/kube-repo (main *%)] $ ls -l
total 16
-rw-r--r--  1 noah  staff    0 Dec  7 11:48 README.md
-rw-r--r--@ 1 noah  staff  497 Dec  7 11:53 kube.yml
-rw-r--r--@ 1 noah  staff   14 Dec  7 11:53 something-else.yml
[testing/kube-repo (main *%)] $ cat something-else.yml
foo:
  bar: 1
```

If I run any of these:
```
$ kube-linter lint .
```

```
kube-linter lint *.yml
```

I don't see any errors or warnings. Perhaps because there's at least 1 valid kubernetes config file it knows to ignore the other yaml files.

If I pass it the non-kubernetes file explicitly however:
```
[testing/kube-repo (main *%)] $ kube-linter lint something-else.yml
Warning: no valid objects found.
[testing/kube-repo (main *%)] $ echo $?
0
```
I get a warning (only) aka exit code == 0.

So maybe we're ok. (You _can_ configure kube-linter to print a non-zero exit code in these cases but seems like we don't want to actually).

plugins/linters/kube-linter/fixtures/__snapshots__/basic_v0.7.1.shot
```diff
@@ -0,0 +1,101 @@
...
+  securityContext:
+    runAsUser: 1000
+    runAsGroup: 3000
+    fsGroup: 2000
+  volumes:
+  - name: sec-ctx-vol",
+      "tool": "kube-linter",
+    },
+    {
+      "category": "CATEGORY_LINT",
```
brynary, line 69: @noahd1 is there any information related to category of the issues in the SARIF output by the analyzer?

If so, we should consider how we might use it.

If there's no data about issue category then we have a bit of a decision... we could leave them as Lint which is effectively the default. Or, for some analyzers which are primarily security related, we override the category to Security. This may be a plugin where we want to do that as well.

noahd1: There's nothing in the sarif output to indicate category (or severity). Is this override something that can be done in the plugin definition? Or does it require CLI code?

noahd1: Oh I see how `checkov` defines this:

```toml
output_category = "vulnerability"
output_level = "medium"
```

brynary:
@noahd1 this is not going to merge main cleanly since we moved some folders around... best is probably to just copy your kube-linter folder out of the repo, force reset this branch to `main`, then copy the folder back

plugins/linters/kube-linter/fixtures/__snapshots__/basic_v0.7.1.shot
brynary: yes exactly

noahd1:
@brynary I tried using a binary but get errors when attempting to download binaries on linux and windows:

Linux error:
```
2024-12-09T22:33:54Z DEBUG [ThreadId(28)]::qlty_check::tool::github (9.3 MB): Found Linux aarch64 system: https://github.com/stackrox/kube-linter/releases/download/v0.7.1/kube-linter-linux_arm64
2024-12-09T22:33:54Z DEBUG [ThreadId(28)]::qlty_check::tool::github (9.3 MB): Found MacOS aarch64 system: https://github.com/stackrox/kube-linter/releases/download/v0.7.1/kube-linter-darwin_arm64
2024-12-09T22:33:54Z DEBUG [ThreadId(28)]::qlty_check::tool::github (9.3 MB): Found Windows aarch64 system: https://github.com/stackrox/kube-linter/releases/download/v0.7.1/kube-linter-windows_arm64.tar.gz
2024-12-09T22:33:54Z ERROR [ThreadId(28)]::qlty_check::tool (15.7 MB): kube-linter: tool installation error: No download URL found for kube-linter@0.7.1 on Linux/X86_64
```

Windows:
```
2024-12-09T22:36:07Z DEBUG [ThreadId(32)]::qlty_check::tool::github (9.3 MB): Found Linux aarch64 system: https://github.com/stackrox/kube-linter/releases/download/v0.7.1/kube-linter-linux_arm64
2024-12-09T22:36:07Z DEBUG [ThreadId(32)]::qlty_check::tool::github (9.3 MB): Found MacOS aarch64 system: https://github.com/stackrox/kube-linter/releases/download/v0.7.1/kube-linter-darwin_arm64
2024-12-09T22:36:07Z DEBUG [ThreadId(32)]::qlty_check::tool::github (9.3 MB): Found Windows aarch64 system: https://github.com/stackrox/kube-linter/releases/download/v0.7.1/kube-linter-windows_arm64.tar.gz
2024-12-09T22:36:07Z ERROR [ThreadId(32)]::qlty_check::tool (9.3 MB): kube-linter: tool installation error: No download URL found for kube-linter@0.7.1 on Windows/X86_64
```

Based on the [releases here](https://github.com/stackrox/kube-linter/releases/tag/v0.7.1) it looks like they have quite a few options, so I'd be surprised if those platforms/arch weren't supported. 

Not sure how much time we want to put into it as the go package download does seem to work.

brynary:
@noahd1 I think we should get to the bottom of using binaries.

What download definitions did you use?

We have a TOML syntax for downloads where the plugin author declares template URLs for different OS and architecture downloads

noahd1:
Tests are passing. w00t.

brynary (approved):
Approving this but the last thing we need to decide is when to auto initialize this plugin.

There are basically two options:

1. Auto initialize the plugin if the user has a configuration file for it present in their repo

2. Auto initialize the plugin if there are matching files

Since we are targeting YAML files for this, and we don't yet have a way to target initialization by file content, in practice I think option 1 is the only real option.

To do this you need to add a line like `suggests = "config"` (I'm on my phone and forget the exact syntax).

noahd1:
Found an example. 

```ts
suggested = "config"
```

PR #1270, opened 2024-12-08
Run fmt if check --fix modifies files

lsegal:


brynary (approved):
This is great! Do you think it would be beneficial to have a `--no-fmt``m or similar to disable this behavior, or overkill?

I suppose we can wait for a request and add it easily later, to avoid options overload.

lsegal:
@brynary given the very edge case for needing non-formatted fixes, I'd say it's better to add on request.

PR #1278, opened 2024-12-09
fix(cli): Prevent stack overflow on deeply nested expressions

marschattha:
Suggested fix for https://github.com/qltysh/qlty/issues/1277

### DIFF
--- a/qlty-analysis/src/code/visitor.rs
+++ b/qlty-analysis/src/code/visitor.rs
@@ -1,5 +1,8 @@
 use crate::Language;
+use tracing::warn;
 use tree_sitter::{Node, TreeCursor};
+
+const MAX_CURSOR_DEPTH: u32 = 400;
 
 pub trait Visitor {
     fn process_node(&mut self, cursor: &mut TreeCursor) {
@@ -8,6 +11,13 @@ pub trait Visitor {
         let language = self.language();
 
         if self.skip_node(&node) {
+            return;
+        }
+
+        if cursor.depth() > MAX_CURSOR_DEPTH {
+            // This is a safety check to prevent stack overflow
+            // due to infinite recursion on deeply nested trees.
+            warn!("Max cursor depth reached");
             return;
         }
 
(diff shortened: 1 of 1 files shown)


qlty-analysis/src/code/visitor.rs
```diff
@@ -1,6 +1,9 @@
 use crate::Language;
+use tracing::warn;
 use tree_sitter::{Node, TreeCursor};
 
+const MAX_CURSOR_DEPTH: u32 = 400;
```
brynary, line 5: How did you select 400?

brynary (commented):
I agree that applying a maximum depth limit like this is necessary, otherwise I think there is no way to avoid stack overflows.

(I am a bit confused at why this does not occur when analyzing a single file, but I don't think we need to get to the bottom of that because this change looks good generally.)

qlty-analysis/src/code/visitor.rs
marschattha: Basically ran `qlty smells --all` a number of times on a folder containing these files and two others while logging the cursor depth, usually stack overflow occurred above 450 depth.

 We can lower it to be safer.

marschattha:
> I am a bit confused at why this does not occur when analyzing a single file

My theory around that is that we are running in multiple threads and we are analyzing multiple files simultaneously for duplication checks?

qlty-analysis/src/code/visitor.rs
brynary: I think 400 is fine.

brynary:
> My theory around that is that we are running in multiple threads and we are analyzing multiple files simultaneously for duplication checks?

Yes, we do use multiple threads, although I thought each gets its own stack allocation. Anyways, no

brynary (approved):
LGTM!

qlty-analysis/src/code/visitor.rs
marschattha: I think going to lower it a bit down to 300 to be safer on more systems.

PR #1291, opened 2024-12-10
feat(plugin/biome): add more file types to biome plugin

lamualfa:
Biome supports more file types than the current plugin define.

https://biomejs.dev/internals/language-support/

### DIFF
--- a/qlty-plugins/plugins/linters/biome/plugin.toml
+++ b/qlty-plugins/plugins/linters/biome/plugin.toml
@@ -3,7 +3,7 @@ config_version = "0"
 [plugins.definitions.biome]
 runtime = "node"
 package = "@biomejs/biome"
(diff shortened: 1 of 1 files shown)


brynary:
Thanks, @lamualfa! These `file_types` keys are references to file types declared in [default.toml](https://github.com/qltysh/qlty/blob/main/qlty-config/default.toml). 

Reviewing those, I see some existing issues: For example, the `javascript` type includes `*.jsx` but with TypeScript the `*.tsx` pattern is separately in a `tsx` type, so that was unexpected. Also, the `javascript` type is missing extensions like `.cjs`. This is something we will need to address separately.

For Biome, based on the current file type definisions, `jsonc` is already covered because it is part of `json`, and `jsx` is currently part of `javascript` so I will remove those for now.

brynary (approved):

PR #1295, opened 2024-12-10
Tweaks to max issues behavior

brynary:
- Reduce `MAX_ISSUES_PER_FILE` from 1k to 100.
  - Rationale: Often these are files which have the wrong analysis configuration or shouldn't be analyzed at all. This better protects the entire analysis process from when a subset of files are noisy files (e.g. 10 "bad" files could previously exhaust the 10k limit)
- When we hit `MAX_ISSUES_PER_FILE` limit, preserve the 100 instead of dropping them all
- Fix: When `MAX_ISSUES` of 10k is hit in `Executor#invoke`, truncate the `issues` vector

Note: Looking at the code, it _seems_ like this truncation should be handled inside `fn build_issue_results` in the form of `break 'invocation_loop;` so I would appreciate an additional set of eyes on this. (We have seen cases where we have seen the total issues logged as up to 38k relative to the intended 10k limit.)

### DIFF
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -37,7 +37,7 @@ use std::{
 use tracing::{debug, error, info, warn};
 
 const MAX_ISSUES: usize = 10_000;
-const MAX_ISSUES_PER_FILE: usize = 1000;
+const MAX_ISSUES_PER_FILE: usize = 100;
 
 #[derive(Debug, Clone)]
 pub struct Executor {
@@ -136,7 +136,7 @@ impl Executor {
         }
 
         self.progress.clear();
-        let issues = Self::build_issue_results(
+        let mut issues = Self::build_issue_results(
             &self.plan.hits,
             &invocations,
             self.plan.settings.skip_errored_plugins,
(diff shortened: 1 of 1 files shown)


marschattha:
As it is currently in this PR, I think it might have some undesired consequences, particularly because of transformers, on PRs I believe transformers are responsible filtering out what issues we would like to report or hide.

Lets say we make change to a file and open up a PR, and that file had 150 issues already in the code previously and we introduced another 20 with the PR, once we truncate the issues before running the transformers we could potentially drop all  the new ones or a mix of them and end up with a potentially empty/partial issue set, even though more were introduced and reported by the plugin but got truncated.

marschattha:
I think we are truncating before running the transformers which filter issues in this PR, right? 
That can introduce a bug where issues that should of had been reported might get truncated.

marschattha:
Also won't it be better to enforce the MAX_ISSUES in `run_invocation_pools` function? 
The issues are transformed at that point and once we hit MAX_ISSUES we can stop running any further invocations.

https://github.com/qltysh/qlty/blob/08208a57c0a81ad61646aa77ba3ccc9ce7123d62/qlty-check/src/executor.rs#L392

brynary:
@marschattha that's a great point about the filtering transformers. The key question is: Is the limit on issues _generated_ (pre-filtering) or is it on issues _emitted_ (post-filtering)?

In theory we could limit both. However, for simplicity, I think limiting at the number of issues _generated_ (pre-filtering) is preferable because that is going to have better results in terms of protecting the resource utilization.

I'll take a look at `run_invocation_pools` now

PR #1302, opened 2024-12-13
Stop running invocations when total MAX_ISSUES is exceeded

brynary:
Tested locally and works as expected

### DIFF
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -408,12 +408,14 @@ impl Executor {
         pool.install(|| {
             invocations
                 .into_par_iter()
-                .map(|plan| {
+                .filter_map(|plan| {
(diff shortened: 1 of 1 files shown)


qlty-check/src/executor.rs
```diff
@@ -557,10 +576,10 @@ fn run_invocation(
     transformers: &[Box<dyn IssueTransformer>],
 ) -> Result<InvocationResult> {
     let task = progress.task(&plan.plugin_name, &plan.description());
-    let mut invocation = plan.driver.run(&plan, &task)?;
+    let mut result = plan.driver.run(&plan, &task)?;
```
brynary, line 579: This is actually an `InvocationResult` not an `Invocation` so I adjusted this local variable name for clarity.

marschattha (approved):

lsegal (approved):

PR #1306, opened 2024-12-13
Ability to abort qlty check during Git hooks

brynary:
- If `qlty check` is being run as a git hook
- Allow the user to press `<enter>` to short-circuit execution
- This is useful both for speed when needed and also to skip past otherwise-blocking issues

Note: Git does not use a tty for it's hooks. Therefore, reading from stdin as usual does not work, so instead we manually try to open and read `/dev/tty`. (Unsure what if anything we would need to do to make this work on Windows.)

### Also

I worked on this on an airplane with no WiFi and so there's some additional teaks in here:

1. Improve qlty configuration, including getting ESLint working on new plugins folder structure
2. Reduce output when not running `--verbose`
3. Fix some minor output errors
4. Fix a panic in `qlty fmt` with a non-existent path argument

### DIFF
--- a/.qlty/qlty.toml
+++ b/.qlty/qlty.toml
@@ -14,36 +14,10 @@
 config_version = "0"
 
 exclude_patterns = [
-    "*_min.*",
-    "*-min.*",
-    "*.min.*",
-    "**/.yarn/**",
-    "**/*.d.ts",
-    "**/assets/**",
-    "**/bower_components/**",
-    "**/build/**",
-    "**/cache/**",
-    "**/config/**",
-    "**/db/**",
-    "**/deps/**",
-    "**/dist/**",
-    "**/extern/**",
-    "**/external/**",
     "**/fixtures/**",
-    "**/generated/**",
-    "**/Godeps/**",
-    "**/gradlew/**",
-    "**/linters/**",
-    "**/mvnw/**",
-    "**/node_modules/**",
     "**/protos/**",
-    "**/seed/**",
-    "**/target/**",
-    "**/testdata/**",
-    "**/tests/lang/**",
-    "**/vendor/**",
-
-    # Ignore symlink
+    "**/*.in/**",
+    "**/*.test.ts/**",
     "plugins/**",
 ]
 
--- a/qlty-plugins/plugins/.eslintrc.cjs
+++ b/qlty-plugins/plugins/.eslintrc.cjs
@@ -1,7 +1,7 @@
 /** @type {import('eslint').Linter.Config} */
 module.exports = {
   root: true,
-  ignorePatterns: ["linters"],
+  ignorePatterns: ["/linters"],
   parserOptions: {
     ecmaVersion: "latest",
     sourceType: "module",


qlty-cli/src/ui/format/text.rs
```diff
@@ -28,11 +28,21 @@ impl Formatter for TextFormatter {
...
+                match self.report.verb {
+                    ExecutionVerb::Check => "Checked",
+                    ExecutionVerb::Fmt => "Formatted",
+                    _ => "Processed",
+                },
+                self.report.targets_count(),
+                if self.report.targets_count() == 1 {
+                    "file"
+                } else {
+                    "files"
```
brynary, line 44: Fixes pluralization of `1 files`

qlty-cli/src/ui/format/text.rs
```diff
@@ -28,11 +28,21 @@ impl Formatter for TextFormatter {
...
+        if self.verbose >= 1 && self.report.targets_count() > 0 {
             writeln!(
                 writer,
-                "{}",
-                style(format!("Checked {} files", self.report.targets_count()))
+                "{} {} {}",
+                match self.report.verb {
+                    ExecutionVerb::Check => "Checked",
+                    ExecutionVerb::Fmt => "Formatted",
+                    _ => "Processed",
```
brynary, line 38: Improved output using verb-specific language

qlty-cli/src/commands/fmt.rs
```diff
@@ -86,6 +88,17 @@ impl Fmt {
         }
     }
 
+    fn validate_options(&self) -> Result<(), CommandError> {
+        for path in &self.paths {
+            if !path.exists() {
+                let message = format!("path '{}' does not exist", path.display());
+                return Err(CommandError::InvalidOptions { message });
```
brynary, line 95: Fixes a panic when running `qlty fmt non_existant_path`

qlty-cli/src/commands/check.rs
```diff
@@ -123,12 +125,21 @@ impl Check {
...
+        }
 
         let plan = Planner::new(ExecutionVerb::Check, &settings)?.compute()?;
 
-        steps.start(LOOKING_GLASS, format!("Analyzing{}...", plan.description()));
-        eprintln!();
+        if self.verbose >= 1 {
+            steps.start(LOOKING_GLASS, format!("Analyzing{}...", plan.description()));
+            eprintln!();
+        }
```
brynary, line 138: Reduce output when not running with `--verbose`

.qlty/qlty.toml
```diff
@@ -14,34 +14,11 @@
...
-    "**/node_modules/**",
     "**/protos/**",
-    "**/seed/**",
-    "**/target/**",
-    "**/testdata/**",
-    "**/tests/lang/**",
-    "**/vendor/**",
+    "**/*.in/**",
+    "**/*.test.ts/**",
+    "plugins/**",
```
brynary, line 21: Trim down exclude_patterns including removing some exclude patterns which were erroneously excluding real code (e.g. in folders with "build", "cache", or "config" in the names)

PR #1315, opened 2024-12-16
Set plugin `mode=disabled`(instead of removing whole plugin) when running `qlty plugins disable PLUGIN`

noahd1:
- Changes behavior of the `qlty plugins disable PLUGIN` to set the plugin mode to `disabled` (instead of removing it from the qlty.toml)
- Adds a couple error checks: errors if no plugins in qlty.toml or if the plugin referenced in the command is not found in their qlty.toml (used `upgrade` command as reference)

Couple notes:

1. The conversion of `IssueMode` to a string duplicates the serialization macro strings (`#[serde(rename = "block")]` e.g.) defined in the struct itself. From what I could tell, if we want to avoid this duplication, we could either a) use `serde_json` to translate the IssueMode value to JSON, and then extract the string out of the JSON (yuck). Or, instead, remove these macros and define our own serialization, using the new `to_string` method defined. I'm less sure where the serialization is actually happening, so not confident I could do this.
2. I looked/tested how `qlty plugins enable PLUGIN` behaves. If a plugin is disabled (with the `disable` command or by hand), you get an error message "plugin PLUGIN already enabled". So we'll have to change this behavior to just remove the disabled mode line (I'm guessing).

### DIFF
--- a/qlty-cli/src/commands/plugins/disable.rs
+++ b/qlty-cli/src/commands/plugins/disable.rs
@@ -29,11 +29,11 @@ impl ConfigDocument {
 
     pub fn disable_plugin(&mut self, name: &str) -> Result<()> {
         if let Some(plugins) = self.document["plugin"].as_array_of_tables_mut() {
-            plugins.retain(|plugin| {
+            plugins.iter_mut().for_each(|plugin| {
                 if let Some(plugin_name) = plugin.get("name").and_then(Item::as_str) {
-                    plugin_name != name
-                } else {
-                    true // Do not modify unknown data
+                    if plugin_name == name {
+                        plugin["mode"] = "disabled".into();
+                    }
                 }
             });
         }
@@ -109,6 +109,11 @@ config_version = "0"
 [[plugin]]
 name = "stays"
 version = "1.0.0"
+
+[[plugin]]
+name = "to_disable"
+version = "1.0.0"
+mode = "disabled"
 
 [[plugin]]
 name = "also_stays"
--- a/qlty-cli/src/commands/plugins/disable.rs
+++ b/qlty-cli/src/commands/plugins/disable.rs
@@ -28,10 +28,6 @@ impl ConfigDocument {
     }
 
     pub fn disable_plugin(&mut self, name: &str) -> Result<()> {
-        if self.document.get("plugin").is_none() {
-            bail!("No plugins found in qlty.toml");
-        }
-
         let mut updated = false;
 
         if let Some(plugin_tables) = self.document["plugin"].as_array_of_tables_mut() {
--- a/qlty-cli/src/commands/plugins/disable.rs
+++ b/qlty-cli/src/commands/plugins/disable.rs
@@ -131,4 +131,30 @@ version = "1.0.0"
 
         assert_eq!(config.document.to_string().trim(), expected.trim());
     }
+
+    #[test]
+    fn test_disable_plugin_wrong_name() {
+        let (temp_dir, _) = sample_repo();
+        let temp_path = temp_dir.path().to_path_buf();
+
+        let workspace = Workspace {
+            root: temp_path.clone(),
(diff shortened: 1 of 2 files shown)


qlty-cli/src/commands/plugins/disable.rs
```diff
@@ -28,14 +28,23 @@ impl ConfigDocument {
...
     pub fn disable_plugin(&mut self, name: &str) -> Result<()> {
-        if let Some(plugins) = self.document["plugin"].as_array_of_tables_mut() {
-            plugins.retain(|plugin| {
-                if let Some(plugin_name) = plugin.get("name").and_then(Item::as_str) {
-                    plugin_name != name
-                } else {
-                    true // Do not modify unknown data
+        if self.document.get("plugin").is_none() {
+            bail!("No plugins found in qlty.toml");
+        }
```
brynary, line 33: Do you see any concrete downside to removing this error check and letting it fall to the second error check? This one seems redundant at first read.

qlty-config/src/config/plugin.rs
```diff
@@ -715,7 +715,15 @@ pub enum IssueMode {
     Disabled = 4,
 }
 
-impl IssueMode {
+impl IssueMode {pub fn to_string(&self) -> String {
```
brynary, line 718: This formatting is wrong. Rust has a canonical auto formatted which you should be able to get your editor to run on save.

qlty-cli/src/commands/plugins/disable.rs
```diff
@@ -110,11 +119,42 @@ config_version = "0"
...
+    fn test_disable_plugin_wrong_name() {
+        let (temp_dir, _) = sample_repo();
+        let temp_path = temp_dir.path().to_path_buf();
+
+        let workspace = Workspace {
+            root: temp_path.clone(),
+        };
+        fs::create_dir_all(&temp_path.join(path_to_native_string(".qlty"))).ok();
+
+        fs::write(
```
brynary, line 145: I assume this is repeating existing pattern but it would be nice if we can test this without touching the filesystem since it's basically a pure function (toml string in, toml string out).

If you were so inclined, it would be a nice refactor

brynary (approved):
Approving with some notes.

Good catch on "enable" subcommand! It will need to get smarter. 

There's a few cases that it will need to handle... I do think we should retain the error from the "enable" if the plugin is already enabled but it will need to be more complex than just looking for the key

qlty-cli/src/commands/plugins/disable.rs
noahd1: No. I don't actually. I copied this from the `upgrade` sub-command, which follows similar logic, but will remove here (and note we can/should probably remove from there as well)

qlty-cli/src/commands/plugins/disable.rs
```diff
@@ -28,14 +28,19 @@ impl ConfigDocument {
...
-                    plugin_name != name
-                } else {
-                    true // Do not modify unknown data
+        let mut updated = false;
+
+        if let Some(plugin_tables) = self.document["plugin"].as_array_of_tables_mut() {
+            for plugin_table in plugin_tables.iter_mut() {
+                if plugin_table["name"].as_str() == Some(name) {
+                    updated = true;
+                    plugin_table["mode"] = value(IssueMode::Disabled.to_string());
```
lsegal, line 37: It might be a better pattern to `return Ok(())` from here and `bail()` at the end of the function as a catch-all, allowing you to drop the mutable book-keeping variable.

qlty-cli/src/commands/plugins/disable.rs
noahd1: I don't want to hold up this PR on this particular refactor. I may look into it, but had started work on updating our `enable` subcommand and want to address that first.

PR #1316, opened 2024-12-17
enhance `qlty plugins enable PLUGIN` to support enabling disabled plugins

noahd1:
Previously `qlty plugins disable PLUGIN` disabled a plugin by entirely removing the plugin definition from the `qlty.toml` file. In this world `qlty plugins enable PLUGIN` worked regardless of whether the plugin was never configured to begin with or was configured but removed with `qlty plugins disable PLUGIN`. 

This changed in #1315 when we changed the behavior of `qlty plugins disable PLUGIN` to instead mark the plugin as disabled (`mode = "disabled"`). Now, a `disabled` plugin differs from one never configured to begin with.  Because the `enable` subcommand was not "mode-aware" -- it assumed if the plugin was defined in the `qlty.toml`, regardless of its "mode", that it was enabled. This had the unfortunate user experience of informing a user that their plugin was already enabled when in fact it was explicitly disabled.

This PR makes the `enable` subcommand mode-aware: if a plugin is explicitly disabled with `mode = "disabled"` this PR removes that line from their configuration, effectively re-enabling that plugin.

This does mean that if at one point the plugin was configured with a mode other than the default (currently `block`), this would need to be put back into place.

### DIFF
--- a/qlty-cli/src/commands/plugins/enable.rs
+++ b/qlty-cli/src/commands/plugins/enable.rs
@@ -2,7 +2,7 @@ use crate::{Arguments, CommandError, CommandSuccess};
 use anyhow::{Context, Result};
 use clap::Args;
 use console::style;
-use qlty_config::Workspace;
+use qlty_config::{config::IssueMode, Workspace};
 use std::fs;
 use toml_edit::{array, table, value, DocumentMut};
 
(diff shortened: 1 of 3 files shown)


noahd1:
I happen to agree with qlty that this has deeply nested control flow. I'm trying to figure out a rust-y way to detangle that.

noahd1:
> I agree with qlty that this has deeply nested control flow. I'm trying to figure out a rust-y way to detangle that.

I addressed this with better match condition that I was struggling with previously.

qlty-cli/src/commands/plugins/enable.rs
```diff
@@ -47,10 +47,24 @@ impl ConfigDocument {
...
+            for plugin_table in plugin_tables.iter_mut() {
+                if plugin_table["name"].as_str() == Some(name) {
+                    match plugin_table.get("mode") {
+                        Some(value) if value.as_str() == Some(IssueMode::Disabled.to_str()) => {
+                            plugin_table.remove("mode");
+                            return Ok(());
+                        }
+                        Some(_) | None => {
+                            eprintln!(
+                                "{} Plugin {} is already enabled",
```
brynary, line 60: Is there a test for this case?

noahd1: There's a test for the `Some(_)` case but not the `None` case. Do you think it's valuable / necessary to cover both cases? I was on the fence.

brynary: No, I think having a test for Some is sufficient, as that's the trickier case

brynary (approved):

PR #1319, opened 2024-12-18
Add and support a `supported_platforms` driver key

noahd1:


noahd1:
@brynary Do you think this code to check supported platforms belongs in `Scanner` (this PR currently) or in `DriverCandidate` ? Seems like `DriverCandidate`'s responsibility is not to do any filtering, but it also seems off for Scanner to even consider a plugin for an unsupported platform.

qlty-config/src/config/plugin.rs
```diff
@@ -113,6 +113,8 @@ pub struct DriverDef {
 
     #[serde(default)]
     pub missing_output_as_error: bool,
+
+    pub supported_platforms: Option<Vec<Platform>>,
```
brynary, line 117: This use of `Option<Vec<T>>` catches my eye.

By using Option and Vec you can differentiate between null/omitted and an empty array. If that distinction is importanr, then this is the right type.

OTOH, if you want to interpret empty array just like an omitted field (which seems fine to me), then you can just use `Vec<T>` which defaults to `[]` and check it with `is_empty()`

qlty-check/src/planner/config.rs
```diff
@@ -54,12 +48,26 @@ fn configure_plugins(planner: &Planner) -> Result<Vec<ActivePlugin>> {
             continue;
         }
 
-        if PLATFORM_DISABLED_PLUGINS.contains(&enabled_plugin.name.as_str()) {
-            debug!(
-                "Plugin {} is disabled on this platform, skipping.",
-                enabled_plugin.name
-            );
-            continue;
+        let current_platform = if cfg!(target_os = "windows") {
```
brynary, line 51: I would extract this out into a dedicated function to encapsulate the cfg! stuff

qlty-check/src/planner/config.rs
```diff
@@ -54,12 +48,26 @@ fn configure_plugins(planner: &Planner) -> Result<Vec<ActivePlugin>> {
...
+        } else if cfg!(target_os = "linux") {
+            "linux"
+        } else {
+            "unknown"
+        };
+
+        if let Some(plugin_def) = planner.config.plugins.definitions.get(enabled_plugin.name.as_str()) {
+            if !plugin_def.supported_platforms.is_empty() {
+                if !&plugin_def.supported_platforms.iter().any(|platform| platform.to_string() == current_platform) {
+                    debug!(
```
brynary, line 64: I would upgrade this to `warn!` level

qlty-check/src/planner/config.rs
```diff
@@ -54,12 +48,26 @@ fn configure_plugins(planner: &Planner) -> Result<Vec<ActivePlugin>> {
...
+                        "Plugin {} is not supported on this platform, skipping.",
```
brynary, line 65: Would add the current platform and supported platforms to this log line

brynary (commented):
Adding some comments.

Normally, I would say the last thing this needs is a test. However, making a test that runs correctly on all platforms is somewhat annoying for this, so I would be find skipping it.

The logic is "shallow" enough that if something goes wrong it should be easy to spot

noahd1:
> Adding some comments.
> 
> Normally, I would say the last thing this needs is a test. However, making a test that runs correctly on all platforms is somewhat annoying for this, so I would be find skipping it.
> 
> The logic is "shallow" enough that if something goes wrong it should be easy to spot

I'd added a test already actually before reading your comment. The test gives a plugin `supported_platforms` of every platform but the current one. It seems to work ok.

noahd1:
I addressed your comments and marked as ready for review.

I haven't done an end to end test -- I only mention this because I'm a bit fuzzy on the deserialization of the plugin definition and whether what I've done is sufficient to read the `supported_plugins` value from a plugin definition on disk. I suspect it is though. The `PluginDef` has a `Deserialize` trait (?) and I assume that is what makes the new field I added to the struct de-serializable.

brynary:
Yes, the deserialize trait is in essence a macro (aka meta-programming) that automatically will read data from the TOML into the struct

noahd1:
Just performed an end to end test, and it worked. Unexpectedly it printed the log line about skipping the plugin twice. Perhaps configure plugins is called more than once.

brynary:
I think I've seen that with another log line, so I think it's a pre-existing condition which should be sorted out someday

brynary (approved):

PR #1321, opened 2024-12-18
Refactors for autofix API

lsegal:


brynary (approved):
I don't think I entirely follow the motivations for the various changes but don't think that is needed.

These are small changes everything in here looks fine and in most cases simpler so approving

PR #1341, opened 2024-12-19
Remove support for HashMap version of 'sources' key in qlty.toml

noahd1:
We have been supporting 2 different ways of defining sources in the `qlty.toml` configuration file:

### As a Vector
```toml
[[source]]
name = "default"
default = true
```

### As a HashMap
```toml
[sources.default]
default = true
```

The HashMap version is considered legacy, and as we look to reference sources more, it will make our lives easier if we only support a single way of defining sources.

NB: The `default_source` method was unused and removed

By removing the hashmap sources definition, we may find a small number of customers who will no longer have a correctly defined source in their qlty.tom. In these scenarios, this is the error I see:

```
❌ Plugin definition not found for name: actionlint
```

Would be preferable, probably, to either:

1. Print an error saying that a source is not defined in the qlty.toml file
2. Default to the default = true behavior in cases where not source is defined.

What do you think?

### DIFF
--- a/qlty-config/src/config.rs
+++ b/qlty-config/src/config.rs
@@ -30,10 +30,10 @@ pub use source::SourceDef;
 
 use crate::config::plugin::EnabledRuntimes;
 pub use crate::config::plugin::PluginsConfig;
-use crate::sources::{Source, SourcesList};
+use crate::sources::SourcesList;
 use crate::version::QLTY_VERSION;
 use crate::Library;
-use anyhow::{anyhow, bail, Result};
+use anyhow::{bail, Result};
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
@@ -67,9 +67,6 @@ pub struct QltyConfig {
 
     #[serde(default)]
     pub plugins: PluginsConfig,
-
-    #[serde(default)]
-    pub sources: HashMap<String, SourceDef>,
 
     #[serde(default)]
     pub language: HashMap<String, Language>,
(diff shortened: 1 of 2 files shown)


brynary:
I don't think that we can assume the default source when no sources are present, because that would have the negative side effect of making adding a custom source remove the default source.

I do think it is reasonable to require at least one source. Having a source does not do anything on it's own, so that still leaves the option open for having the source but not plugins.

Therefore, I would probably start by adding a separate PR for adding an error if there are no sources :)

noahd1:
I found code which purports to already detect when no sources are found, but it didn't seem to trigger for me. 

https://github.com/qltysh/qlty/blob/1ed884387a90789eed4ee523f94e6c3f1ceba43e/qlty-config/src/config/builder.rs#L104-L107

Also means, I found some code that references `sources` still and could be removed in this PR.

noahd1:
I think this means we're not properly propagating errors. Trying to determine where / why.

brynary:
> I think this means we're not properly propagating errors. Trying to determine where / why.

It's likely a specific issue not generalized / universal. 

You'll probably need to walk the call stack up from that function to see where it is called.

noahd1:
Ok, I was able to track it down and it appears that this is why an error isn't raised:

https://github.com/qltysh/qlty/blob/c3db04e3af9c8e3cbdb9a6825a3e0a6a2e063e9d/qlty-config/src/config/builder.rs#L39-L43

It appears that it's intentional, but not working as intended. Aka it falls back if sources aren't found, but it's falling back to a default toml file that ALSO doesn't specify sources. So, this is the error you get when using the fallback toml file. 

I think there's a few problems here, but I think the remedy might be to remove this fallback entirely, unless you see reason / intent to keep it?

brynary:
Ooof. There's a lot of complexity here and I believe some of it is important complexity but also some of it could probably be simplified. I with I had left some code comments.

I believe `extract_sources` (where you found the `bail!`) in particular is used specifically as a "cheat" to pull out source related data to perform operations on sources without loading the config regularly. This could probably all be burned down and re-thought as I was a Rust newbie when I did it, but it will be kicking a bee hive.

To minimize the risk of this project blowing up in scope, I would bail on the idea of throwing an error if there are no sources. The error `❌ Plugin definition not found for name: actionlint` is clear enough.

brynary (approved):

PR #1355, opened 2024-12-20
Warn user is using deprecated default source (repository reference instead of `default = true`)

noahd1:
We recently changed the default source for newly generated `qlty.toml` files from:

```
[[source]]
name = "default"
branch = "main"
```

to:

```
[[source]]
name = "default"
default = "true"
```

This was so that the plugins referenced, by default, are guaranteed to be compatible with the binary used, because the plugins are compiled into the binary.

This works for newly generated qlty.toml files, but old generated ones in the wild should, ideally, be updated. They will continue to work as is, and a repository source is still a valid way to load sources, however, it is inherently more susceptible to incompatability issues than `default = true`.

This PR adds a warning if using the old-style default source.

Warning looks like this currently:

image


qlty-config/src/config.rs
```diff
@@ -86,6 +87,8 @@ pub struct QltyConfig {
     pub source: Vec<SourceDef>,
 }
 
+const OLD_DEFAULT_SOURCE_REPOSITORY: &str = "https://github.com/qltysh/qlty";
```
brynary, line 90: You probably also want the old repository which was `qltyai/qlty`

noahd1: I tried adding this, but there's code that first attempts and fails to fetch from this remote.  I received instead an error:

```
[qltysh/qlty (nd-source-warning *)] $ cargo run -- check --filter=yamllint
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.34s
     Running `target/debug/qlty check --filter=yamllint`
❌ Failed to fetch base refspecs from remote origin https://github.com/qltyai/qlty

Caused by:
    remote authentication required but no callback set; class=Http (34); code=Auth (-16)
```

qlty-config/src/workspace.rs
```diff
@@ -58,7 +58,11 @@ impl Workspace {
     }
 
     pub fn config(&self) -> Result<QltyConfig> {
-        Builder::full_config_for_workspace(self)
+        let config = Builder::full_config_for_workspace(self);
+        if let Ok(config) = &config {
+            config.print_deprecation_warnings();
+        }
+        config
```
brynary, line 65: 💄 please add newlines around the `if` statement

qlty-config/src/config.rs
```diff
@@ -136,6 +139,41 @@ impl QltyConfig {
...
+                    && source
+                        .repository
+                        .as_ref()
+                        .unwrap()
+                        .starts_with(OLD_DEFAULT_SOURCE_REPOSITORY)
+                {
+                    warn!("qlty.toml default source is a repository-style reference to qltysh.");
+                    eprintln!(
+                        r#"
+{} Warning: your qlty.toml is using a deprecated, repository-based, default source. Please change your default source in your qlty.toml to:
```
brynary, line 162:
```suggestion
{} Warning: qlty.toml is using a deprecated, repository-based, default source.

Please change the default source in your qlty.toml to:
```

qlty-config/src/config.rs
brynary: Sorry, I don't understand this. I was imagining that `OLD_DEFAULT_SOURCE_REPOSITORY` would become an array, and the deprecation warning would fire if either of them matched.

Are you saying that making that change caused the Git error above? I don't think I follow how adding another case to the deprecation conditional could change the behavior.

(Also, to my knowledge, there should be no remaining references to `qltyai/qlty` in our code.)

Tried giving you a ring, please give me a call so we can sort out

noahd1: Ah yeah, I could have been clearer here.

I did write the code you suggested but found, to my surprise, that it didn’t work when I tested it. Digging into why…

There’s code that runs before the new deprecation warning (specifically, I was running `qlty check`) which extracts the sources from the config file before loading the whole config file (where the deprecation code is). That extraction (a git fetch operation) worked fine for the qltysh repository but did not work for the qltyai repository. It threw the error printed above.

I don’t know why this would be — I wasn’t able to replicate the error just calling git clone for that repository from the command line. Something about the way git is called and the environment caused that particular git operation to fail for the old qltyai repository. (I saw, in the code, we do a remote add and then a git fetch, so my test wasn’t exactly a pure replication case.)

I am suspicious of the older repository because it’s archived and could have weird git edge cases. So, I’d assumed (perhaps incorrectly) that this would happen to anyone with a `qltyai/qlty` repository source stanza, and, based on that assumption, thought it doesn’t make sense to add code that isn’t executed, and that furthermore, is a breaking behavior that serves the purpose just as well (if less gracefully).

brynary: Ah, I think I understand... The error that you pasted above came from the _testing_ you tried to do, not simply extending the print conditional itself. That was the critical part I was missing, because I couldn't figure out how the deprecation printer changed behavior :)

The error itself is actually straightforward: `remote authentication required but no callback set`. This is the error that occurs when trying to clone a private repo, which https://github.com/qltyai/qlty is.

So I agree that source cannot be in use and we can skip this suggestion.

noahd1:
Addresssed comments / merged in main to get tests passing.

qlty-cli/tests/cmd/check/using_deprecated_default_source.stdout
brynary, line 1: This file can be removed. We only need stdout / stderr files if we are verifying them. So basically we don't want to snapshot extraneous files otherwise it just creates annoying maintenance work to keep them updated.

brynary (approved):

PR #1357, opened 2024-12-21
Re-run checks if changes are applied

brynary:


brynary:
@lsegal I am seeing a strange [test failure](https://github.com/qltysh/qlty/actions/runs/12440859098/job/34736956853?pr=1357) on Windows:

```
---- expected: stdout
++++ actual:   stdout
   1    1 |  ERRORS: 1 
   2    2 | 
   3    3 | exists  Error  Exited with code 0 in [..].[..]s  .qlty/out/invoke-[..].yaml
   4      -         The plugin crashed for some reason
        4 +         "The plugin crashed for some reason"
```

This is straightforward enough... The output does not match the snapshot, because the actual output is rendering quotes around line 4. However, this does not seem to happen on Mac or Linux, and I am not aware of any behavior which would cause printing output to vary on Windows vs. other platforms.

This is the line where the output is origininating:

```
tw.write_all(format!("\t{}", style(line).red()).as_bytes())?;
```

PR #1361, opened 2024-12-22
Categorize various flavors of `if` in Ruby as `if` nodes

noahd1:
In https://github.com/orgs/qltysh/discussions/1313 @nk-ty noted that `qlty` and `codeclimate` CLIs calculate the cognitive complexity of a method with trailing unless statements differently. After some investigation, I determined that both were incorrectly calculating cognitive complexity. This PR is concerned with the cognitive complexity calculation in qlty.

The following are equivalent ways of expressing an `if` conditional in Ruby:

```ruby
if !foo
    puts "hi"
end
```

```ruby
unless foo
    puts "foo"
end
```

```ruby
puts "hi" if !foo
```

```ruby
puts "hi" unless foo
```

The first was correctly being calculated. The second was an obvious oversight. And the 3rd and 4th ones were less obvious oversights (the parser calls them `if_modifier` / `unless_modifier` nodes not `if` / `unless` nodes.

This PR categorizes all 4 as `if_nodes` which the cognitive complexity calculator uses, and correctly penalizes all 4 as 1 point.

### DIFF
--- a/qlty-analysis/src/lang/ruby.rs
+++ b/qlty-analysis/src/lang/ruby.rs
@@ -149,7 +149,12 @@ impl Language for Ruby {
     }
 
     fn if_nodes(&self) -> Vec<&str> {
-        vec![Self::IF, Self::UNLESS, Self::IF_MODIFIER, Self::UNLESS_MODIFIER]
+        vec![
+            Self::IF,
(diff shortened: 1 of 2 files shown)


brynary (approved):
Nicely done!

PR #1368, opened 2024-12-23
Add a browser based login flow to qlty auth commands

lsegal:
Refactor token management to load retrieve token from browser flow.

Follow-up work: re-introduce setting a manual token via command line parameters.


Cargo.toml
```diff
@@ -106,6 +111,7 @@ time = { version = "0.3.37", features = [
     "formatting",
     "parsing",
 ] }
+tokio = { version = "1.42.0", features = ["full"] }
```
brynary, line 114: We made an architectural decision to use synchronous rather than async Rust and I think that has been working well overall. I would like to continue with that, so I would request that we implement this feature using synchronous Rust.

qlty-cloud/src/token/auth_flow.rs
```diff
@@ -0,0 +1,234 @@
...
+        AppState {
+            original_state: Uuid::new_v4().to_string(),
+            server_handle: Arc::new(Mutex::new(None)),
+            credential_user: DEFAULT_USER.to_string(),
+        }
+    }
+}
+
+#[get("/")]
+async fn auth_flow(
```
brynary, line 51: This is one synchronous Rust HTTP server we can use to avoid pulling in Tokio and Actix:

https://github.com/tomaka/rouille

qlty-cloud/src/token/auth_flow.rs
```diff
@@ -0,0 +1,234 @@
...
+            }
+            Err(e) => {
+                warn!("Failed to store auth token in credential storage: {}", e);
+                HttpResponse::BadRequest()
+            }
+        }
+    }
+}
+
+fn login_server(state: Arc<AppState>) -> Result<(Server, u16)> {
```
brynary, line 83: There's enough code related to the login server, and it is a distinct enough and important concept, that  I might extract it into its own file.

qlty-cloud/src/token/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        Entry::new(SERVICE, &self.user).with_context(|| {
+            format!(
+                "Failed to create keyring entry for service '{}' and user '{}'",
+                SERVICE, self.user
+            )
+        })
+    }
+}
+
+pub fn load_auth_token() -> Result<String> {
```
brynary, line 65: Since this function will invoke the browser flow if a token is not available (which is a big impact) I would probably try to reflect that in the name. Maybe something like load_or_authorize_token.

qlty-cloud/src/token/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+pub const SERVICE: &str = "qlty-cli";
+pub const DEFAULT_USER: &str = "default";
+
+pub struct Token {
+    pub user: String,
+}
+
+impl Default for Token {
+    fn default() -> Self {
+        Self::new(DEFAULT_USER)
```
brynary, line 17: This default user concept was not intuitive to me. Is this trying to represent anonymous usage? If so, can this be represented in the type system as `Option<Token>`?

qlty-cloud/src/token/auth_flow.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+}
+
+#[get("/")]
+async fn auth_flow(
+    query: Query<AuthFlowQueryParams>,
+    state: Data<Arc<AppState>>,
+) -> HttpResponseBuilder {
+    if state.original_state != query.state {
+        warn!(
```
brynary, line 56: Would probably bump this up to error!

qlty-cloud/src/token/auth_flow.rs
```diff
@@ -0,0 +1,234 @@
...
+            "Invalid state parameter in auth flow: {:?} did not match received state {:?}",
+            state.original_state, query.state
+        );
+        HttpResponse::BadRequest()
+    } else {
+        let token = Token::new(&state.credential_user);
+
+        match token.set(&query.code) {
+            Result::Ok(_) => {
+                debug!(
```
brynary, line 66: Would probably bump this up to info! 

Also, this doesn't log anything sensitive right?

qlty-cloud/src/token/auth_flow.rs
```diff
@@ -0,0 +1,234 @@
...
+                    "Auth token stored in credential storage: {}.{}",
+                    SERVICE, state.credential_user
+                );
+                stop_login_server(state);
+                HttpResponse::TemporaryRedirect()
+                    .insert_header((LOCATION, query.redirect_uri.clone()))
```
brynary, line 72: Can / should we validate this redirect destination?

Truthfully I do not entirely understand how this flow works, so I don't know the answer. I just usually ensure strict validation of redirects to limit vulnerability risks whenever possible

qlty-cloud/src/token/auth_flow.rs
```diff
@@ -0,0 +1,234 @@
...
+                    .take()
+            }
+            Err(e) => {
+                warn!("Failed to store auth token in credential storage: {}", e);
```
brynary, line 76: Probably would upgrade to error! level

brynary (commented):
Nice! Added some comments.

A bit of a silly question... Is this adhering to an OAuth spec? If so, which spec?

The prior implementation in the Cloud app was just a system for long-lived tokens. It was also designed to support API access tokens outside of the CLI use case. 

I am wondering how we will approach the data and implementation on the Cloud to support "regular" access tokens as well as OAuth flows.

Finally, on the topic of OAuth, I wonder if we should be supporting Refresh Tokens or not. I suspect that will come up in security reviews.

Cargo.toml
lsegal: This isn't really possible as the actix-web library uses async (as do the majority of Rust http server libraries).

qlty-cloud/src/token/mod.rs
lsegal: I'm not sure I follow this. A "user" value is required for keyring. It is not optional.

Note that this was present in the original implementation of token.rs, not something I changed-- the only modification was adding a const for the user value so that it can be changed during tests.

qlty-cloud/src/token/mod.rs
lsegal: `auth_token` is the vocabulary itself for the token value, so `authorize_token` is kind of ambiguous in this context. `load_or_retreive_auth_token()`?

qlty-cloud/src/token/auth_flow.rs
lsegal: Nothing sensitive is logged, no.

qlty-cloud/src/token/auth_flow.rs
lsegal: Can you explain what you mean by this? This file is already pretty much entirely dedicated to the login server and handlers. We could move auth_via_browser out, which may be a source of confusion here.

qlty-cloud/src/token/auth_flow.rs
lsegal: I think there are some good benefits to relying on the better supported axum/actix libs, one of which is better type-safe support of query string deserialization, something this handler relies heavily on and something that rouille does not seem to support in [its router](https://docs.rs/rouille/latest/rouille/macro.router.html). Testing is also another specific ergonomic benefit, as rouille doesn't seem to have great support there. Finally, rouille doesn't seem to have a very stable approach to stopping the server, something we absolutely require. IMO all of this justifies the small cost of having one or two async uses here.

qlty-cloud/src/token/auth_flow.rs
lsegal: Yes I think we can manage this with some extra logic.

lsegal:
> A bit of a silly question... Is this adhering to an OAuth spec? If so, which spec?

This is not OAuth. OAuth is provided by our login backend. Our API still vends tokens as before. This is an auth flow to retrieve a CLI token as before. The auth mechanism for CLI has not changed by this PR-- only the means to retrieve the token.

Cargo.toml
lsegal: Going to resolve this so that we can have the rest of this convo below when talking specifically about rouille, as it's the only reasonably maintained non-async web server and has its own issues worth discussing.

qlty-cloud/src/token/auth_flow.rs
lsegal: Resolving as I've moved the auth_via_browser to a separate file and what remains is all login-server-related.

qlty-cloud/src/token/mod.rs
brynary: Ahh thanks for clarifying. Sorry when I read this I was forgot about the key ring user concept and conflated Qlty users with key ring users. 

So in essence I confused myself :) We could optionally just rename this variable to something like DEFAULT_KEYRING_USER

brynary:
> This is not OAuth. OAuth is provided by our login backend. Our API still vends tokens as before. This is an auth flow to retrieve a CLI token as before. The auth mechanism for CLI has not changed by this PR-- only the means to retrieve the token.

Got it. Can you help me understand the motivation for naming the `returnTo` URL parameter?

qlty-cloud/src/token/auth_flow.rs
brynary: Got it. So it sounds like you appreciate that Actix+Tokio has additional functionality which we can benefit from. Also, it has more users and is more actively developed than Rouille. Based on these factors and others, Actix has better developer ergonomics

I agree those seem like reasonable advantages of Actix+Tokio. They are somewhat the flip side of Actix vs. Rouille in that Rouille is less active, less features, fewer users.

As far as which way to go, I would like to collect more data before making a decision. Would you be up for doing an alternative implementation in Rouille so that we can compare the two?

In case it is helpful, I did a bit of research and found some capabilities that might be useful to paid with a Rouille experiment:
- serde has a typed query string deserializer in that ecosystem — https://docs.rs/serde-querystring/latest/serde_querystring/
- Rouille has a built-in mechanism for constructing fake Requests for testing — https://docs.rs/rouille/latest/rouille/struct.Request.html#method.fake_http
- It looks like Rouille has a mechanism to be shutdown gracefully using a channel communication — https://docs.rs/rouille/latest/rouille/struct.Server.html#method.stoppable

lsegal:
> Got it. Can you help me understand the motivation for naming the returnTo URL parameter?

This is for compatibility with the OAuth flow that we rely on on the backend. It's not an OAuth flow, but for simplicity and in order to not reinvent protocols, we borrow a subset of the approach. `redirect_uri` is a standard parameter in OAuth 2.0.

brynary (approved):

PR #1370, opened 2024-12-23
Authenticate requests to /fixes endpoint

lsegal:
This required "use AI features" to be enabled in cloud user settings.

Follow-up work:

* Add CLI prompt to automatically enable AI features on user's behalf


qlty-check/src/llm/fixer.rs
```diff
@@ -220,7 +220,7 @@ impl Fixer {
         }
 
         if let Some(path) = issue.path() {
-            let client = Client::default();
+            let client = Client::authenticated()?;
```
brynary, line 223: Will this trigger an authentication flow if the CLI does not have a stored token?

lsegal: It should, yes.

brynary: OK, we may want to tweak this function name to make that explicit.

What are your thoughts on how this should behave when the CLI is run without a TTY?

lsegal: I think there's an interesting question raised as to what the expectation is. There's no actual tty prompt in the case of `qlty auth login`-- the prompting happens externally to the terminal. You could argue that this requires user intervention and thus shouldn't be allowed, but, keep in mind that, on macOS, for example, running `Client::authenticated()` will prompt the user outside of the tty to use the Keychain, i.e.

image

This would happen for every use of authenticated qlty commands, including every `qlty check` command until (if) the user selects "Always Allow", regardless of tty. For that reason I don't know that "no external prompts" is necessarily a standard for how the CLI should behave without tty access.

We could always defer to exiting on the login operation if there is no tty, but there's no guarantee that a pre-logged-in user

As a sidenote, there's still work to be done to allow a user to manually supply a CLI token via some `qlty auth login --token <token|->` command. We could disable the tty once that escape hatch is added (but note that macOS will still perform keychain prompting).

brynary: Thanks for this write up. That's a good point about the keychain being a background operation, I think that suggests that we should _not_ do a TTY check here.

(Side note: I think on Mac you can authorize using the Always Allow and it will skip the login step for that specific binary program. So when developing and recompiling a lot, it will prompt frequently; however, with released versions of the CLI, it should be less noisy.)

This is probably a question we're just going to need to get a feel for in dogfood daily usage and refine based on that experience.

brynary (approved):

PR #1373, opened 2024-12-24
Update cargo lock

marschattha:
Cargo.lock seems to be going out of sync every week.

Usually comes after dependabot PRs are merged, something to do with dependabot?


brynary (approved):
Yeah I think it's the Dependabot PRs that cause this

PR #1374, opened 2024-12-24
Fix language duplication filter patterns

marschattha:
https://github.com/qltysh/qlty/issues/1298

### DIFF
--- a/qlty-config/src/config.rs
+++ b/qlty-config/src/config.rs
@@ -143,7 +143,6 @@ mod test {
     use crate::Workspace;
 
     #[test]
-    #[ignore] // always requires network connection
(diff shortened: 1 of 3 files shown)


brynary:
Can you please add a test which proves this works?

brynary (approved):

qlty-config/src/config.rs
```diff
@@ -143,7 +143,6 @@ mod test {
     use crate::Workspace;
 
     #[test]
-    #[ignore] // always requires network connection
```
marschattha, line 146: No longer needs to be ignored since default source is now packaged as a crate and no longer requires a network call.

PR #1376, opened 2024-12-25
Fallback to upstream when remote commit not present locally

marschattha:
https://github.com/qltysh/qlty/issues/1335

### DIFF
--- a/qlty-cli/src/commands/check.rs
+++ b/qlty-cli/src/commands/check.rs
@@ -9,6 +9,7 @@ use anyhow::Result;
 use autofix::autofix;
 use clap::Args;
 use console::{style, Emoji};
+use duct::cmd;
 use qlty_check::planner::Plan;
 use qlty_check::{planner::Planner, CheckFilter, Executor, Processor, Report, Settings};
 use qlty_cloud::format::JsonFormatter;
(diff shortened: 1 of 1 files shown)


qlty-cli/src/commands/check.rs
```diff
@@ -307,6 +308,16 @@ impl Check {
                 bail!("Missing remote commit ID from Git pre-push input")
             }
 
+            // Check if the remote commit ID is present locally
+            let remote_commit_present_locally = cmd!("git", "cat-file", "-e", remote_commit_id)
```
brynary, line 312: I wonder if this can be done with the `git2` crate instead of needing to shell out to a `git` sub-command

qlty-cli/src/commands/check.rs
```diff
@@ -307,6 +308,16 @@ impl Check {
...
+                .run()
+                .is_ok();
+
+            // If the remote commit ID is not present locally, revert to the upstream branch.
+            if !remote_commit_present_locally {
+                return Ok(self.upstream.clone());
```
brynary, line 318: Would add a `warn!` level log line for this condition

brynary (approved):
Nice, thanks. Approving.

One small improvement would be to use the `workspace.repo()` function which does `Repository:/open()` under the hood. L

The advantage of this is to have a single piece of code responsible for opening a repository for a workspace. (For example if we increased error handling we would be able to do it in one place.)

brynary (approved):

PR #1388, opened 2025-01-02
C# Support

noahd1:


qlty-analysis/src/lang.rs
```diff
@@ -3,6 +3,7 @@ use core::fmt;
 use std::sync::Arc;
 use tree_sitter::{Node, Parser, Query};
 
+mod c_sharp;
```
brynary, line 6: I would probably call this `csharp`, because we are going to have `cpp` represent C++ and it would be a bit odd to have that as `c_p_p`

qlty-cli/tests/lang/csharp/basic.in/FileComplexity.cs
```diff
@@ -0,0 +1,41 @@
...
+                                if (foo % 7 == 0)
+                                {
+                                    if (foo % 11 == 0)
+                                    {
+                                        if (foo % 13 == 0)
+                                        {
+                                            if (foo % 17 == 0)
+                                            {
+                                                Console.WriteLine("Nested!");
+                                            }
```
rossta, line 30: LGTM :shipit:

brynary: Thanks for the extra set of eyes, @rossta !

brynary (approved):
Nice!

PR #1395, opened 2025-01-03
Improve error message when no version found

marschattha:


brynary:
I took this one and brought it a bit further: https://github.com/qltysh/qlty/pull/1394

PR #1404, opened 2025-01-06
Refactor Token struct to simpler API

lsegal:
Allows mocking of keyring::Entry across platforms (no more macos switch)


qlty-cloud/src/auth/credentials.rs
```diff
@@ -0,0 +1,48 @@
+use anyhow::{Context, Result};
+use keyring::Entry;
+use std::sync::{Arc, LazyLock, Mutex};
+
+pub const SERVICE: &str = "qlty-cli";
+pub const DEFAULT_USER: &str = "default";
+
+static ENTRY: LazyLock<Mutex<Option<Arc<Entry>>>> = LazyLock::new(|| Mutex::new(None));
```
brynary, line 8: Do we need both LazyLock and Mutex?

brynary (approved):

qlty-cloud/src/auth/credentials.rs
lsegal: > Do we need both LazyLock and Mutex?

Yes because LazyLock is for initialization, not updates after init. I believe we could use lazy_static to simplify the line, but you'd still need the mutex.

brynary: Got it, thanks. We use lazy_static in a few other places

PR #1405, opened 2025-01-07
Use json! macro to avoid value structs

brynary:
@lsegal I _think_ this patch is behavior equivalent. Since we are specifying the JSON input for the API literally in the code, and using the `json!` macro already, I took it all the way and it allowed slimming down the code by eliminating these serialization structs.

Note: I do think the serialization structs are useful for typing, but only if they are generated from the API specification. So I would be inclined to revisit this down the road once we iterate our approach for client SDK generation.


lsegal:
The point here was definitely to gain some typing checks and avoid typos in JSON, but I can see how that's a pretty shallow check in the context of manual definitions.

lsegal (approved):

PR #1407, opened 2025-01-07
Set token via `qlty auth login --token <TOKEN>`

lsegal:
If "" or "-" are provided, token is read from STDIN

### DIFF
--- a/qlty-cli/src/commands/auth/login.rs
+++ b/qlty-cli/src/commands/auth/login.rs
@@ -1,13 +1,41 @@
+use std::io::stdin;
+
 use crate::{Arguments, CommandError, CommandSuccess};
 use anyhow::Result;
 use clap::Args;
-use qlty_cloud::load_or_retrieve_auth_token;
+use console::style;
+use qlty_cloud::{load_or_retrieve_auth_token, store_auth_token};
 
 #[derive(Args, Debug)]
-pub struct Login {}
+pub struct Login {
+    /// Provide a CLI token manually or use "-" to read from standard input (see https://qlty.sh/user/settings/cli)
+    #[arg(long)]
+    pub token: Option<String>,
+}
 
 impl Login {
     pub fn execute(&self, _args: &Arguments) -> Result<CommandSuccess, CommandError> {
+        if let Some(mut token) = self.token.clone() {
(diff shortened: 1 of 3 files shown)


qlty-cli/src/commands/auth/login.rs
```diff
@@ -1,13 +1,41 @@
...
+        if let Some(mut token) = self.token.clone() {
+            if token == "" || token == "-" {
+                eprint!(
+                    "Generate a token from {} and paste it here.\nToken: ",
+                    style("https://qlty.sh/user/settings/cli")
+                        .underlined()
+                        .green()
+                );
+                let line = &mut String::new();
+                stdin().read_line(line)?;
```
brynary, line 27: Rather than printing and reading input directly, we have been using the dialoguer crate, which I think would make sense to use here for consistency:

https://docs.rs/dialoguer/latest/dialoguer/struct.Input.html

qlty-cli/src/commands/auth/login.rs
```diff
@@ -1,13 +1,41 @@
...
 use anyhow::Result;
 use clap::Args;
-use qlty_cloud::load_or_retrieve_auth_token;
+use console::style;
+use qlty_cloud::{load_or_retrieve_auth_token, store_auth_token};
 
 #[derive(Args, Debug)]
-pub struct Login {}
+pub struct Login {
+    /// Provide a CLI token manually or use "-" to read from standard input (see https://qlty.sh/user/settings/cli)
```
brynary, line 11: This help string may be a bit long and cause wrapping. I generally try to keep these to a length of ~60-70 max

brynary (approved):
Left a couple small comments but overall LGTM!

qlty-cli/src/commands/auth/login.rs
lsegal: The wrapping is handled by clap:

```
Options:
      --token <TOKEN>     Provide a CLI token manually or use "-" to read from standard input (see
                          https://qlty.sh/user/settings/cli)
```

brynary: Ah lovely, didn't realize that. Thanks

PR #1410, opened 2025-01-07
Load token via QLTY_TOKEN if provided

lsegal:


qlty-cloud/src/auth/mod.rs
```diff
@@ -7,10 +7,17 @@ use auth_flow::{launch_login_server, AppState};
...
-use std::{thread, time::Duration};
+use std::{env, thread, time::Duration};
 use tracing::{info, warn};
 
+const TOKEN_ENV_VAR: &str = "QLTY_TOKEN";
+
 pub fn load_or_retrieve_auth_token() -> Result<String> {
+    if let Ok(token) = env::var(TOKEN_ENV_VAR) {
+        // bypass validation when env var is set since this is an intentional override of credential lookup
+        return Ok(token);
```
brynary, line 18: This is a bit of a dumb question, but is it possible to have an env var be set but empty? Should we add an `is_empty()` check here?

brynary (approved):
Added one question about env var semantics, but LGTM! In short, is provided, the env var always wins

lsegal (commented):
> This is a bit of a dumb question, but is it possible to have an env var be set but empty? Should we add an `is_empty()` check here? 

Yes this is a good callout, I had this in another version and forgot to port it.

PR #1411, opened 2025-01-07
Fix coverage paths sanitization issue

marschattha:
Attempt to fix https://github.com/qltysh/qlty/issues/1392

### DIFF
--- a/qlty-coverage/src/transformer.rs
+++ b/qlty-coverage/src/transformer.rs
@@ -1,12 +1,7 @@
 use anyhow::Result;
 use globset::{Glob, GlobSet, GlobSetBuilder};
-use qlty_analysis::utils::fs::path_to_string;
 use qlty_types::tests::v1::{CoverageMetadata, CoverageSummary, FileCoverage};
-use std::{
-    env::current_dir,
-    fmt::Debug,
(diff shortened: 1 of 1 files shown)


brynary (commented):
Can you please add unit tests for this?

brynary:
Also, for the benefit of the future, could you please add content to the PR description with the bug you are working on and the theory behind the fix?

brynary (commented):
This new test looks simple and direct - what I am not understanding is why the code doesn't pass that test before this change. Can you please explain that?

marschattha:
Since we are using file paths to match coverage of different files as strings the strings need to be exact matches, and some reporters can report in absolute paths instead of paths relative to project root or working directory. 

So we had this StripPrefix transformer in place, the issue with the existing implementation was that it was using strings as input and striping strings, which does not account for extra backslashes that might remain after striping base paths. 

For example, if you strip `/home/user/project_root`  from `/home/user/project_root/app/nested/file.rb` as strings you will get `/app/nested/file.rb` where as converting them into paths and then striping will get you `app/nested/file.rb` also that will take care of windows/cross platform compatibility seamlessly.

brynary (approved):
I am still having trouble seeing the bug in the code before this patch. However that is not essential so approving

PR #1412, opened 2025-01-08
Use ignore::WalkParallel for native parallelization instead of rayon

lsegal:
This potentially fixes a deadlock with `rayon::par_bridge` which relies on a shared mutex. If we recursively rely par_bridge through recursive walking, it's _possible_ that this creates a deadlock situation.

Thankfully ignore implements its own parallelized walker which we should be using instead of par_bridge.

I could not detect if this deadlock is actually caused by this line of code, but I believe these stalls did start happening more often when we added more parallelization to CLI.


lsegal:
Ran tests multiple times with no stall failures. Hard to use this as data, but it does seem to bode well for potentially fixing a deadlock.

PR #1413, opened 2025-01-09
Ignore `import` statements  (Java) and `using` directives  (C#) when calculating duplication

noahd1:
Ignores `import` statements (in Java) and `using` directives (in C#) when calculating duplication.


brynary:
Can we please get a test for this which fails before this patch and passes after?

qlty-cli/tests/cmd/smells/ignore_duplication_patterns.stderr
```diff
@@ -0,0 +1,4 @@
+     [0/3] [..]Analyzing all targets... [..]s
+     [1/3] [..]Checking structure of 4 files...  [..]s
+     [2/3] [..]Looking for duplication across 4 files...  [..]s
```
brynary, line 3: I would delete this file, because the snapshot testing framework will not match against stderr if there is no file. This will make the test less brittle

brynary (approved):
I added a line comment about how to make the test less fragile, and approving

PR #1417, opened 2025-01-11
Fix incorrect reference to `plugin_cache` directory

lsegal:
Use `plugin_cachedir` instead.

Fixes a configuration issue from #1408

### DIFF
--- a/.qlty/.gitignore
+++ b/.qlty/.gitignore
@@ -1,5 +1,5 @@
 logs
 out
-plugin_cachedir
+plugin_cache*
 results
 sources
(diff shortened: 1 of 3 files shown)


lsegal:
@brynary been thinking about this patch-- I'm thinking we might want to consider making the .gitignore template exclude-all-with-exceptions rather than the block-specific-files approach we have now. This will cause future-forward upgrade issues for end-users as we potentially add more ignored paths.

tl;dr:

```
*
!qlty.toml
!configs/**
```

This template would be pretty future proofed. If we added any more _tracked_ files we could add those as exclusions in the gitignore when added.

PR #1423, opened 2025-01-13
Gitignore .qlty/* by default and use exclusions for better future proofing

lsegal:
See https://github.com/qltysh/qlty/pull/1417#issuecomment-2587943374 for motivation


brynary (approved):
I agree this is better. I was just starting to look at adding a `db/` sub directory and that would require updates.

PR #1424, opened 2025-01-13
Allow raw coverage files with same name

marschattha:
Right now if someone has two coverage files with same name in different directories it throws the error: `invalid Zip archive: Duplicate filename`

Adding dir path to zip should prevent that while maintaining uniqueness in identifying files.

Reference: https://discord.com/channels/1256822430505373696/1308114848403685406/1328180907554766910


tim-gq:
💯 Thank you!  Looking forward to seeing this shipped.  😄

brynary (approved):

PR #1427, opened 2025-01-14
Fix off-by-1 offset in ripgrep column info

lsegal:
Ripgrep uses 0-indexed column values which is not used by other linter output

### DIFF
--- a/qlty-check/src/parser/ripgrep.rs
+++ b/qlty-check/src/parser/ripgrep.rs
@@ -100,9 +100,9 @@ impl Parser for Ripgrep {
                             path: path.text.clone(),
                             range: Some(Range {
                                 start_line: line_number,
-                                start_column: submatch.start + 1,
+                                start_column: submatch.start + 1, // submatch uses 1-based indexing
                                 end_line: line_number,
-                                end_column: submatch.end + 1,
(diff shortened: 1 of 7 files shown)


larkinscott (approved):

qlty-check/src/parser/ripgrep.rs
```diff
@@ -100,9 +100,9 @@ impl Parser for Ripgrep {
                             path: path.text.clone(),
                             range: Some(Range {
                                 start_line: line_number,
-                                start_column: submatch.start,
+                                start_column: submatch.start + 1,
```
brynary, line 103: Not sure if there's a docs page which references this behavior but  if not could you please add a comment just clarifying the discovery and the need for this +1?

brynary, line 103: Not sure if there's a docs page which references this behavior but  if not could you please add a comment just clarifying the discovery and the need for this +1?

brynary (approved):

marschattha:
The plugin tests might also need updates

qlty-check/src/parser/ripgrep.rs
lsegal: Haven't really found any explicit documentation for this, it was discovered while dogfooding our tools.

PR #1428, opened 2025-01-15
coverage: default prefix stripping to workspace root

lsegal:
Also: various linter fixes in the qlty-coverage package

### DIFF
--- a/qlty-cli/src/commands/coverage/publish.rs
+++ b/qlty-cli/src/commands/coverage/publish.rs
@@ -76,8 +76,7 @@ pub struct Publish {
 impl Publish {
     // TODO: Use CommandSuccess and CommandError, which is not straight forward since those types aren't available here.
     pub fn execute(&self, _args: &crate::Arguments) -> Result<CommandSuccess, CommandError> {
-        let workspace = Workspace::require_initialized()?;
-        let root = path_to_string(workspace.root);
+        let root = path_to_string(Workspace::assert_within_git_directory()?);
         self.print_initial_messages();
 
         let token = match self.load_auth_token() {
--- a/qlty-cli/src/commands/coverage/publish.rs
+++ b/qlty-cli/src/commands/coverage/publish.rs
@@ -2,7 +2,6 @@ use anyhow::Result;
 use clap::Args;
 use console::style;
 use indicatif::HumanBytes;
-use qlty_analysis::utils::fs::path_to_string;
 use qlty_config::version::LONG_VERSION;
 use qlty_config::{QltyConfig, Workspace};
 use qlty_coverage::eprintln_unless;
@@ -51,7 +50,8 @@ pub struct Publish {
     pub transform_add_prefix: Option<String>,
 
     #[arg(long)]
-    /// The prefix to remove from absolute paths in coverage payloads, to make them relative to the project root. This is usually the directory in which the tests were run. Defaults to current working directory.
+    /// The prefix to remove from absolute paths in coverage payloads to make them relative to the project root.
+    /// This is usually the directory in which the tests were run. Defaults to the root of the git repository.
     pub transform_strip_prefix: Option<String>,
 
     #[arg(long, short)]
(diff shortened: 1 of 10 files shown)


marschattha:
If I am reading this correctly, won't it undo https://github.com/qltysh/qlty/pull/1194

qlty-coverage/src/transformer.rs
```diff
@@ -151,14 +151,6 @@ pub struct StripPrefix {
     prefix: PathBuf,
 }
 
-impl Default for StripPrefix {
-    fn default() -> Self {
-        Self {
-            prefix: current_dir().unwrap_or_else(|_| PathBuf::from(".")),
```
marschattha, line 157: Actually instead of removing this why don't we do 
`prefix: Workspace::assert_within_git_directory()?`
Here and we won't have to make any change to publish or transform at the command level and it'll be in one place instead of two?

qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -75,6 +76,8 @@ pub struct Publish {
 impl Publish {
     // TODO: Use CommandSuccess and CommandError, which is not straight forward since those types aren't available here.
     pub fn execute(&self, _args: &crate::Arguments) -> Result<CommandSuccess, CommandError> {
+        let workspace = Workspace::require_initialized()?;
+        let root = path_to_string(workspace.root);
```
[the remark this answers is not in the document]
marschattha: We can scratch this if going with the comment below.

qlty-coverage/src/transformer.rs
lsegal: @marschattha the problem with this approach is that we can't use `?` or any form of error reporting. We would need to manually catch the err and best case log it, or panic, the latter we definitely don't want to be doing, and the former would not allow us to exit out on failure, which we definitely _do_ want to be doing. Because of this, we need to trigger the function in a place where that error can be correctly bubbled up and handled.

lsegal:
I think it could make sense to lower the burden to `assert_within_git_directory`, especially for cloud builds.

qlty-coverage/src/transformer.rs
lsegal: > Also slight bonus would be you can run coverage commands outside of a git repo (although not sure if that would be desirable) by using strip-prefix flags if need be

I think this would be questionably error-prone if allowed, given that any data we collected in this way would be incorrectly stripped unless the user knew about this setting, which is a big discoverability issue. We could warn, but the basis for qlty coverage publish is that it is used within the qlty cloud ecosystem, so I think it's fair to assume there's at least a git repo.

lsegal: As for changing the return: Default trait does not allow this. We would have to switch it out with a separate function. I suppose it's possible to do this, but by the same token, I think it also makes a lot of sense to require a prefix at the usage level-- it will almost never make sense to pass unstripped paths to our coverage uploader, and so I think it's fair to make this requirement explicit at the top of the call stack and "fail early".

It's possible we would want to rename `strip_prefix` to `root_path` to make that more clear.

lsegal: All this said I think I will make the change in Processor to keep this PR simpler and avoid any potential usage regressions, since technically it is a breaking change.

marschattha: > I think this would be questionably error-prone if allowed, given that any data we collected in this way would be incorrectly stripped unless the user knew about this setting, which is a big discoverability issue.

Well if the user is unaware/doesn't use the `strip-prefix` flag, it will default to using the git directory. Only when the user manually specifies a custom `strip-prefix` path it will not lookup the git directory path and use that instead. But yea not sure if there is a good use case for that, so the limit might be better.

> It's possible we would want to rename strip_prefix to root_path to make that more clear.

Agreed with that, that would probably be more clear.

marschattha (approved):

PR #1448, opened 2025-01-22
Error if prepare_script fails

lsegal:


marschattha (approved):

qlty-check/src/planner/invocation_directory.rs
```diff
@@ -265,7 +265,9 @@ mod test {
 
         for (target, result) in targets_results {
             let invocation_directory = invocation_directory_planner.compute(&target).unwrap();
-            assert_eq!(invocation_directory, result);
+            let result_str = path_to_string(result);
+            let result = result_str.split("/./").last().unwrap_or(&result_str);
+            assert!(invocation_directory.ends_with(result));
```
brynary, line 270: Is this change related to the error if prepare_script fails behavior?

qlty-cli/tests/cmd/check/diff_tests/diff_mode.stdout
```diff
@@ -1,6 +1,6 @@
+[..]JOBS: 1[..]
 
- ISSUES: 1 
-
-:0:0
-    0:0  high    exists failed  exists:fail
+Plugin  Result   Targets   Time   Debug File
+exists  Success  1 target[..]
 
+Checked 1 modified file
```
brynary, line 6: If I am reading this diff correctly, this seems to be saying that prior to this PR, this integration test would generate a single issue "exists failed", and after this PR, it produces no issues.

Is that expected?

brynary (commented):
Erroring if the prepare script fails sounds correct, but I am bit unsure about some of the other content in this diff and added some questions

qlty-check/src/planner/invocation_directory.rs
lsegal: This is fixing a macOS temporal unit test failure.

qlty-cli/tests/cmd/check/diff_tests/diff_mode.stdout
lsegal: Yes. The approach was changed such that we're no longer testing a lint failure and instead using --verbose to log how many targets were linted (which is what we care about here).

qlty-cli/src/initializer/scanner.rs
```diff
@@ -446,7 +446,7 @@ latest_version = "1.0.0"
 known_good_version = "1.0.0"
 
 [plugins.definitions.exists.drivers.lint]
-prepare_script = "mkdir ${linter} && echo dir %2 > ${linter}/ls.cmd"
+prepare_script = "mkdir ${linter} && echo dir %2 > ${linter}/ls.cmd || echo dir %2 > ${linter}/ls.cmd"
```
marschattha, line 449: What is the point of a bunch of these  ` || echo dir %2 > ${linter}/ls.cmd` changes?

PR #1461, opened 2025-01-28
build: export qlty.toml with other exported build results

lsegal:


qlty-cloud/src/export/analysis.rs
```diff
@@ -62,6 +68,15 @@ impl AnalysisExport {
             .write_to_file(&self.path.join("issues.json.gz"))?;
 
         GzFormatter::new(JsonEachRowFormatter::new(self.report.stats.clone()))
-            .write_to_file(&self.path.join("stats.json.gz"))
+            .write_to_file(&self.path.join("stats.json.gz"))?;
+
+        GzFormatter::new(CopyFormatter::boxed(Self::qlty_config_path()?))
+            .write_to_file(&self.path.join("qlty.toml.gz"))?;
```
brynary, line 74: Thoughts on if we need the .gz at all? It all ends up in a zip anyway, I think.

lsegal: I added this for completeness. Not sure where we use the zipped version. If we're not using it at all we might consider a dumping all of this zip support in one shot.

PR #1462, opened 2025-01-28
Improve coverage token error handling

marschattha:
https://github.com/qltysh/cloud/issues/5353

image


brynary:
I wouldn't recommend we adopt the JSON:API spec across the board, but for inspiration here is how they model errors as JSON:

https://jsonapi.org/format/#error-objects

marschattha:
Well the `error` field is something that already exists in the coverage api endpoint and it is string in all cases.

PR #1464, opened 2025-01-31
Omit large strings when formating invocations

marschattha:
Sometimes invocations can have extremely large outputs (> 100 MB) which can lead to issues during build data ingestion.
This PR formats build invocations so that we omit any data going into stdout, stderr and tmpfile_contents if they are deemed above limit.


brynary (approved):
It had never occurred to me to handle this at the formatted layer but actually I think it's quite smart!

I was thinking of adjusting the loading of the string data into the invocation however that effectively means we would have invocations which are "lies" and that could cause unexpected issues. This neatly encapsulates the concern to the output step.

PR #1472, opened 2025-02-03
Add negated ignores support

marschattha:
Initial PR for adding negated ignores support


brynary:
@marschattha in other contexts like Classic and Git, does order matter?

marschattha:
I don't believe order matters.

brynary:
ChatGPT o3-mini-high says:

> Yes, order matters. Git processes the .gitignore file top-to-bottom, so a negative pattern (using “!”) only re-includes files that were ignored by rules placed above it.

I think it's right.

I think we should do the same (with localized unit tests to prove it), otherwise users will probably expect it to work like that and we'll get escalatons.

brynary:
@marschatthaets talk about this one when we see each other this week!

brynary:
As discussed, we want to adjust this to respect the ordering of the positive and negative exclusions.

qlty-analysis/src/workspace_entries/matchers.rs
```diff
@@ -234,6 +234,43 @@ fn globs_to_globset(globs: &[String]) -> Result<GlobSet> {
...
+                } else {
+                    None // Immediately exclude the file if a normal ignore rule matches
+                };
+            }
+        }
+
+        Some(workspace_entry)
+    }
+}
+
```
brynary, line 273: This matchers.rs file is getting pretty big. I recommend we make a `matchers/` folder and break it up...

This new IgnoreGroupsMatcher likely deserves it's own Rust file with inline tests, especially since the iter-reverse-match-none-some algorithm is quite tricky.

For the existing matches, I could imagine one file for the boolean (And + Any) matches, one for the Language-related matches (Language Globs, Language Shebang), not sure about the best way to organize the rest.

brynary (commented):
This is looking good. Added some comments.

brynary (approved):
This looks like a good start. A question -- In other contexts like  gitignore or Classic, does the order mattee

marschattha:
> This looks like a good start. A question -- In other contexts like gitignore or Classic, does the order mattee

https://github.com/qltysh/qlty/pull/1472#issuecomment-2643285385 not sure about classic, don't see anything in the [docs](https://docs.codeclimate.com/docs/excluding-files-and-folders)

PR #1475, opened 2025-02-04
Add support for simplecov-json in simplecov parser

marschattha:
Some customers are reported to be using simeplcov-json gem to generate coverage report, which generates it differently from the current and legacy simplecov format that we current support, this PR adds support for simplecov-json format.
https://github.com/vicentllongo/simplecov-json


brynary (approved):
Am I reading this correctly that the single Simplecov parser will now auto-detect "regular"/legacy simplecov vs simplecov-json?

marschattha:
Yes, should we add it as a separate parser? 
Might make things difficult on auto detect.

Previously we were only detecting simplecov and legacy simplecov format and if we got neither it would return empty.
This PR adds a third state(simplecov-json) if first two fail before returning empty if none.

PR #1476, opened 2025-02-05
Fix path target mode when used in subdirectory 

marschattha:
Right now if you apply a formatting change bubbled up from `qlty check` command while in a sub directory, the application fails with an `Result::unwrap() on an Err value`.

The root of that issue lies in the fact that when you apply the format this way, formatter is run with only selected file paths (which are project root relative and not relative to cwd) and TargetMode is set to Path when that happens. Instead of normalizing relative paths based on cwd I think just using absoluete paths is a better approach and less likely to cause any issues.


brynary:
Does this fix only apply to formatting or also issue fixes?

marschattha:
> Does this fix only apply to formatting or also issue fixes?

Just tested with eslint suggestion, worked, not sure about the ai autofixes, don't have that setup.

brynary (approved):

marschattha:
Merging - https://github.com/qltysh/support/issues/7#issuecomment-2644144800

PR #1478, opened 2025-02-05
Change default behavior for missing output

marschattha:
https://github.com/qltysh/qlty/pull/1466 again with `output_missing = "parse"` for all linters except eslint and kubelint


brynary (approved):
Approving. So these snapshot changes look OK?

marschattha:
Added snapshot tests for trufflehog and markdownlint for scenarios where there is no output/issues to report.
Not sure if any other plugins behave similarly but these two do so added their tests.

PR #1481, opened 2025-02-06
Change composer to install packages in worksapce root

marschattha:
For a lack of better way to pass workspace root to composer.rs this is a draft PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ flamegraph.svg
 /.markdownlint.json
 /.rustfmt.toml
 /.yamllint.yaml
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ flamegraph.svg
 /.markdownlint.json
 /.rustfmt.toml
 /.yamllint.yaml
-.DS_Store
--- a/qlty-check/src/tool/php.rs
+++ b/qlty-check/src/tool/php.rs
@@ -16,7 +16,7 @@ use sha2::Digest;
 use std::collections::HashMap;
 use std::env::split_paths;
 use std::fmt::Debug;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use tracing::debug;
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ flamegraph.svg
 /.markdownlint.json
 /.rustfmt.toml
 /.yamllint.yaml
+.DS_Store
```
brynary, line 9: @marschattha you can configure your Git to ignore .DS_Store files everywhere on your system, I think

qlty-check/src/tool/php.rs
```diff
@@ -243,6 +247,7 @@ pub mod test {
             reroute_tools_root(&temp_path, pkg);
 
             let composer = Composer {
+                workspace_root: temp_path.path().to_path_buf(),
```
brynary, line 250: What is confusing to me about this line is that it seems to be assigning `temp_path` to the `Composer`'s `workspace_root`. However, we have another variable `self.workspace_root` so I would expect we would be assigning that to it.

qlty-check/src/tool/php.rs
```diff
@@ -169,7 +172,7 @@ impl Tool for PhpPackage {
         let mut env = HashMap::new();
         env.insert(
             "COMPOSER_VENDOR_DIR".to_string(),
-            path_to_native_string(PathBuf::from(format!("{}/vendor", self.directory()))),
+            path_to_native_string(self.workspace_root.join("vendor")),
```
brynary, line 175: I notice that we seem to construct COMPOSER_VENDOR_DIR here and also on line 79 in `composer.rs` with the same logic. Are these producing the same values?

If so, and the values need to be in sync, we should only compute the value once to avoid a bug if one location changes.

brynary (commented):
I added some specific comments, however I am a bit confused about the overall strategy, specifically the usage of workspace_root vs. temp_path. I think addressing the comments should hopefully help clarify.

However, this is important behavior and necessarily somewhat complex, so I think it would be valuable to add a thorough explanation of how this works as comments to the top of `composer.rs` file

.gitignore
marschattha: Oh yea this wasn't supposed to be commited

qlty-check/src/tool/php.rs
marschattha: Actually I think this is no longer needed anymore, since we are going to be installing packages in the `project_root/vendor anyway`. 
This was here before to tell the plugin the vendor path during invocation runs which was previously in the tool install.

qlty-check/src/tool/php.rs
brynary: Whoops, sorry I missed the `mod test`!

PR #1482, opened 2025-02-07
Copy phpstan config into tool dir

marschattha:
Similar to eslint, phpstan config files also load dependencies relative to config file path, so we should copy the phpstan configs into tool directory where we install the packages.


lsegal (approved):

brynary (approved):
Approving the code, but would like to get confirmation from the customer this works before merging

marschattha:
Merging - https://github.com/qltysh/support/issues/7#issuecomment-2644144800

PR #1491, opened 2025-02-11
Add `--skip-missing-files` flag for coverage publish

marschattha:
With the `--skip-missing-files` flag the publish command will check if files exist on the file system and if not it will ignore it, to have the same behaviour as classic, but is this the best way to handle it?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install cosign
         if: github.event_name != 'pull_request'
-        uses: sigstore/cosign-installer@v3.7.0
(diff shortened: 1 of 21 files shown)


marschattha:
So this is the [PR](https://github.com/codeclimate/test-reporter/pull/382) which introduced this in classic. 

I kind of understand why it was done but I think this is the ideal way of handling the issue. 
That said I have been unsuccessful in finding a better way to handle it, I'll take another look into it.

brynary:
From that PR description, it sounds like there are some known cases where coverage data is in an input file but are for known-non-existent paths.

Is that what is happening for the customer we are investigating?

PR #1498, opened 2025-02-17
Fix qlty.toml clippy version

lsegal:


brynary (approved):
Approving, but do we understand the trigger of the breakage?

lsegal:
@brynary rust was updated because we're on nightly:

```
[runtimes.enabled]
rust = "nightly"
```

We can be a bit more conservative if we want to avoid breakages, but it might be worth the dogfooding?

PR #1505, opened 2025-02-18
tool(ruby): fix linux/arm64 download URL

lsegal:


qlty-check/src/tool/ruby.rs
```diff
@@ -226,7 +226,7 @@ impl Ruby {
                     },
                     System {
                         url:
-                            "https://github.com/ruby/ruby-builder/releases/download/toolcache/ruby-${version}-ubuntu-20.04.tar.gz"
+                            "https://github.com/ruby/ruby-builder/releases/download/toolcache/ruby-${version}-ubuntu-22.04-arm64.tar.gz"
```
brynary, line 229: Is it intentional that this one is 22.04 and the Amd64 one is 20.04?

brynary (approved):

qlty-check/src/tool/ruby.rs
lsegal: Yes. There is no 20.04 support for Ruby/linux/arm.

PR #1509, opened 2025-02-19
Fix bug in DiffSource::build() file metadata calculation

lsegal:
Implementation was grabbing fs::metadata of the root directory instead of the actual file, returning the same metadata for every file causing cache issues.


brynary (approved):
WOW! Great find!!!

PR #1510, opened 2025-02-19
Fix qlty check always asking to format

lsegal:
Update print_unformatted() to only prompt if apply_mode == Ask

### DIFF
--- a/qlty-cli/src/ui/fixes.rs
+++ b/qlty-cli/src/ui/fixes.rs
@@ -10,7 +10,7 @@ use std::io::Write as _;
 use std::{fmt, io::IsTerminal as _, path::Path};
 use tracing::warn;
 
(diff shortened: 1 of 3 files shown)


brynary:
Agreed.

I believe that passing the `--yes` argument triggers `ApplyMode::All`. As a user, I would expect this would apply any formatting corrections without prompting.

Also agreed on the refactoring. In essence, there is a Boolean determination we need to make which is a function of the apply mode and a prompt (when in ask mode).

Approving this PR so we can land this improvement and defer to you on whether you'd like to make any further adjustments.

brynary (approved):

lsegal:
I can make that fix in a follow up, I think that's currently broken with or without this patch so it's separate.

PR #1515, opened 2025-02-20
Change deprecated sources warning to error

marschattha:


brynary (approved):
We should get assistance from support to make a best effort attempt to notify affected customers before we deploy this.

PR #1516, opened 2025-02-20
Handle missing hits on qlty coverage parser

marschattha:
Right now if a qlty coverage file is uploaded with files with no hits, we error, this PR handles that.

Example, if someone transforms a coverage file which includes coverage for files without any line hits and tries to upload the qlty format file generated they will get an error right now.


brynary:
I think this behavior is OK, but what is generating data with no hits array?

marschattha:
Simplest answer would be an empty file, or depending on the reporter it could just be ignoring the whole file as "trivial lines".

brynary (approved):

PR #1517, opened 2025-02-20
Use composer to install php plugins

marschattha:
Instead of using phar files like we have done so far, using composer to install php plugins seems to help remove certain compatibility issues when using extra packages.

Attempts to resolve compatibility issues with php-codesniffer and its extra optional standards.

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -53,18 +53,17 @@ jobs:
           cache-dependency-path: plugins/package-lock.json
 
       - name: Generate cache key
-        run: echo "CACHE_MONTH=$(date +'blowup-cache')" >> $GITHUB_ENV
+        run: echo "CACHE_MONTH=$(date +'%Y-%m')" >> $GITHUB_ENV
 
       - name: Cache qlty tools
         uses: actions/cache@v4
         with:
           path: ~/.qlty
-          key: ${{ runner.os }}-qlty-${{ env.CACHE_MONTH }}
+          key: ${{ runner.os }}-qlty-${{ env.CACHE_MONTH }}-v1
 
       - name: Setup PHP with PECL extension
         uses: shivammathur/setup-php@v2
-        with:
-          extensions: xmlwriter  # Install xmlwriter extension
+        if: contains(matrix.os, 'macos')
 
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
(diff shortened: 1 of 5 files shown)


.github/workflows/plugins.yml
```diff
@@ -63,7 +63,8 @@ jobs:
 
       - name: Setup PHP with PECL extension
         uses: shivammathur/setup-php@v2
-        if: contains(matrix.os, 'macos')
+        with:
+          extensions: xmlwriter  # Install xmlwriter extension
```
brynary, line 67: Can we please change this comment to say the reason for the need for xmlwriter?

brynary (commented):
Looks very interesting. Does this help with the samples issues we've been tracking?

marschattha:
On my local macOS machine I was able to get qlty working as intended on https://github.com/GetDKAN/contracts by using composer to install php-codesniffer instead of phar file. But doesn't seem to be compatible with linux or windows, windows I get but linux is the bigger issue. 

I was able to get it working on local linux machine by manually installing php-xmlwritter extension, which I tried by adding the extension to php setup action, but it seems there was probably something else as well.

.github/workflows/plugins.yml
marschattha: I believe the installation of php plugins is failing in CI because of missing php extensions. Installing php-xmlwriter on my local linux got it working for me.

marschattha:
So it turns out it was just a cache issue in the CI

brynary:
Thanks Arslan. Could you please add a description fof what we expect this to fix and what we don't expect this to fix?

qlty-plugins/plugins/linters/gitleaks/plugin.toml
```diff
@@ -1,7 +1,7 @@
 config_version = "0"
 
 [plugins.releases.gitleaks]
-github = "zricethezav/gitleaks"
```
brynary, line 4: Is this change related?

brynary (approved):

qlty-plugins/plugins/linters/gitleaks/plugin.toml
marschattha: Not related, but it was a one line fix for Gitleaks to get it to stop crashing tests while I was testing.
 I can move this into a separate PR.

marschattha:
Note:
Needed to add 
```
  "config": {
    "allow-plugins": true
  }
```
To the composer.json package file to allow plugins, spent too much time double checking everything today since I deleted all old files.

PR #1518, opened 2025-02-21
Update issue cache semantics for detecthing file changes

lsegal:
Change heuristics to use mtime only when file is "dirty" in the git tree. If not dirty, use the git tree SHA instead, not mtime.

### DIFF
--- a/qlty-check/src/cache.rs
+++ b/qlty-check/src/cache.rs
@@ -2,11 +2,13 @@ use crate::planner::config_files::PluginConfigFile;
 use crate::planner::target::Target;
 use crate::tool::Tool;
 use anyhow::Result;
+use git2::{Repository, Status};
 use itertools::Itertools;
 use prost::Message;
 use qlty_analysis::cache::{Cache, CacheKey, HashDigest};
 use qlty_config::config::PluginDef;
 use qlty_config::version::QLTY_VERSION;
+use qlty_config::Workspace;
 use qlty_types::analysis::v1::Issue;
 use std::sync::Arc;
 use std::{collections::HashMap, fmt::Debug, path::PathBuf};
@@ -91,6 +93,8 @@ pub struct IssuesCacheHit {
 
 #[derive(Debug, Clone)]
 pub struct IssuesCacheKey {
+    repository_sha: Option<String>,
+    dirty_paths: Vec<PathBuf>,
     pub digest: HashDigest,
 }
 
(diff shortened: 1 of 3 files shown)


qlty-check/src/cache.rs
```diff
@@ -271,7 +275,16 @@ impl IssuesCacheKey {
...
 
+        let mut repository_sha: Option<String> = None;
+        let mut dirty_paths = Vec::new();
+        if let Ok(repository) = Workspace::new().and_then(|w| w.repo()) {
+            repository_sha = Self::repository_sha(&repository).ok();
+            dirty_paths = Self::collect_dirty_paths(&repository);
+        }
+
         Self {
+            repository_sha,
```
brynary, line 286: Do we need to add the repository SHA to the digest also?

qlty-check/src/cache.rs
```diff
@@ -91,6 +93,8 @@ pub struct IssuesCacheHit {
 
 #[derive(Debug, Clone)]
 pub struct IssuesCacheKey {
+    repository_sha: Option<String>,
```
brynary, line 96: We may want to clarify this is the root tree SHA (not a commit SHA, which is more common)

qlty-check/src/cache.rs
```diff
@@ -290,18 +303,52 @@ impl IssuesCacheKey {
...
+                self.digest.add("target_sha", sha);
+                return true;
+            }
+        }
+
+        false
+    }
+
+    fn repository_sha(repo: &Repository) -> Result<String> {
+        Ok(repo.head()?.resolve()?.target().unwrap().to_string())
```
brynary, line 337: This unwrap seems a bit risky. Recommend we avoid a potential panic unwrap

brynary (approved):
Added a couple of small comments but generally +1.

We'll surely learn more in the wild but great creative idea for this!

qlty-check/src/cache.rs
lsegal: This is what gets added to the digest.

qlty-check/src/cache.rs
lsegal: I'm not really sure what the difference is. The commit sha should match the tree in the case of clean file nodes in the workspace.

brynary: In terms of naming, the commit SHA and the tree SHA are different values, so that is the potential confusion that I think it would be valuable to resolve.

From reading the code in this PR, it looks like we are using the tree SHA rather than the commit SHA but the libgit2 calls don't make it entirely clear.

Functionally, the tree SHA will behave marginally better than the commit SHA because there are a number of cases like merges which roll the commit SHA but don't change the tree.

qlty-check/src/cache.rs
brynary: Ah thanks. I missed how add_target_sha was taking care of that.

PR #1528, opened 2025-02-25
Compress files into a single coverage.zip and update upload logic accordingly

noahd1:
Instead of uploading 4 separate files for coverage uploads, this uploads a single zip file containing the 4 files.

This cannot be merged until a server side component which knows how to read the single zip file is merged.

### DIFF
--- a/qlty-cloud/src/export/coverage.rs
+++ b/qlty-cloud/src/export/coverage.rs
@@ -70,18 +70,23 @@ impl CoverageExport {
             .cloned()
             .collect();
 
-        compress_files(raw_file_paths, &directory.join("raw_files.zip"))
+        compress_files(raw_file_paths, &directory.join("raw_files.zip"))?;
+
+        let files_to_zip = vec![
+            "report_files.json.gz",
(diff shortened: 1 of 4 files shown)


qlty-cloud/src/export/coverage.rs
```diff
@@ -70,18 +70,23 @@ impl CoverageExport {
...
             .collect();
 
-        compress_files(raw_file_paths, &directory.join("raw_files.zip"))
+        compress_files(raw_file_paths, &directory.join("raw_files.zip"))?;
+
+        let files_to_zip = vec![
+            "report_files.json.gz",
+            "file_coverages.json.gz",
+            "metadata.json",
+            "raw_files.zip",
```
brynary, line 79: What do you think of including raw files directly instead of in a nested zip?

noahd1: Considered (and can still do). The raw files retain the folder structure from the client, so I'd probably want to move them to a new top level folder.

brynary (approved):
Nice simplification!

PR #1533, opened 2025-02-26
Fix fmt path issues with prefixes

marschattha:
Right now path for formatters when using prefixes were not matching correctly causing issues during both check and fmt. 
This PR fixes that.

### DIFF
--- a/qlty-check/src/executor/invocation_result.rs
+++ b/qlty-check/src/executor/invocation_result.rs
@@ -16,7 +16,7 @@ use qlty_types::analysis::v1::{
 use serde::Serialize;
 use std::{collections::HashMap, path::PathBuf};
 use std::{process::Output, sync::Arc};
-use tracing::{debug, error};
+use tracing::{debug, error, info};
 
 #[derive(Debug, Clone)]
 pub struct InvocationResult {
@@ -280,9 +280,18 @@ impl InvocationResult {
     fn handle_output_rewrite(&mut self) -> Result<()> {
         let mut formatted = vec![];
 
-        for workspace_entry in self.plan.workspace_entries.iter() {
-            let workspace_path = self.plan.workspace.root.join(&workspace_entry.path);
-            let staged_path = self.plan.target_root.join(&workspace_entry.path);
+        for workspace_entry in self.invocation.target_paths.iter() {
+            let prefixed_path = if let Some(prefix) = &self.plan.plugin.prefix {
+                PathBuf::from(prefix).join(&workspace_entry)
+            } else {
+                PathBuf::from(&workspace_entry)
+            };
+
+            let workspace_path = self.plan.workspace.root.join(&prefixed_path);
+            let staged_path = self.plan.target_root.join(&workspace_entry);
+
+            info!("workspace_path file {:?}", &workspace_path);
+            info!("staged_path file {:?}", &staged_path);
 
             let workspace_contents = match std::fs::read_to_string(&workspace_path) {
                 Ok(content) => content,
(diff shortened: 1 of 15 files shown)


qlty-check/src/executor/invocation_result.rs
```diff
@@ -280,9 +280,18 @@ impl InvocationResult {
...
 
-        for workspace_entry in self.plan.workspace_entries.iter() {
-            let workspace_path = self.plan.workspace.root.join(&workspace_entry.path);
-            let staged_path = self.plan.target_root.join(&workspace_entry.path);
+        for workspace_entry in self.invocation.target_paths.iter() {
+            let prefixed_path = if let Some(prefix) = &self.plan.plugin.prefix {
+                PathBuf::from(prefix).join(workspace_entry)
+            } else {
+                PathBuf::from(&workspace_entry)
+            };
```
brynary, line 288: There's a pattern we seem to do twice in this PR which is to prefix-or-not a path based on a plugin definition's prefix field.

I would extract an and reuse function which accepts a plugin definition and a path and performs this operation. (I would _not_ put this function on PluginDef itself because that is intended to be a data transfer object or a configuration object and not get a bunch of behavior added to the impl)

qlty-check/src/executor/invocation_result.rs
```diff
@@ -296,9 +305,10 @@ impl InvocationResult {
                 .with_context(|| format!("Failed to read staged file {:?}", &staged_path))?;
 
             if workspace_contents != staged_contents {
+                info!("Rewriting file {:?}", &workspace_path);
```
brynary, line 308: Nice. This is a highly valuable log line to have and I think info is the right level given the importance of the operation (and relative small number)

qlty-check/src/executor/invocation_result.rs
```diff
@@ -280,9 +280,18 @@ impl InvocationResult {
...
-            let staged_path = self.plan.target_root.join(&workspace_entry.path);
+        for workspace_entry in self.invocation.target_paths.iter() {
+            let prefixed_path = if let Some(prefix) = &self.plan.plugin.prefix {
+                PathBuf::from(prefix).join(workspace_entry)
+            } else {
+                PathBuf::from(&workspace_entry)
+            };
+
+            let workspace_path = self.plan.workspace.root.join(&prefixed_path);
+            let staged_path = self.plan.target_root.join(workspace_entry);
```
brynary, line 291: Aside from the log lines, are we using both of these variables?

qlty-check/src/executor/invocation_result.rs
```diff
@@ -280,9 +280,18 @@ impl InvocationResult {
...
+
+            info!("workspace_path file {:?}", &workspace_path);
+            info!("staged_path file {:?}", &staged_path);
```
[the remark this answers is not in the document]
marschattha: Oh sorry was using this while debugging.

qlty-check/src/executor/invocation_result.rs
marschattha: Yes they are being used in existing code below.

qlty-check/src/executor/invocation_result.rs
marschattha: While doing this, realized some of the variable naming was confusing and there was another bug hiding in there that effected file names when reporting a formatters issues during check run.

qlty-cli/tests/cmd/network/check/prefix.in/.qlty/qlty.toml
```diff
@@ -0,0 +1,13 @@
+config_version = "0"
```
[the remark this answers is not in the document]
brynary: Thanks. What do you think about disconnecting this from networking by using a dummy formatted instead of a real formatter? For example we could have a fake formatter with a `script` that is something like `sed`, or `tr`, or `echo foo > ${target}`? (One question would be something that can work on windows.)

brynary (approved):

qlty-cli/tests/cmd/network/check/prefix.in/.qlty/qlty.toml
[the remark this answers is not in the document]
marschattha: Gonna merge this in and try on a separate PR.

PR #1534, opened 2025-02-28
Allow --skip-errored-plugins to handle install errors

marschattha:


qlty-check/src/tool.rs
```diff
@@ -154,6 +155,10 @@ pub trait Tool: Debug + Sync + Send {
         Ok(())
     }
 
+    fn is_installed(&self) -> bool {
+        self.tool_type() == ToolType::NullTool || (self.donefile_path().exists() && self.exists())
```
brynary, line 159: Instead of doing the tool_type check here, can we override `fn is_installed()` in `NullTool` to be `true`?

brynary (approved):

qlty-check/src/tool.rs
marschattha: Yes that makes more sense

PR #1546, opened 2025-03-11
Use /fixes/batch API to batch AI fixes

lsegal:
* Updates IssueTransformer to support batched transform
* Batch up to 15 issues per file (per tool) at a time

### DIFF
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -692,9 +692,11 @@ fn run_invocation(
                 return;
             }
 
+            let mut issues = file_result.issues.clone();
             for transformer in transformers {
-                file_result.issues = transformer.transform_batch(file_result.issues.clone());
+                issues = transformer.transform_batch(issues);
             }
+            file_result.issues = issues;
         });
     }
 
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -692,27 +692,9 @@ fn run_invocation(
                 return;
             }
 
-            let mut issues = file_result.issues.clone();
             for transformer in transformers {
-                if let Some(group_issues) = transformer.transform_batch(&issues) {
-                    issues = group_issues;
-                }
+                file_result.issues = transformer.transform_batch(file_result.issues.clone());
             }
-
-            file_result.issues = issues
-                .par_iter()
-                .cloned()
-                .filter_map(|mut issue| {
-                    for transformer in transformers {
-                        if let Some(transformed_issue) = transformer.transform(issue) {
-                            issue = transformed_issue;
-                        } else {
-                            return None;
-                        }
-                    }
-                    Some(issue)
-                })
-                .collect();
         });
     }
 
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -692,8 +692,14 @@ fn run_invocation(
                 return;
             }
 
-            file_result.issues = file_result
-                .issues
+            let mut issues = file_result.issues.clone();
+            for transformer in transformers {
+                if let Some(group_issues) = transformer.transform_batch(&issues) {
+                    issues = group_issues;
+                }
+            }
+
+            file_result.issues = issues
                 .par_iter()
                 .cloned()
                 .filter_map(|mut issue| {
--- a/qlty-check/src/llm/fixer.rs
+++ b/qlty-check/src/llm/fixer.rs
@@ -2,7 +2,7 @@ use crate::planner::Plan;
 use crate::source_reader::SourceReader;
 use crate::ui::ProgressBar as _;
 use crate::{executor::staging_area::StagingArea, Progress};
-use anyhow::{bail, Result};
+use anyhow::Result;
 use itertools::Itertools;
 use lazy_static::lazy_static;
 use qlty_cloud::Client;
@@ -32,7 +32,7 @@ pub struct Fixer {
     progress: Progress,
     staging_area: StagingArea,
     r#unsafe: bool,
-    attempts_per_file: Arc<Mutex<HashMap<Option<String>, AtomicUsize>>>,
+    attempts_per_file: Arc<Mutex<HashMap<String, AtomicUsize>>>,
     total_attempts: Arc<AtomicUsize>,
 }
 
@@ -42,16 +42,26 @@ impl IssueTransformer for Fixer {
     }
 
     fn transform_batch(&self, issues: &[Issue]) -> Option<Vec<Issue>> {
-        if issues.is_empty() {
-            return None;
-        }
-        Some(
-            issues
-                .chunks(MAX_BATCH_SIZE)
-                .flat_map(|chunk| self.fix_issue(chunk))
-                .flatten()
-                .collect_vec(),
-        )
+        let no_path_issues = issues
+            .iter()
+            .filter(|issue| issue.path().is_none())
+            .cloned()
+            .collect_vec();
+        let processed_issues = issues
+            .iter()
+            .filter(|issue| issue.path().is_some())
+            .cloned()
+            .into_group_map_by(|issue| issue.path().unwrap_or_default())
+            .iter()
+            .map(|(path, issues)| {
+                issues
+                    .chunks(MAX_BATCH_SIZE)
+                    .flat_map(|chunk| self.fix_issue(path, chunk))
+                    .collect_vec()
+            })
+            .collect_vec()
+            .concat();
+        Some([no_path_issues, processed_issues].concat())
     }
 
     fn clone_box(&self) -> Box<dyn IssueTransformer> {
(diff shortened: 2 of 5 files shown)


brynary (commented):
The `transform_batch` function in the `IssueTransformer` interface feels like a bit of a bolt-on.

I think we should consider alternative ways to structure the code so that we can keep the `IssueTransformer` focused, for example by limiting it to a single issue use.

lsegal:
@brynary would you recommend adding a separate transformer step? I think the "bolt on" is an attempt to keep interfaces minimal for the transformer phase.

brynary:
I was thinking of a few ways to adjust the abstractions, for example having a BatchIssueTransformer trait.

However, that seems premature, as this is our only use case for that currently.

Instead, what if we simply adjust the fixer so that it is not an IssueTransformer at all?

That would allow us to remote the `transform_batch` from the IssueTransformer trait while retaining it on the fixer.

lsegal:
> Instead, what if we simply adjust the fixer so that it is not an IssueTransformer at all?

I think this would be a step backwards. We want this to run during the executor step so that we take advantage of other transformers and benefit from issue caching, as this explicitly _is_ an "issue transformer"-- in that it is transforming and mutating issues.

A BatchIssueTransformer trait would be fine, but would require a separate list of transformers to run. I'm fine with this, but the "premature" thing was the part that got me to decide to include it in our existing transformer stack. Separating these functions into separate traits should be a simple future refactor should we need it, but my guess is we wouldn't. Conceptually it makes sense for an "IssueTransformer" to support single or multi issue transforms.

brynary:
OK. Let me try to describe my concerns a bit better and maybe we can find another solution...

My first concern is that the `fn transform_batch` with the default implementation of `None` in the `IssueTransformer` trait is undesirable because a critical part of IssueTransformer's semantics is that `None` means "drop the issue." So, by having the `None` default implementation, it looks like it is saying that all IssueTransformers, by default, should drop issues.

My second concern is that by having both `transform` and `transform_batch` defined on every IssueTransformer, it is ambiguous when each transformer is run. We are sort of using the batch vs. non-batch factor to subtely determine _when_ in the flow each transformer is executed, but the concept of individual vs. batch processing is orthogonal to the processing stage.

How could we avoid these two issues?

lsegal:
> My first concern is that the fn transform_batch with the default implementation of None

I think this concern is being formulated in a very specific way such that the implementation detail of return type matters more when it should not. It sounds to me like the underlying root need here is to be able to "drop issues". The None is an important implementation detail to making this possible for single issues, **but this is not needed for batch issues**.

Specifically: it is currently possible in the existing implementation to "drop issues"-- simply `Some(vec![])` and you will drop all issues. If you want to drop certain issues, you can even return `Some(issues.iter().filter(...).collect_vec())`. Basically because this is a list and not a single issue, we do not need to rely on a sentinel "no issue" value, we can just return the list omitting the issue. 

In other words, `None` is an _implementation detail_ of `transform()`-- we could have used a Result or even an enum type to be more explicit about whether an issue should be dropped or not:

```rust
enum TransformResult {
    Keep,
    Discard,
    Modify(Issue),
}

pub trait IssueTransformer: Debug + Send + Sync + 'static {
    fn initialize(&self) {}
    fn transform(&self, issue: Issue) -> TransformResult;
}

// later
pub fn some_fn(issue: Issue) {
    match transformer.transform(issue) {
        TransformResult::Modify(new_issue) => { ... }
        TransformResult::Discard => { ... }
        TransformResult::Keep => { ... }
    }
}
```

Instead of the above, None in the `transform_batch` function serves a different purpose: it allows for short-circuiting the transform operation without [cloning the full list of issues on each transformer](https://github.com/qltysh/qlty/pull/1546/files#diff-7b14e6e2c53290cc0e0f9188cedb865309516f4289ae4d5c8117c5ff086a5531R696-R700) that may not implement batching (95% of them). Because most transformers do not need to implement a batch operation, the default implementation is None, thereby meaning "do not run this".

We could change this to always run this function and return a clone-- or we could split the trait into a separate list such that not all transformers need to call this method-- or we could use the enum approach as above for all transformers which would make this much more explicit. I'm ok with any approach.

> My second concern is that by having both transform and transform_batch defined on every IssueTransformer, it is ambiguous when each transformer is run. 

I think this is a reasonable concern, although I'm unsure how you would solve this without some explicit "staged" state machine runner. Splitting the transformers into `plan.transformers` && `plan.batch_transformers` wouldn't necessarily be any more descriptive about when these transformers ran.

In reality, it shouldn't really matter "when" they are run, as you're typically only implementing one or the other, and in the off case where you implement both, it would probably be something that should run order-independently anyway (order is not guaranteed by _any group_ of transformers).

I'm specifically unclear about this:

> the concept of individual vs. batch processing is orthogonal to the processing stage.

Can you elaborate? I'm trying to understand why this is relevant to processing order, or why single vs multi processing order matters that much (given that you already acknowledge that these transformers will mostly only process one OR the other type). Seems to me like a refactor to separate traits would basically solve the confusion around the order problem, unless I'm missing something.

tl;dr:

1. Beef up the interface to return an enum type.
2. Split the traits into 2 so that there is less confusion about ordering.

brynary:
Thanks, this is quite helpful.

Worth noting -- I somewhat mid-read the code and I thought that the location of the batch transformation loop was not adjacent to the loop for the individual transformations. I now see them at they are adjacent which sufficiently reduces my concern about subtle ordering issues.

AFAIK we are not relying on any guarantees for the order that the transformers run in. Assuming that stays accurate, I'm not worried about the ordering.

The enum is super explicit, which is nice. Overall, I think I probably underuse enums relative to what would be ideal, as they feel like the best part of the Rust type system in many ways.

(For example, we could have an enum for a FilePath which has Absolute and Relative variants -- that could save some bugs.)

Let me float one last idea that does not require an enum and does not require new trait...

What if we adjusted the IssueTransformer so that all transformers are executed as batch transformers. My "ah hah" moment is that individual transforms can be expressed via batch transform but not the reverse.

So the IssueTransformer could have a default implementation for `transform_batch` which calls `transform` on every issue. The default implementation of `transform` would simply be to consume and return the issue -- the "keep" behavior. The result is that a default IssueTransformer is a no-op.

Then we can override whichever function we need depending on the transformer, and the executor can execute them all in a loop (which it does anyway, the executor just does the outer N loops and the inner M loops move into the transformers themselves.

What do you think?

lsegal:
I like this :)

qlty-check/src/executor.rs
```diff
@@ -692,21 +692,9 @@ fn run_invocation(
...
-                            issue = transformed_issue;
-                        } else {
-                            return None;
-                        }
-                    }
-                    Some(issue)
-                })
-                .collect();
+            for transformer in transformers {
+                file_result.issues = transformer.transform_batch(file_result.issues.clone());
```
brynary, line 696: Is the clone() needed here?

qlty-config/src/config/issue_transformer.rs
```diff
@@ -1,9 +1,21 @@
...
-    fn transform(&self, issue: Issue) -> Option<Issue>;
+
+    fn transform(&self, issue: Issue) -> Option<Issue> {
+        Some(issue)
+    }
+
+    fn transform_batch(&self, issues: Vec<Issue>) -> Vec<Issue> {
+        issues
+            .par_iter()
+            .filter_map(|issue| self.transform(issue.clone()))
```
brynary, line 15: I don't think we should need to clone an individual issue in order to transform it.

I'm not sure if it'll work with the filter_map construct exactly the same, but the previous code avoided the clone of each issue.

lsegal: The previous code used `.cloned()` which cloned all elements implicitly, so this is the same result. We could use `.cloned()` here as well, I'm impartial.

qlty-check/src/executor.rs
lsegal: Technically in this usage it is. We could modify a bit to avoid the clone, but I'm unclear if this actually avoids "cloning".

brynary (approved):
Looks great!

qlty-check/src/llm/fixer.rs
```diff
@@ -30,20 +32,33 @@ pub struct Fixer {
...
-        if !self.reached_max_fixes(&issue) {
-            issue = self
-                .fix_issue(&issue)
-                .inspect(|issue| self.update_max_fixes(issue))
-                .unwrap_or(issue);
-        }
-
-        Some(issue)
+    fn transform_batch(&self, issues: Vec<Issue>) -> Vec<Issue> {
+        let no_path_issues = issues
```
brynary, line 41: Issues without a path are possible but quite rare. So this block of code to separate them out stood out to me. Could we handle the issues without paths in the main iterator instead to avoid needing to keep track of two vectors?

lsegal: Yea I like this idea.

PR #1549, opened 2025-03-12
fix(fixer): resolve full path to file when patching

lsegal:


brynary (approved):

marschattha:
If you change https://github.com/qltysh/qlty/blob/3d5a78755ec7089d3f03f51dd8baa73aafc0b548/qlty-cli/src/ui/fixes.rs#L63 to `path: path_to_string(full_path),` that will avoid passing workspace around.

lsegal:
> that will avoid passing workspace around.

Tried this but we use that candidate.path when printing the prompts to stdout for the user, so this changes the output to display full paths in the prompts, which isn't ideal.

marschattha:
Also had the workspace root in the argument if that was the only thing needed.
https://github.com/qltysh/qlty/blob/3d5a78755ec7089d3f03f51dd8baa73aafc0b548/qlty-cli/src/ui/fixes.rs#L43

PR #1550, opened 2025-03-12
Fix GitDiff calculation of repository paths

lsegal:
* GitDiff::compute path is passed as the root of repository, but we use the .parent() when computing diffs
* Our test `sample_repo()` returns a Repository object whose `.path()` returns the subdir `.git/` directory, meaning we are testing the function differently from how it is used.

tldr: the parent() is unnecessary. This causes matching issues when running from subdirectories (issues are filtered out because they do not match the path). To be honest, I'm not 100% sure why this works when running from root, but we have been passing the wrong path for those files too (i.e. for `/path/to/qltysh/qlty` we used `/path/to/qltysh` as the repo dir).

### DIFF
--- a/qlty-analysis/src/git/diff.rs
+++ b/qlty-analysis/src/git/diff.rs
@@ -81,10 +81,8 @@ impl GitDiff {
         debug!("Found {} changed files", changed_files.len());
         trace!("Changed files: {:?}", changed_files);
 
-        let line_filter = DiffLineTransformer::new(Self::plus_lines_index(
-            &diff,
-            path.parent().unwrap().to_path_buf(),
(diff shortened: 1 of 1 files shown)


brynary (approved):
Nice find! Did this cause a breakage or did you find it from code reading?

If I am understanding properly that the `path` arg to `GitDiff::compute` is supposed to be a Git repository root, what do you think of renaming the variable to `repository_root` to increase the chance that an error like this would stand out?

lsegal:
The error is present in the CLI, it was discovered while debugging another issue (#1549). As far as renaming the variable, I think that's fine.

PR #1553, opened 2025-03-13
Add stylelint 16 support

marschattha:
Credit goes to fonji

https://discord.com/channels/1256822430505373696/1349319727406649374/1349706080954875926


brynary (approved):

brynary:
@marschattha the original error that was reported on this one was:

```
stylelint  Parse error  EOF while parsing a value at line 1 column 0
```

What do you think about adjusting the `output_missing` behavior for stylelint from `parse` to `error`, for one or more versions of stylelint?

PR #1555, opened 2025-03-14
Update composer tool to install packages on first attempt

marschattha:
Thanks to visibility from debug files it seems previously we were retrying php installation process at least once for every php installation that had extra packages, this PR makes sure the composer.json file in php tool directory is correctly updated on first try and packages are correctly installed without needing a second try.

### DIFF
--- a/qlty-check/src/tool/php.rs
+++ b/qlty-check/src/tool/php.rs
@@ -281,10 +281,9 @@ pub mod test {
                             temp_path.path().display(),
                             composer.directory_name()
                         )),
-                        "install",
+                        "update",
                         "--no-interaction",
-                        "--ignore-platform-reqs",
-                        "--no-plugins"
+                        "--ignore-platform-reqs"
                     ]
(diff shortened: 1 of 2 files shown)


brynary:
@marschattha I've been concerned about the retry-by-default installation logic for awhile, and this is another case of it causing a problem.

The fix here looks good, but what do you think of going further and also limiting the installation-retry logic to only be used in very specific, known cases?

marschattha:
If memory serves correctly, I think checkov is another where this is needed

brynary (approved):
OK, we should handle any changes to retry policy separately. For that, my goal would to be as narrow as possible (e.g. if a download has unstable networking, maybe we have a `download_retries = 1` option which only retries just that specific part. Anyways, let's table that for now.

As for the content in this PR, I don't have a great understanding of how this particular works, but I don't see anything wrong with these changes and my understanding is that the integration tests are solid. So approving

PR #1556, opened 2025-03-14
Add installation errors to cli output

marschattha:


brynary:
@marschattha I did a quick read through of this one and would like to discuss a bit the new `InstallationError` struct as well as the change of the `fn install()` return type to `Result<(Vec<Message>, Vec<InstallationError>)>`

Jumping on a call in a bit so maybe we can pick this up tomorrow

brynary (commented):
@marschattha with this PR, what is sort of nagging me is how this is how we now have to pass around an extra `installation_errors` field in many places.

However, we already are pushing a `Message` which nealry basically all the information that is in the `InstallationError` struct.

To simplify this PR, what I would recommend is that we get rid of `InstallationError` and when we print the output instead of iterating over `report.installation_errors` we iterate over the `messages` where the `ty` field is `"executor.install.error"`.

For now I think we can just leave the Debug Files column blank and we can potentially revisit the debug files idea for the installations later but I think we can remove that from this PR for simplicity

marschattha:
https://github.com/qltysh/qlty/pull/1556/commits/94335c401a8321ca3d54c61efcae05b6fd3e97da

qlty-cli/src/ui/invocations.rs
```diff
@@ -217,6 +217,24 @@ pub fn print_invocations(
...
+    //             style("Installation error").red(),
+    //             installation_error.error_message.as_ref().unwrap(),
+    //             style(installation_error.directory.clone().unwrap_or_default())
+    //                 .dim()
+    //                 .underlined(),
+    //         )
+    //         .as_bytes(),
+    //     )
+    //     .unwrap();
+    // }
```
brynary, line 236: Commented code

qlty-check/src/executor.rs
```diff
@@ -90,6 +89,14 @@ impl Executor {
...
                         ..Default::default()
                     });
+
+                    // println!("Error installing tool {}: {}", name, err.to_string());
+
+                    // jobs.push(InstallationError {
+                    //     tool_name: name.clone(),
+                    //     error_message: Some(err.to_string()),
+                    //     directory: Some(directory),
+                    // });
```
brynary, line 99: Commented code

qlty-check/src/executor.rs
```diff
@@ -59,19 +59,18 @@ impl Executor {
...
-            result.messages.push(message);
-        }
+        result.messages.extend(install_messages);
 
         Ok(result)
     }
 
     pub fn install(&self) -> Result<Vec<Message>> {
         let mut install_messages = vec![];
+        // let mut jobs = vec![];
```
brynary, line 73: Can you please remove the commented code in this one? (A couple more locations below)

brynary (approved):

brynary:
Approved, with a couple comments

PR #1579, opened 2025-03-26
Update Radarlint support

marschattha:


marschattha:
Radarlint related tests are passing now, only failure is due to the ruby issues.

brynary (approved):
This looks good. I think the last thing is changing the parser to use a dot instead of a forward slash in the rule key.

PR #1584, opened 2025-03-27
Disable batching for php-cs-fixer

marschattha:
Batching with php-cs-fixer requires a config file to be present, we can either change suggested to "config" or disable batching.

Disabling batching will make it slower, but will give more value


brynary (approved):
Seems like the way to go but a great case for docs someday. For now can you add a comment?

PR #1586, opened 2025-03-29
Improve Radarlint parser robustness

brynary:
## Summary
- Update Radarlint parser to use i32 instead of u32 for TextRange
- Convert negative values to 0 instead of passing them through
- Make rule_key, primary_message, and file_uri required fields
- Return errors on invalid JSON instead of skipping lines silently
- Add comprehensive test coverage for edge cases

## Test plan
- Added unit tests for each edge case
- Ran all existing tests to ensure backward compatibility

### DIFF
--- a/qlty-check/src/parser/radarlint.rs
+++ b/qlty-check/src/parser/radarlint.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
 struct RadarlintIssue {
+    #[serde(default = "default_severity")]
     severity: String,
     rule_key: String,
     primary_message: String,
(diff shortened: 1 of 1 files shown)


qlty-check/src/parser/radarlint.rs
```diff
@@ -53,11 +62,33 @@ impl Parser for Radarlint {
...
+                            let end_line = if text_range.end_line < 0 {
+                                0
+                            } else {
+                                text_range.end_line as u32
+                            };
+                            let end_column = if text_range.end_line_offset < 0 {
+                                0
+                            } else {
+                                text_range.end_line_offset as u32
+                            };
```
brynary, line 85: Can we use `max` to simplify this pattern? If so, let's use that for all four of these assignments. And, let's get rid of the local variables and just put it into the `Range { ... }` constructor

brynary:
I've simplified the negative value handling using max() as suggested and removed the local variables. The code is now much cleaner and still handles the edge cases correctly.

PR #1590, opened 2025-03-29
Fix PHP tool unwrap calls

brynary:
# Error Handling Improvements

This PR improves error handling throughout the codebase by replacing unwraps with proper error handling using the Result type.

## Changes

1. ✅ Updated InvocationResult to handle env() errors by properly propagating them with context
2. ✅ Updated initialize_installation to return Result<Installation> and use the ? operator to return Err if tool.env() is an Err
3. ✅ Updated fn extra_env_vars to return Result<HashMap<String, String>>
4. ✅ Modified fn extra_env_vars in ruby.rs to treat errors from self.package_file_envs as fatal using the ? operator

The PR makes error handling more consistent throughout the codebase, ensuring that errors are properly propagated and displayed to the user with appropriate context.

### DIFF
--- a/qlty-check/src/tool/php.rs
+++ b/qlty-check/src/tool/php.rs
@@ -65,13 +65,15 @@ impl Tool for Php {
     }
 
     fn extra_env_paths(&self) -> Vec<String> {
-        split_paths(
-            &std::env::var("PATH")
-                .with_context(|| "PATH not found for php runtime")
-                .unwrap(),
-        )
-        .map(path_to_native_string)
-        .collect_vec()
+        match std::env::var("PATH") {
+            Ok(path) => split_paths(&path)
+                .map(path_to_native_string)
+                .collect_vec(),
+            Err(_) => {
+                debug!("PATH not found for php runtime, using empty path");
+                Vec::new()
+            }
+        }
     }
 }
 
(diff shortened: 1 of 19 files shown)


qlty-check/src/tool/php.rs
```diff
@@ -65,13 +65,15 @@ impl Tool for Php {
...
-        .collect_vec()
+        match std::env::var("PATH") {
+            Ok(path) => split_paths(&path)
+                .map(path_to_native_string)
+                .collect_vec(),
+            Err(_) => {
+                debug!("PATH not found for php runtime, using empty path");
+                Vec::new()
+            }
+        }
```
brynary, line 76: Let's change the return type for `fn extra_env_paths` from `Vec<String>` to `Result<Vec<String>>` in the `Tool` trait. Then, return an `Err` here if the PATH variable is unavailable

qlty-check/src/tool/php/composer.rs
```diff
@@ -64,13 +64,15 @@ impl Tool for Composer {
...
-        .collect_vec()
+        match std::env::var("PATH") {
+            Ok(path) => split_paths(&path)
+                .map(path_to_native_string)
+                .collect_vec(),
+            Err(_) => {
+                debug!("PATH not found for composer, using empty path");
+                Vec::new()
+            }
+        }
```
brynary, line 75: After updating the return type for `extra_env_paths` we can update this code to return an `Err` instead of logging and returning an empty Vec

qlty-check/src/tool/php/composer.rs
```diff
@@ -175,7 +180,13 @@ impl Composer {
             // use the original composer.json contents, merging package_file contents on top.
             // this will retain any existing dependencies provided by the initial tool installation
             let contents = std::fs::read_to_string(&staged_file)?;
-            data_json = serde_json::from_str::<Value>(&contents).unwrap_or_default();
+            data_json = match serde_json::from_str::<Value>(&contents) {
+                Ok(json) => json,
+                Err(err) => {
+                    debug!("Failed to parse existing composer.json: {}", err);
```
brynary, line 186: Upgrade this from a debug level to `error!` level

PR #1594, opened 2025-03-30
Fix running `qlty check PATH` in a subdir

brynary:
This patch fixes `qlty check foo/` when in a subdirectory.

As part of this fix, I simplified the concurrency controls of the `workspace_entry_finder_builder`. Previously, it was an `Arc<Mutex<PluginWorkspaceEntryFinderBuilder>>` which I reduced to a simple `PluginWorkspaceEntryFinderBuilder` that is cloned.

### DIFF
--- a/qlty-check/src/planner/plugin_workspace_entry_finder_builder.rs
+++ b/qlty-check/src/planner/plugin_workspace_entry_finder_builder.rs
@@ -1,4 +1,4 @@
-use anyhow::{Result};
+use anyhow::Result;
 use qlty_analysis::utils::fs::path_to_string;
 use qlty_analysis::workspace_entries::{IgnoreGroupsMatcher, TargetMode};
 use qlty_analysis::{
@@ -41,9 +41,9 @@ impl PluginWorkspaceEntryFinderBuilder {
 
     pub fn diff_line_filter(&mut self) -> Result<Box<dyn IssueTransformer>> {
         match self.mode {
-            TargetMode::HeadDiff | TargetMode::UpstreamDiff(_) => {
-                Ok(Box::new(self.git_diff.as_ref().unwrap().line_filter.clone()))
-            }
+            TargetMode::HeadDiff | TargetMode::UpstreamDiff(_) => Ok(Box::new(
+                self.git_diff.as_ref().unwrap().line_filter.clone(),
+            )),
             _ => Ok(Box::new(NullIssueTransformer)),
         }
     }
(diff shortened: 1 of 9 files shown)


brynary:
@marschattha can you please take a look at the changes to the way we use `PluginWorkspaceEntryFinderBuilder` here? I remove the `Arc<Mutex<T>>` wrapping and locking code.

As I did that, I noticed that we seemed to be _modifying_ the state of the `prefix` field on a _shared_ `PluginWorkspaceEntryFinderBuilder`, which is generally not advised.

I don't have context for why that was introduced so I want to be sure I'm not missing something. Basically I'd like you to look at two things:

1. Do you see anything in the code that suggests these changes will produce inaccurate results?

2. Based on these changes, do you see anything that suggests a performance regression?

My testing is OK, so this is just an extra review

lsegal:
Have we benchmarked this change under load? I would expect this to have some type of impact on projects with a lot of files.

brynary:
I've been staring at this code for the same reason -- to see if there is a material performance degradation from the changes to the data sharing. I can't find one.

I compiled into release mode and do a bit of ad hoc testing locally and wasn't able to detect a timing slowdown. However, with linters in play, it's hard to tell.

As I type this out, I am realizing that I think any slowdown would only affect plan stage. We had an N^2 issue in the planner at one point and it was massively slow, which we found and fixed. I did not see anything obvious like that with this PR. Fortunately, we print the planner latency so if it's slow (> 500ms under normal circumstances) we should see it clearly.

Best I can tell, the key thing we need to do is walk the targets once and then hand that data to each plugin planner. This PR still does that. It copies more memory but we have not yet reached the level of latency strictness where we are worrying about memory allocations. 

All that said, I could easily be missing something. Do you see a spot in the code where it looks like a slowdown, besides extra memory cloning?

qlty-check/src/planner/plugin_workspace_entry_finder_builder.rs
```diff
@@ -20,61 +20,47 @@ pub struct PluginWorkspaceEntryFinderBuilder {
     pub paths: Vec<PathBuf>,
     pub file_types: HashMap<String, FileType>,
     pub ignores: Vec<Ignore>,
-    pub cached_git_diff: Option<GitDiff>,
+    pub git_diff: Option<GitDiff>,
     pub prefix: Option<String>,
```
marschattha, line 24: I think this `prefix` attribute on `PluginWorkspaceEntryFinderBuilder` struct is no longer needed?

marschattha (approved):
One little comment and we should add a test around the bug fix? 

Otherwise don't see any other issues.

PR #1626, opened 2025-04-03
Add aarch 64 linux musl target for cargo dist

marschattha:


brynary:
@marschattha is your understanding that we have never supplied Linux + ARM + MUSL binaries?

I remember a conversation with @lsegal about Linux + ARM from awhile back... Maybe we just never got to MUSL. @lsegal do you happen to remember?

marschattha:
We have linux Arm 64 with gnu support but not with musl as far as I can tell.

lsegal:
There may or may not be compatibility issues, but it should be easy to verify by running the necessary builders / tests. If they run, we're fine.

PR #1658, opened 2025-04-05
Copy lockfiles when no package_filters

marschattha:
A common issue I have noticed in testing this is that if the version of the linter in the lockfile is specified to be different than in qlty.toml (often the default known_good_version) we get the version mismatch error during version check post tool install. 

Maybe we should skip version check in this case?

### DIFF
--- a/qlty-check/src/tool/node/package_json.rs
+++ b/qlty-check/src/tool/node/package_json.rs
@@ -66,6 +66,27 @@ impl PackageJson {
         let final_package_file = serde_json::to_string_pretty(&user_json)?;
         debug!("Writing {} package.json: {}", tool_name, final_package_file);
 
+        if self.plugin.package_filters.is_empty() {
+            if let Some(package_file) = &self.plugin.package_file {
+                let package_file_path = PathBuf::from(package_file);
+                if let Some(parent_path) = package_file_path.parent() {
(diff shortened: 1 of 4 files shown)


brynary:
@marschattha on the version check, what do you think of this?:

- IF a `version` is specified in the qlty.toml we check for it
- ELSE we do not run the version check

marschattha:
> @marschattha on the version check, what do you think of this?:
> 
> * IF a `version` is specified in the qlty.toml we check for it
> * ELSE we do not run the version check

Thinking and doing the prototype of that, might not be a good idea as it would stop checking for a lot more cases than it needs to as it is find to do that when the version is latest or known_good, just not ideal when installing using lockfile. 
I think the version check provides a good constraint and should prevent bugs.
Maybe limiting it to cases where we copy lockfiles might be enough?

brynary:
Hmm, I think the issue here is that the version check does two different, important things:

1. Make sure the linter is installed properly
2. Validate the version is expected

We can get the benefit of number 1 by running the version command and simply not validating the output.

So, what if we break down the concept of a version check into a validated version check which checks the output vs. an invalidated version check which just makes sure it exits 0.

Would that help?

marschattha:
Ah we were already skipping version check for ruby and node packages when using package files, was just missing it in php.

brynary (approved):

PR #1684, opened 2025-04-07
Default to known good version instead of latest

marschattha:
We are now automatically updating latest version for most plugins but they might not always be compatible (may have breaking changes), so we should default to known_good instead of latest I believe


brynary:
@marschattha yes, this seems like the right way to go. Do we have any gaps as in plugins where we don't have automated updates?

marschattha:
Yes there are a few, Radarlint plugins, Rust plugins (since they come packaged with rust) and Go plugins.

List of covered: Github, Ruby, Python, Php, Node and Runnable Archive/Java.

brynary (approved):

PR #1773, opened 2025-04-07
Test against lastest snapshot

marschattha:
Now that we may not have a known good version snapshot we should test with the highest version snapshot available for linter version updates

### DIFF
--- a/qlty-plugins/plugins/tests/utils.ts
+++ b/qlty-plugins/plugins/tests/utils.ts
@@ -11,11 +11,8 @@ export interface EnvOptions {
   /** Prevents the deletion of sandbox test dirs. */
   sandboxDebug: boolean;
 
-  /** Test the output against the known good version output of the linter. */
-  testAgainstKnownGoodVersion: boolean;
-
   /** Test the output against the latest snapshots of the linter. */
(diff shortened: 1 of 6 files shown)


qlty-plugins/plugins/tests/utils.ts
```diff
@@ -13,6 +13,9 @@ export interface EnvOptions {
 
   /** Test the output against the known good version output of the linter. */
   testAgainstKnownGoodVersion: boolean;
+
+  /** Test the output against the latest snapshots of the linter. */
+  testAgainstLatestSnapshot: boolean;
```
brynary, line 18: The input flags to the testing can get a bit confusing with the various versions. What do you think about adjusting the naming so we have like: `runKnownGoodVersions: boolean` and `compareLatestSnapshots: boolean`. 

So `run*` flags would control what linter version we execute and `compare*` refers to the snapshots

marschattha: We can actually remove testAgainstKnownGoodVersion since it is not being used now that we have compareLatestSnapshots

brynary (approved):

PR #1819, opened 2025-04-08
fix: update SARIF formatter in RFC3339 format

TartanLeGrand:
Close this :

https://github.com/qltysh/qlty/issues/1485

This pull request includes updates to the `qlty-cli/src/format/sarif.rs` file to enhance the SARIF formatter with dynamic release date handling and the addition of a new import.

Enhancements to SARIF formatter:

* [`qlty-cli/src/format/sarif.rs`](diffhunk://#diff-d124dbb2d858adb3d6504326903340cf51ac36feadbacabfdf11dc7095f0d02dR9): Added `chrono::Utc` to the imports for handling date and time operations.
* [`qlty-cli/src/format/sarif.rs`](diffhunk://#diff-d124dbb2d858adb3d6504326903340cf51ac36feadbacabfdf11dc7095f0d02dR301-R306): Introduced logic to dynamically set the `release_date` based on whether `BUILD_DATE` contains a time component, defaulting to the current UTC time in RFC 3339 format if not.
* [`qlty-cli/src/format/sarif.rs`](diffhunk://#diff-d124dbb2d858adb3d6504326903340cf51ac36feadbacabfdf11dc7095f0d02dL311-R318): Updated the SARIF output to use the dynamically determined `release_date` instead of the static `BUILD_DATE`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ flamegraph.svg
 /.markdownlint.json
 /.rustfmt.toml
 /.yamllint.yaml
+.idea
+.vscode
--- a/qlty-config/build.rs
+++ b/qlty-config/build.rs
@@ -48,6 +48,6 @@ fn main() {
         std::env::var("PROFILE").unwrap()
     );
 
-    let build_date = chrono::Utc::now().format("%Y-%m-%d").to_string();
+    let build_date = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string();
     println!("cargo:rustc-env=BUILD_DATE={}", build_date);
 }


brynary:
Hi @TartanLeGrand -- Thanks for this PR! I'm trying to understand the SARIF spec a bit better as it relates to this.

Here is what I've found so far in the spec:

1. The [`releaseDateUtc`[(https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10540986) property says "whose value is a string in the format specified in §3.9"
2. [Section 3.9](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10540986) says that the string but be compatible with ISO8601 and lists as an example `2016-02-08`.

In our code `BUILD_DATE` is generated by [`qlty-config/build.rs`](https://github.com/qltysh/qlty/blob/main/qlty-config/build.rs#L51) in the format `%Y-%m-%d` in the UTC timezone. 

Reading these together, the it seems like the `releaseDateUtc` values we generate are compatible with the spec because they are ISO8601 `YYYY-MM-DD` dates in UTC format.

The change in this PR looks like it would cause the `releaseDateUtc` to be computed at _runtime_ instead of at release time when we compile the package.

What do you think?

TartanLeGrand:
Hello @brynary,

I believe the UTC format you provided is fine 👍. Unfortunately, some tools don't properly handle it (so it's more of an issue on their end). That said, I figured the best approach is to do what we can and adjust things a bit. I've moved the date generation to qlty-config/build.rs. Does that work for you? It's not a huge change, right? 🚀

I also modified the format to include the time, adhering to the ISO8601 standard as specified in section 3.9 (e.g., 2016-02-08T16:08:25Z). Let me know what you think! 😊

brynary:
Hi @TartanLeGrand -- OK, got it. Yes, I think they need to adjust their parsing.

I'd like to stick with the `YYYY-MM-DD` format in the Qlty CLI because it is the simplest date format that is compatible with ISO8601 and the SARIF spec. My hope is that the SARIF processors can be updated soon.

I'm looking forward to looking into the GitLab component more closely soon!

TartanLeGrand:
I've find a temp work around : https://gitlab.com/nova-edge/components/qlty/-/blob/main/templates/code_quality.yml?ref_type=heads#L32

brynary:
OK, I think that's the way to go until they fix the reading of that field. I'm going to close this one for now.

PR #1850, opened 2025-04-08
Fixes for issues created by update linter version workflow.

marschattha:
1. Skip tsc since it doesn't have a test.
2. Add support for golangci-lint v2.
3. Normalize version numbers in documentation urls in test snapshots, simplifies upgrades for plugins such as tflint and bandit which have different version numbers in urls.

### DIFF
--- a/qlty-plugins/plugins/linters/golangci-lint/plugin.toml
+++ b/qlty-plugins/plugins/linters/golangci-lint/plugin.toml
@@ -22,12 +22,24 @@ description = "A powerful Go linter runner"
 security = true
 suggested_mode = "comment"
 
-[plugins.definitions.golangci-lint.drivers.lint]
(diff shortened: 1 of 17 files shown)


qlty-plugins/plugins/linters/golangci-lint/plugin.toml
```diff
@@ -32,6 +33,17 @@ output_format = "golangci_lint"
 suggested = "targets"
 output_missing = "parse"
 
+[[plugins.definitions.golangci-lint.drivers.lint.version]]
+version_matcher = ">=2.0.0"
```
brynary, line 37: The version matcher above has `<=` and this one has `>=` so they overlap on `2.0.0`

brynary (approved):

PR #1882, opened 2025-04-09
Skip dockerfmt tests on windows

marschattha:
Skip dockerfmt tests on windows as it does not support windows


brynary (approved):
Approving, but how did this get through CI for the release?

PR #1884, opened 2025-04-09
Update Sarif parser to support missing uri in originalUriBaseIds

marschattha:


brynary (commented):
@marschattha have you compared this change to the official SARIF spec to make sure that what Brakeman is outputting is valid?

marschattha:
It seems to be valid according to https://sarifweb.azurewebsites.net/Validation

brynary (approved):

PR #1913, opened 2025-04-11
Add support for --total-parts-count=NUM to `qlty coverage publish` command

noahd1:
Enables taking advantage of server side coverage aggregation by informing the server how many parts to expect.


qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -83,13 +83,28 @@ pub struct Publish {
 
     #[arg(long, hide = true)]
     pub skip_missing_files: bool,
+
+    #[arg(long)]
+    /// The total number of parts that qlty cloud should expect. Each call to qlty publish will upload one part.
+    /// (The total parts count is per coverage tag e.g. if you have 2 tags each with 3 parts, you should set this to 3)
+    pub total_parts_count: Option<u32>,
```
brynary, line 90: This is a verbose description for the arg. How does this look in the `--help` output?

qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -83,13 +83,28 @@ pub struct Publish {
...
     pub fn execute(&self, _args: &crate::Arguments) -> Result<CommandSuccess, CommandError> {
         self.print_initial_messages();
 
+        if let Some(total_parts) = self.total_parts_count {
+            if total_parts == 0 {
+                eprintln!(
+                    "{}",
+                    style("Error: total_parts_count must be greater than 0.").red()
+                );
+                std::process::exit(1);
```
brynary, line 104: Commands shouldn't be `process:exit()`-ing, on error because it breaks flow control. Instead, they should be returning a `CommandError`

qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -83,13 +83,28 @@ pub struct Publish {
...
+    /// (The total parts count is per coverage tag e.g. if you have 2 tags each with 3 parts, you should set this to 3)
+    pub total_parts_count: Option<u32>,
 }
 
 impl Publish {
     // TODO: Use CommandSuccess and CommandError, which is not straight forward since those types aren't available here.
     pub fn execute(&self, _args: &crate::Arguments) -> Result<CommandSuccess, CommandError> {
         self.print_initial_messages();
 
+        if let Some(total_parts) = self.total_parts_count {
```
brynary, line 98: I'd extract this to some kind of validate function to keep it out of the main flow

qlty-cli/src/commands/coverage/publish.rs
noahd1: Lots of output here already ...

image

brynary: Hmm OK. Let's go with this for now and I'll do an editing pass someday.

Basically we need man pages

brynary (approved):

PR #1915, opened 2025-04-12
Add `--validate` option to `qlty coverage publish`

marschattha:
Add a new `qlty coverage validate ...` subcommand that takes raw coverage files as input and validates if the files in the report match the files on the file system, if more than a certain threshold of files in the coverage report are not present on the filesystem the command errors out.


brynary:
@marschattha nice!

So I was thinking about this a bit and I was wondering... maybe the `coverage validate` subcommand should accept a Zip file as input.

That would ensure that it's validating the exact archive that would be sent up to the server. The strip prefix instructions etc. would be limited to the Zip file generation phase.

So the full process with validation would be:

1. Generate the Zip file
2. Validate Zip file
3. Publish -- Obtain signed upload URL and upload the Zip file

We might need to re-arrange some things to separate the zip file generation from the publish. E.g. a `coverage package` subcommand, but it seems cleaner over the long term. 

What do you think?

marschattha:
Well if you pass `--dry-run` flag to `qlty coverage publish` it does what one would expect package to do, just write files to disc. Also don't really need to read from the zip file as there is a final [file_coverages.jsonl](https://github.com/qltysh/qlty/blob/ma/coverage_validation_command/qlty-coverage/src/export.rs#L58) file created which is then [placed into zip](https://github.com/qltysh/qlty/blob/ma/coverage_validation_command/qlty-coverage/src/export.rs#L80) so we can just pass that into `qlty coverage validate` command and not need the zip reading.

brynary:
Oh, I forgot about `--dry-run`. That's exactly what we need!

However, I still think we want to the `validate` subcommand to target the Zip file instead of a file_coveeages JSONL file. There are a couple of reasons...

One is that over time I see the `validate` command validating more than just file coverages. It could validate any data in the Zip.

The second is ergonomics. The command really should be `qlty coverage validate PATH_TO_ZIP` because that allows you to validate any Zip file.

As an example, in theory if we are debugging coverage for a user we should be able to validate their Zip file without needing to unpack it.

qlty-cli/src/commands/coverage/validate.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl Validate {
+    pub fn execute(&self, _args: &crate::Arguments) -> Result<CommandSuccess, CommandError> {
+        let validation_result = ValidationResult::compute(&self.path, self.threshold)?;
+
+        if self.json {
+            println!("{}", serde_json::to_string_pretty(&validation_result)?);
+        }
+
+        match validation_result.status {
```
brynary, line 30: We have this concept of transforming a validation result into a command result in a couple places

I think we can refactor this into a trait implementation of the `From` trait

qlty-cli/src/commands/coverage/validate.rs
```diff
@@ -0,0 +1,43 @@
...
+use qlty_coverage::validate::{ValidationResult, ValidationStatus};
+use serde_json;
+
+#[derive(Debug, Args)]
+pub struct Validate {
+    #[arg(long)]
+    /// The minimum percentage of coverage report files to match the file system for validation to succeed.
+    /// If not specified, defaults to 90%.
+    pub threshold: Option<f64>,
+
```
brynary, line 13: We should probably qualify this to be more specific. Like `--paths-threshold`

qlty-cli/src/commands/coverage/validate.rs
marschattha: That would be something like https://github.com/qltysh/qlty/pull/1915/commits/e13c25454e293156c99b1e21a5902c59a2ed85c1?

But https://github.com/qltysh/qlty/pull/1915/commits/9b6261256ba7facd895975889e66db21059f273a might be simpler.

PR #1917, opened 2025-04-13
Replace single "cli" coverage tag with tag per OS

noahd1:


brynary:
@noahd1 nice and simple. Just realized that we are missing the tag from the CLI output also. We should add that.

PR #1937, opened 2025-04-15
Add plugin specific path ignores for fmt

marschattha:
https://github.com/qltysh/qlty/issues/1565


brynary:
@marschattha I think the current `[[ignore]]` handling is just too complex. It's had too many bugs, and I want to simplify it.

Here is my proposal: https://github.com/qltysh/cloud/issues/5925

I would like to close this bug fix PR and pivot to finalizing and implementing that proposal in a way that simplifies the code and makes it less bug prone in the code and less error prone for the user writing the config.

PR #1938, opened 2025-04-16
Fix missing smells when ignore rules are used in qlty.toml

lsegal:
The following qlty.toml config will cause IgnoreGroup to register a blanket wildcard (`*`) ignore:

```
[[ignore]]
rules = ["tool:rule_name"]
```

Due to the way the qlty.toml is loaded, the `file_patterns` for this ignore definition will be set to `*`, thereby creating an ignore group that looks like:

```
["**/protos/**", "**/other_ignore/**", "*", "**/tailwind.css", ...]
```

Which will match (and ignore) all files.

The proposed fix is to skip an ignore wildcard if any rules are defined, as this should not be treated as a blanket workspace_entry ignore (an invocation must be run to determine what needs to be ignored).

There is a possible optimization for `tool:*` rules for the plugin workspace entry matcher, but I did not want to complicate the implementation.

### DIFF
--- a/qlty-analysis/src/workspace_entries/matchers/languages_shebang_matcher.rs
+++ b/qlty-analysis/src/workspace_entries/matchers/languages_shebang_matcher.rs
@@ -2,7 +2,7 @@ use super::WorkspaceEntryMatcher;
 use crate::{code::language_detector::get_language_from_shebang, WorkspaceEntry};
 use std::collections::HashMap;
 
-/// Matches workspace entries that have a sheband line that includes an interpretter
+/// Matches workspace entries that have a sheband line that includes an interpreter
 /// This requires reading the contents of the file.
 #[derive(Debug)]
 pub struct LanguagesShebangMatcher {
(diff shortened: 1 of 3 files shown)


brynary:
Good find. The ignore syntax is problematic and I think we may need to address this at the root cause.

This is probably the third bug where we've had unpredictable consequences of various ignore syntax in different context. There's another PR open for a similar problem with ignored applying to `fmt`

The issues are fixable but the complexity of the code has grown too high already and each fix requires more contortions. 

I am pretty much at the point of declaring bankruptcy on the `[[ignore]]` syntax entirely and working out a new syntax. We can support the old one as legacy for awhile, with a warning.

I'll work on use cases and a proposed config syntax this week.

brynary:
I just read the [other card](https://github.com/qltysh/support/issues/54) and realized that this is a regression. This seems like a serious regression so I'm going to merge and release this now and we can handle a simplificaiton to make this code less likely to attract bugs in the future.

PR #1947, opened 2025-04-17
Add coverage_tool and generation_command to CoverageMetadata

marschattha:


qlty-coverage/src/ci.rs
```diff
@@ -13,6 +13,8 @@ pub use gitlab::GitLab;
 use qlty_types::tests::v1::CoverageMetadata;
 pub use semaphore::Semaphore;
 
+const QLTY_CI_ACTION_VERSION: &str = "QLTY_CI_ACTION_VERSION";
```
brynary, line 16: This should be UPLOADER_VERSION to align with the names we are using in the protobufs

qlty-coverage/src/ci.rs
```diff
@@ -45,13 +47,29 @@ pub trait CI {
     fn build_url(&self) -> String;
 
     fn metadata(&self) -> CoverageMetadata {
+        let ci_action_version = std::env::var(QLTY_CI_ACTION_VERSION).ok();
+        let mut uploader_tool = None;
+        let mut uploader_tool_version = None;
+
+        if let Some(version) = ci_action_version {
+            let mut parts = version.split('@');
```
brynary, line 55: Rather than doing parsing/splitting of this value in the CLI, for simplicity, could we arrange it so the CLI can just pass through the strings without needing to inspect them?

Basically similar to a "user-agent" string which is mostly opaque

brynary (approved):
Approving with a couple comments

PR #1950, opened 2025-04-17
Add validation of coverage upload plan and fatal errors

brynary:
Screenshot 2025-04-17 at 2 00 38 pm


qlty-cli/tests/cmd/check/prepare_script_fail.stderr
```diff
@@ -1,2 +1,2 @@
-❌ [..]: prepare_script exited with non-zero code [..]:[..]INVALID_COMMAND[..]
```
marschattha, line 1: Not sure if this is intended?
Kind of makes this test pointless.

PR #1954, opened 2025-04-18
Add excludes to replace ignores

marschattha:
Refactor current `ignores` concept into `excludes` 
Part of https://github.com/qltysh/cloud/issues/5925


.qlty/qlty.toml
```diff
@@ -91,10 +91,11 @@ name = "yamllint"
...
 # on our test fixtures directories
-[[ignore]]
+[[exclude]]
 plugins = ["osv-scanner", "trufflehog"]
 file_patterns = ["qlty-cli/tests/**"]
 
-[[ignore]]
-rules = ["markdownlint:MD024"]
-file_patterns = ["CHANGELOG.md"]
+# Temporarily comment this.
```
brynary, line 98: Is this intended?

qlty-analysis/src/workspace_entries/matchers/plugin_specific_exclude_matcher.rs
```diff
@@ -0,0 +1,215 @@
+use crate::{utils::fs::path_to_string, WorkspaceEntry};
+use qlty_config::config::Exclude;
+use std::path::PathBuf;
+
+use super::WorkspaceEntryMatcher;
+
+#[derive(Debug)]
+pub struct PluginSpecificExcludeMatcher {
+    plugin_name: String,
+    excludes: Vec<Exclude>,
```
brynary, line 10: My understanding is that the `Exclude` structs have plugin information on them already, so I wasn't expecting to see them alongside a `plugin` field like this.

This `PluginSpecificExcludeMatcher` seems a bit... specific :) I am wondering if there might be a more elegant way to compose this behavior. Ideally the code would feel like a clean composition. 

What do you think?

qlty-check/src/planner/plugin_workspace_entry_finder_builder.rs
```diff
@@ -91,15 +94,33 @@ impl PluginWorkspaceEntryFinderBuilder {
...
+                .collect::<Vec<_>>();
+
+            if !plugin_specific_excludes.is_empty() {
+                matchers.push(Box::new(PluginSpecificExcludeMatcher::new(
+                    plugin_name.clone(),
+                    plugin_specific_excludes,
+                    self.root.clone(),
+                )));
+            }
+        }
```
brynary, line 122: This section of code reads a bit unexpected and somewhat confusing to me.

Is this some form of optimization?

qlty-cli/src/initializer/scanner.rs
```diff
@@ -353,11 +353,13 @@ impl Scanner {
         }
     }
 
-    fn compute_default_ignores(&mut self) {
-        self.default_ignores = self.default_config.ignore.clone();
-        self.default_ignores
+    // Not sure why we have this?
+    // Maybe it should be removed.
```
brynary, line 357: We should address this comment before merging.

.qlty/qlty.toml
marschattha: Yes rules are no longer supported by the new exclude [syntax](https://github.com/qltysh/cloud/issues/5925) and this should be moved to [triage](https://github.com/qltysh/qlty/pull/2005) once that lands.

qlty-analysis/src/workspace_entries/matchers/plugin_specific_exclude_matcher.rs
marschattha: So tying this with the other [comment](https://github.com/qltysh/qlty/pull/1954/files#r2086759578). 

From my understanding you can have two types of [excludes](https://github.com/qltysh/cloud/issues/5925). One with only `file_patterns` (will match regardless of plugin) and then targeting `file_patterns` for given set of plugins.

This specific matcher is for the second case, if plugin(s) are specified. Otherwise the generic `ExcludeGroupsMatcher` (formerly `IgnoreGroupsMatcher`) is used.

If memory serves correct previously we did not have anything that actually worked to filter the plugins even though the syntax was there which caused rebuilding/breaking down of ignores and overrides into excludes and triage.

qlty-config/src/config/builder.rs
```diff
@@ -160,11 +159,40 @@ impl Builder {
...
+
+            for ignore in &config.ignore {
+                if ignore.file_patterns.is_empty() {
+                    eprintln!(
+                        "{} The use of `{}` field in qlty.toml without {} is no longer supported. Skipping ignore without file_pattern.",
+                        style("WARNING:").bold().yellow(),
+                        style("ignore").bold(),
+                        style("file_patterns").bold()
+                    );
+                    continue;
```
brynary, line 179: @marschattha can you please remind me what an `[[ignore]]` without `file_patterns` can be used for today in the current CLI? Can they do anything useful?

qlty-check/src/planner/config_files.rs
```diff
@@ -93,15 +90,17 @@ pub fn plugin_configs(planner: &Planner) -> Result<HashMap<String, Vec<PluginCon
...
-    let ignore_globset = ignore_globset(&planner.config.ignore)?;
+    let exclude_globset = exclude_globset(&planner.config.exclude_patterns)?;
 
     for entry in planner.workspace.walker() {
         let entry = entry?;
         if let Some(os_str) = entry.path().file_name() {
             let file_name = os_str.to_os_string();
             for plugin_config in &plugins_configs {
+                // Why do we have exclude_globset.is_match(entry.path()) here?!
+                // If you want to exclude the file, you should not add it to the configs.
```
brynary, line 101: We should address this comment before merging this

qlty-config/src/config/exclude.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::sync::RwLock;
+
+fn ensure_non_empty_vec<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
+where
+    D: Deserializer<'de>,
+    T: Deserialize<'de>,
+{
+    let vec = Vec::<T>::deserialize(deserializer)?;
+    if vec.is_empty() {
+        Err(serde::de::Error::custom("Vector cannot be empty"))
```
brynary, line 15: If this error triggers, will it indicate the field name?

marschattha: Example: 
image

qlty-check/src/planner/config_files.rs
marschattha: I was unsure about the reasoning/actual case for this one so added a comment to bring it up.
But I suppose if it was added there might of had been a case with a user where they want to ignore configs from a certain directory.

qlty-config/src/config/builder.rs
marschattha: But we should have this functionality back one we finish up triage work.

qlty-check/src/planner/config_files.rs
marschattha: Actually there is a test case which covers this as well, so probably needed.

qlty-config/src/config/exclude.rs
brynary: Hmm, I think this screenshot is a different case right? (missing `plugins` field is caught by Serde, but this special error is for `[]`)

qlty-config/src/config/builder.rs
brynary: Ah, so these would become `[[triage]]` blocks in the new system, right?

qlty-cli/src/initializer/scanner.rs
brynary: Seems right, I think we can safely remove it but we should probably do that in a tiny, isolated change later

brynary:
@marschattha OK, I think this PR is in good shape!

My only concern is that if we merge this it seems like we will break the `file_pattern`-less. That could cause customer issues so I only see two real possibilities:

1. Figure out how to get enough confidence that customers aren't using that feature. However, as I type this -- we are using it, and some customers are likely using it to

2. The other idea would be to merge the `[[triage]]` feature _before_ this one and then update this one to create triage rules.

What do you think?

brynary:
Yeah, agreed. I just reviewed #2005 and I think it's very close. So let's plan to get #2005 merged first and then circle back to this one.

qlty-config/src/config/triage.rs
```diff
@@ -19,16 +20,56 @@ pub struct Match {
...
     pub file_patterns: Vec<String>,
 
+    #[serde(default)]
+    pub levels: Vec<Level>,
+
     #[serde(skip)]
     pub glob_set: RwLock<Option<GlobSet>>,
 }
 
+impl<'de> Deserialize<'de> for Match {
```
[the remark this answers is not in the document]
marschattha: So we can use the Level Enum for levels attribute instead of String and have it parse human readable "low" instead of "LEVEL_LOW".

qlty-config/src/config/triage.rs
```diff
@@ -19,16 +20,56 @@ pub struct Match {
     #[serde(default)]
     pub file_patterns: Vec<String>,
 
+    #[serde(default)]
+    pub levels: Vec<Level>,
+
```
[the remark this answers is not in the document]
marschattha: Yup we can safely remove the `#[serde(default)]` bit here above the levels since that is always going to have an array from the custom deserializer.

brynary (approved):

PR #1965, opened 2025-04-24
Update coverage publish args

marschattha:
https://github.com/qltysh/cloud/issues/5932

### DIFF
--- a/qlty-cli/src/commands/coverage/publish.rs
+++ b/qlty-cli/src/commands/coverage/publish.rs
@@ -197,12 +197,6 @@ impl Publish {
                 "WARNING: --transform-strip-prefix is deprecated, use --strip-prefix instead\n"
             );
         }
-        if self.token.is_some() {
-            eprintln_unless!(
-                self.quiet,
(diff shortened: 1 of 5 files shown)


qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -228,19 +302,17 @@ impl Publish {
         if let Some(override_pr_number) = &self.override_pr_number {
             eprintln_unless!(self.quiet, "    override-pr-number: {}", override_pr_number);
         }
-        if let Some(transform_add_prefix) = &self.transform_add_prefix {
-            eprintln_unless!(
-                self.quiet,
-                "    transform-add-prefix: {}",
-                transform_add_prefix
-            );
+        if let Some(add_prefix) = &Self::prefer_active_over_deprecated_option(
```
brynary, line 300: We have a bit of a code construction problem where we are repeating this here on line 305 what we already did on line 137. We should refactor so that we only are doing the coalesce once

PR #1986, opened 2025-04-29
Add swiftformat plugin

marschattha:


brynary:
On this one, I have the same question about GitHub-release based downloads vs. URL-based downloads from #1987

qlty-plugins/plugins/linters/swiftformat/plugin.toml
```diff
@@ -0,0 +1,35 @@
+config_version = "0"
+
+[plugins.downloads.swiftformat]
```
brynary, line 3: @marschattha we need to update this one too like #1987 right?

marschattha: Yup wanted to do this afterwards, since merging that one introduced a merge conflict.

brynary (approved):

PR #1987, opened 2025-04-29
Add stringslint plugin

marschattha:


qlty-plugins/plugins/tests/runLinterTest.ts
```diff
@@ -12,7 +12,8 @@ Debug.inspectOpts!.hideDate = true;
 
 // Currently unsupported tools on Windows
 const SKIP_LINTERS = {
-  win32: ["semgrep", "swiftlint", "dockerfmt"],
+  win32: ["semgrep", "swiftlint", "dockerfmt", "stringslint"],
+  linux: ["stringslint", "swiftlint"],
```
brynary, line 16: Not for this PR, but I wonder if we can infer SKIP_LINTERS behavior from the `supported_platforms` data someday

qlty-plugins/plugins/linters/stringslint/plugin.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[[plugins.downloads.stringslint.system]]
+cpu = "x86_64"
+os = "macos"
+url = "https://github.com/dral3x/StringsLint/releases/download/${version}/portable_stringslint.zip"
+
+[[plugins.downloads.stringslint.system]]
+cpu = "aarch64"
+os = "macos"
+url = "https://github.com/dral3x/StringsLint/releases/download/${version}/portable_stringslint.zip"
```
brynary, line 16: @marschattha I might be mis-remembering but I thought that the `github = ...` syntax was an _alternative_ to specifying per-cpu/os download URLs. Don't we typically use one or the other?

For example, we have `dral3x/StringsLint` in both the `github =` line and in the URLs... It seems like this creates a risk of them being out of sync.

marschattha: Since the Github release syntax tries to match platform and architecture in the assets list, generic name like `portable_stringslint.zip` doesn't get matched and it fails.

brynary: @marschattha right, that makes sense, but what I'm saying is that do we need the `[plugins.downloads.stringslint]` block at all? It seems redundant.

I think we have two, mutually exclusive ways to specify downloads:

1. Automatic detection of download URLs GitHub releases; OR
2. Download URLs (which may be on GitHub, but the system just understands them as a URL)

So I think the issue with the generic name `portable_stringslint.zip` means we would avoid the GitHub release approach entirely.

marschattha: Actually the downloads/DownloadDef doesn't even have github, that is the `release`

brynary: Ah ok. Maybe we need warnings for invalid config vars, or a "strict" mode we can use in development. I've been caught by that also

PR #1988, opened 2025-04-29
Prevent .qlty/.gitignore from ignoring itself

marschattha:
Add `!.gitignore` to .gitignore template so it gets tracked and prevent vcs issues.

### DIFF
--- a/.qlty/.gitignore
+++ b/.qlty/.gitignore
@@ -1,4 +1,7 @@
 *
 !configs
+!configs/**
 !hooks
+!hooks/**
 !qlty.toml
+!.gitignore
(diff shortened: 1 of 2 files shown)


.qlty/.gitignore
```diff
@@ -1,4 +1,7 @@
 *
 !configs
+!configs/**
```
lsegal, line 3: Is the extra wildcard necessary? Technically everything within the configs dir should already be ignored because of the above line.

marschattha: Mostly pasted what was in https://github.com/qltysh/qlty/blob/main/qlty-cli/src/initializer/templates/gitignore.txt.

But according to bot:

Pattern | Meaning
-- | --
!configs | Unignore the configs folder itself — i.e., the folder will exist, but files inside are still ignored unless otherwise specified.
!configs/** | Unignore everything inside configs recursively — all files and subfolders will be unignored.

Although not sure if that always makes a difference.

PR #1996, opened 2025-04-30
Fix init --source flag rendering

marschattha:
https://github.com/qltysh/cloud/issues/6027


qlty-cli/src/initializer/templates/source_default.toml
```diff
@@ -1,3 +1,3 @@
 [[source]]
 name = "default"
-default = true
\ No newline at end of file
+default = true
```
brynary, line 3: The error was the missing newline at end of this file, is that right?

marschattha: Yup, it was rendering `default = true[[source]]`

brynary (approved):

PR #1997, opened 2025-04-30
Copy plugin configs from source during qlty plugins enable

marschattha:
https://github.com/qltysh/cloud/issues/6028

I think the some duplication is fine, trying to reduce it added more complexity. 

Also noticed for plugins with multiple config files, if any of those config files are present we shouldn't be copying over config file from source.


brynary (commented):
Can we please break this down into two PRs... One to change the behavior when we found a config to not copy other configs, and then a second PR for the `plugins enable` behavior?

I've been trying hard to keep up the CHANGELOG and it helps a lot to be able to derive it from the PR list with one changelog per PR

marschattha:
Updated this PR to just copy configs during qlty plugins enable. 
Opening separate PR for changing existing scanner behaviour.

brynary (approved):

PR #2000, opened 2025-04-30
Add support for source.toml in source root

marschattha:
https://github.com/qltysh/cloud/issues/6029

I wonder if there is a better way to test this though.


brynary (commented):
@marschattha for the testing, you can use `assert_yaml_snapshot` which I believe should match based on YAML-equality not string-equality and therefore be immune to key sorting issues

brynary (approved):

PR #2003, opened 2025-05-01
Fix coverage test snapshot

marschattha:
https://github.com/qltysh/qlty/pull/1999 introduced a change which broke this snapshot test, updating the snapshot.


noahd1:
@marschattha thank you!

PR #2005, opened 2025-05-01
Deprecate override in favour of triage

marschattha:
https://github.com/qltysh/cloud/issues/5925


qlty-config/src/config/triage.rs
```diff
@@ -0,0 +1,134 @@
...
+#[derive(Debug, Serialize, Deserialize, Default, Clone, JsonSchema)]
+pub struct Set {
+    #[serde(default)]
+    pub level: Option<String>,
+
+    #[serde(default)]
+    pub category: Option<String>,
+
+    #[serde(default)]
+    pub mode: Option<IssueMode>,
```
brynary, line 46: I think `disabled` is allowed as an `IssueMode` but should be rejected here. So I guess what we currently call "IssueMode" is really a _plugin_ mode and an Issue Mode is a plugin mode _except_ "disabled"

qlty-config/src/config/triage.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Default, Clone, JsonSchema)]
+pub struct Set {
+    #[serde(default)]
+    pub level: Option<String>,
+
+    #[serde(default)]
+    pub category: Option<String>,
```
brynary, line 43: Can this be an enum instead a string?

qlty-config/src/config/triage.rs
```diff
@@ -0,0 +1,134 @@
...
+            file_patterns: self.file_patterns.clone(),
+            glob_set: RwLock::new(None),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Default, Clone, JsonSchema)]
+pub struct Set {
+    #[serde(default)]
+    pub level: Option<String>,
```
brynary, line 40: Can this be an enum instead a string?

qlty-config/src/config/triage.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    #[serde(default)]
+    pub ignored: bool,
+}
+
+#[derive(Debug, Serialize, Deserialize, Default, Clone, JsonSchema)]
+pub struct Triage {
+    #[serde(default)]
+    #[serde(rename = "match")]
+    pub _match: Match,
```
brynary, line 56: I think we can do `r#match` as an alternative to `_match`

qlty-check/src/planner.rs
```diff
@@ -268,12 +271,53 @@ impl Planner {
...
+                        level: issue_override.level.as_ref().map(|l| level_from_str(l)),
+                        category: issue_override
+                            .category
+                            .as_ref()
+                            .map(|c| category_from_str(c)),
+                        mode: issue_override.mode.map(|m| match m {
+                            IssueMode::Block => PluginMode::Block,
+                            IssueMode::Comment => PluginMode::Comment,
+                            IssueMode::Monitor => PluginMode::Monitor,
+                            IssueMode::Disabled => PluginMode::Monitor, // Mapping Disabled to Monitor as closest equivalent
```
brynary, line 304: Instead of doing this here, I think you can implement the `From` trait to tell Rust how to transform one enum to the other, which will clean up the code

qlty-config/src/config/triage.rs
```diff
@@ -0,0 +1,165 @@
+use super::ignore::is_rule_issue_match;
+use super::plugin::PluginMode;
+use crate::config::issue_transformer::IssueTransformer;
+use globset::{Glob, GlobSet, GlobSetBuilder};
+use qlty_types::analysis::v1::{Category, Issue, Level};
+use qlty_types::{category_from_str, level_from_str};
+use schemars::JsonSchema;
```
brynary, line 7: Can you please explain the motivation for introducing this schemars dependency?

marschattha: schemars was needed to allow adding the enums Category and Level instead if String for level and category attributes of Set. Since Set is part of Triage which is part of QltyConfig which implements JsonSchema. So Category and Level and everything in between also needs to implement JsonSchema

brynary: Ah, I understand now. I had missed the fact that schemars was already in use.

qlty-types/src/lib.rs
```diff
@@ -383,3 +387,65 @@ impl AddAssign for tests::v1::CoverageSummary {
         *self = *self + other;
     }
 }
+
+impl JsonSchema for analysis::v1::Level {
```
brynary, line 391: Instead of manually implementing the `JsonSchema` trait, I think it would be better if we could have the Prost protobuf generator include a `JsonSchema` in the `derive` macro during codegen.

It looks like this is possible: https://github.com/tokio-rs/prost/issues/1064

https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.type_attribute

However, before going down this rabbit hole, what I would recommend is that you _manually edit_ the generated files to add `JsonSchema` to the existing derive statements, so:

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
#[repr(i32)]
pub enum Level {
    Unspecified = 0,
    Note = 10,
    Fmt = 20,
    Low = 30,
    Medium = 40,
    High = 50,
}
```

Becomes:

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration, JsonSchema)]
#[repr(i32)]
pub enum Level {
    Unspecified = 0,
    Note = 10,
    Fmt = 20,
    Low = 30,
    Medium = 40,
    High = 50,
}
```

If the manual edit works, then we just need to get Prost to output the code the way we want.

marschattha: I was concerned that manually modifying generated file will cause long term friction, every time the files are generated as JsonSchema would get removed and then someone would have to figure out the compile error put it back in place.

brynary:
@claude review this PR

brynary (approved):

PR #2006, opened 2025-05-01
Add new incomplete flag

noahd1:
Adds a new flag `--incomplete` which, at the moment does nothing except create invalid / incomplete coverage reports. 

However, combined with a yet to be done `qlty coverage complete` sub-command, it can eventually be used to perform server side aggregation when the total parts count is not known in advance.

### DIFF
--- a/qlty-cli/src/commands/coverage/publish.rs
+++ b/qlty-cli/src/commands/coverage/publish.rs
@@ -174,8 +174,7 @@ impl Publish {
         let add_prefix = Self::coalesce_args(&self.add_prefix, &self.transform_add_prefix);
         let strip_prefix = Self::coalesce_args(&self.strip_prefix, &self.transform_strip_prefix);
 
-        let incomplete: bool = self.incomplete
-            || (self.total_parts_count.is_some() && self.total_parts_count > Some(1));
+        let incomplete: bool = self.incomplete || self.total_parts_count.unwrap_or(1) > 1;
 
         Settings {
             override_build_id: self.override_build_id.clone(),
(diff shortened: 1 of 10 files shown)


qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -616,6 +626,12 @@ impl Publish {
                     message: String::from("Total parts count must be greater than 0"),
                 });
             }
+
+            if self.incomplete {
+                return Err(CommandError::InvalidOptions {
+                    message: String::from("The --incomplete flag is unnecessary when using --total-parts-count as it automatically marks the upload as incomplete. Please see https://qlty.sh/d/server-side-merging for more information on server-side merging."),
```
brynary, line 632: Unnecessary yes but I don't see why we'd bother the user about it

qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -303,6 +311,8 @@ impl Publish {
             eprintln_unless!(self.quiet, "    total-parts-count: {}", total_parts_count);
         }
 
+        eprintln_unless!(self.quiet, "    incomplete: {}", self.incomplete);
```
brynary, line 314: Generally we only print settings provided by the user, to balance signal vs noise. In this case, that means I would probably omit this line unless the arg is provided.

qlty-cli/src/commands/coverage/publish.rs
```diff
@@ -169,6 +174,9 @@ impl Publish {
         let add_prefix = Self::coalesce_args(&self.add_prefix, &self.transform_add_prefix);
         let strip_prefix = Self::coalesce_args(&self.strip_prefix, &self.transform_strip_prefix);
 
+        let incomplete: bool = self.incomplete
+            || (self.total_parts_count.is_some() && self.total_parts_count > Some(1));
```
brynary, line 178: This can be simplified to 

```
self.total_parts_count.unwrap_or(1) > 1
```

brynary (approved):
Approved with one small comment

PR #2007, opened 2025-05-01
linux: Add preflight checks to Ruby binary install to detect libyaml

lsegal:


lsegal:
Closing this in favor of #2009 as it provides a much better user experience that works on our supported ubuntu/debian systems (and does not break other support). Preflights may be necessary in the future, but not for this library.

PR #2008, opened 2025-05-01
Fix bubbling of stderr to tool install log files

lsegal:
Running duct::cmd::run() by default bails on non-zero exit codes, avoiding capture of any stderr. This change updates to run with `unchecked()` so that stderr can be captured into log files.

Due to this change, the specifics of Tool.run_command had to be updated to extract the original command for error display to maintain the original (somewhat helpful errors), as the Err object containing the command must now be synthesized manually (duct returned the err in the form `["program", "arg1", ...]` but this list of program/args is not available on the Expression object except via before_spawn).

### DIFF
--- a/qlty-check/src/tool.rs
+++ b/qlty-check/src/tool.rs
@@ -34,9 +34,12 @@ use regex::Regex;
 use sha2::Digest;
 use std::env::join_paths;
 use std::env::split_paths;
+use std::io::Error;
 use std::io::Write;
 use std::path::Path;
 use std::process::Output;
+use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


brynary (approved):
Thanks for breaking this up. It helped me follow the need better.

Not directly related, but someday soon we should add a field `rerun` to the invocation debug files. The `rerun` should be the _full_ command needed to reproduce the invocation inclusive of the limited and customized env vars, etc.

(Previously I tried to get this from Duct and it was not reliably available -- I found a ticket that went nowhere about it.)

This trick may be useful for that.

PR #2009, opened 2025-05-01
ruby(linux): Install libyaml as dependency in binary install

lsegal:
This avoids the need for a preflight check and allows a hands-off sandboxed approach to Ruby installs.

Obviates #2007

### DIFF
--- a/.github/workflows/cli_integration.yml
+++ b/.github/workflows/cli_integration.yml
@@ -71,18 +71,18 @@ jobs:
           ]
         include:
           - container: ubuntu:24.04
-            package_install: apt-get update && apt-get install -y git curl libyaml-dev libgmp-dev build-essential
+            package_install: apt-get update && apt-get install -y git curl build-essential
           - container: debian:trixie-slim
-            package_install: apt-get update && apt-get install -y git curl libyaml-dev libgmp-dev build-essential
+            package_install: apt-get update && apt-get install -y git curl build-essential
           #- container: debian:bookworm-slim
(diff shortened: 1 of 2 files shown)


brynary (approved):
Looks fine to merge. I wonder if there's a way we could integration test this in a way that fails before this patch but passes after this patch.

lsegal:
>  I wonder if there's a way we could integration test this in a way that fails before this patch but passes after this patch.

@brynary it's easy to reproduce on any Ubuntu/debian system that is missing `libyaml-0-2`. This was manually "integration tested":

```sh
docker run --rm -it ubuntu bash -c "apt update && apt install -y build-essential xz-utils curl git && mkdir /app && cd /app && git init && curl https://qlty.sh | sh && ~/.qlty/bin/qlty init && ~/.qlty/bin/qlty plugins enable rubocop && ~/.qlty/bin/qlty install"
```

And add `libyaml-0-2` to the apt install command:

```sh
docker run --rm -it ubuntu bash -c "apt update && apt install -y libyaml-0-2 build-essential xz-utils curl git && mkdir /app && cd /app && git init && curl https://qlty.sh | sh && ~/.qlty/bin/qlty init && ~/.qlty/bin/qlty plugins enable rubocop && ~/.qlty/bin/qlty install"
```

lsegal:
Actually the change in https://github.com/qltysh/qlty/pull/2009/commits/cd4a9fa80f73d25507cb8b3007b4231bdb738987 is the necessary integration test.

PR #2011, opened 2025-05-02
Add a `qlty coverage complete` sub-command

noahd1:
Intended to help support a new server-side aggregation model where client sends partial reports (with --incomplete) and then sends, at the end of the build, a `qlty coverage complete` to inform qlty.sh that the build has finished sending parts.

Some notes:

It takes a subset of the arguments that `qlty coverage publish` takes. Here's the current arguments:

image

Some arguments `qlty coverage publish` takes that `qlty coverage complete` does not:

--dry-run (only relevant when uploading coverage)
--format <FORMAT> (only relevant when uploading coverage)
--override-build-id <OVERRIDE_BUILD_ID> (this one I'm on the fence about ... `qlty coverage publish` actually _requires_ this argument, but I don't think it's necessary for this server side aggregation algorithm
--add-prefix (only relevant when uploading coverage)
--strip-prefix (only relevant when uploading coverage)
--print (prints coverage, and there's no coverage to print)
--total-parts-count (not relevant)
--incomplete (not relevant)

I had Claude help with this code, and there's a fair amount of duplication.

### DIFF
--- a/qlty-cli/src/commands/coverage/complete.rs
+++ b/qlty-cli/src/commands/coverage/complete.rs
@@ -4,6 +4,7 @@ use super::utils::{
 use crate::{CommandError, CommandSuccess};
 use anyhow::{bail, Context, Result};
 use clap::Args;
+use console::style;
 use qlty_cloud::Client as QltyClient;
 use qlty_coverage::{
     publish::{Plan, Planner, Settings},
@@ -55,6 +56,7 @@ impl Complete {
 
         let settings = self.build_settings();
 
+        self.print_section_header(" SETTINGS ");
         print_settings(&settings);
 
         let token = load_auth_token(&self.token, self.project.as_deref())?;
@@ -62,7 +64,10 @@ impl Complete {
 
         self.validate_plan(&plan)?;
 
+        self.print_section_header(" METADATA ");
         print_metadata(&plan, self.quiet);
+
+        self.print_section_header(" AUTHENTICATION ");
         print_authentication_info(&token, self.quiet);
 
         let timer = Instant::now();
@@ -70,6 +75,15 @@ impl Complete {
         self.print_complete_success(timer.elapsed().as_secs_f32());
 
         CommandSuccess::ok()
+    }
+
+    fn print_section_header(&self, title: &str) {
+        if self.quiet {
+            return;
+        }
+
+        eprintln!("{}", style(title).bold().reverse());
+        eprintln!();
     }
 
     fn build_settings(&self) -> Settings {
@@ -100,10 +114,7 @@ impl Complete {
             return;
         }
 
-        eprintln!(
-            "    Coverage marked as complete in {:.2}s!",
-            elapsed_seconds
-        );
+        eprintln!("    Coverage marked as complete in {elapsed_seconds:.2}s!");
         eprintln!();
     }
 
--- a/qlty-cli/src/commands/coverage/complete.rs
+++ b/qlty-cli/src/commands/coverage/complete.rs
@@ -104,7 +104,7 @@ impl Complete {
             "    Coverage marked as complete in {:.2}s!",
             elapsed_seconds
         );
-        eprintln!("");
+        eprintln!();
     }
 
     fn request_complete(
--- a/qlty-cli/src/commands/coverage/complete.rs
+++ b/qlty-cli/src/commands/coverage/complete.rs
@@ -1,0 +1,403 @@
+use crate::{CommandError, CommandSuccess};
+use anyhow::{anyhow, Context};
+use anyhow::{bail, Result};
+use clap::Args;
+use console::style;
+use git2::Repository;
+use qlty_cloud::Client as QltyClient;
+use qlty_config::version::LONG_VERSION;
+use qlty_config::{QltyConfig, Workspace};
+use qlty_coverage::ci::{GitHub, CI};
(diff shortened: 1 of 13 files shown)


brynary (commented):
I wouldn't worry about the duplication for the superficial code like printing output but I do worry about it for critical code like making API calls and extracting metadata like repository names.

I think that needs to be refactored out and the most important thing is deciding where the different concerns should live. Critically, Rust does not allow circular dependencies between crates, so the dependencies between crates need to form a DAG.

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// Override the branch from the CI environment
+    pub override_branch: Option<String>,
+
+    #[arg(long)]
+    /// Override the commit SHA from the CI environment
+    pub override_commit_sha: Option<String>,
+
+    #[arg(long)]
+    /// Override the pull request number from the CI environment
+    pub override_pr_number: Option<String>,
```
brynary, line 39: I would add override_build_id to keep it in sync with publish. L

Despite build ID not being required, it is useful, and we would prefer it be sent with every API call

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+        self.print_settings(&settings);
+
+        let token = self.load_auth_token()?;
+        let plan = Planner::new(&Self::load_config(), &settings).compute()?;
+
+        self.validate_plan(&plan)?;
+
+        self.print_metadata(&plan);
+
+        self.print_section_header(" PREPARING TO UPLOAD... ");
```
brynary, line 70: This print statement is wrong

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+        self.print_authentication_info(&token);
+
+        self.print_section_header(" UPLOADING... ");
```
brynary, line 73: Wrong

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+        } else if token.starts_with(COVERAGE_TOKEN_PROJECT_PREFIX) {
+            "Project Token"
+        } else if let Ok(oidc_regex) = Regex::new(OIDC_REGEX) {
+            if oidc_regex.is_match(token) {
+                "OIDC"
+            } else {
+                "Unknown"
+            }
+        } else {
+            "ERROR"
```
brynary, line 192: This level of complexity is about where I start to get concerned about duplication

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+        eprintln_unless!(self.quiet, "");
+    }
+
+    fn load_auth_token(&self) -> Result<String> {
+        self.expand_token(match &self.token {
+            Some(token) => Ok(token.to_owned()),
+            None => std::env::var("QLTY_COVERAGE_TOKEN").map_err(|_| {
+                anyhow::Error::msg("QLTY_COVERAGE_TOKEN environment variable is required.")
+            }),
+        }?)
```
brynary, line 214: Definitely need to remove duplication for auth token selection

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    fn request_complete(
+        &self,
+        metadata: &qlty_types::tests::v1::CoverageMetadata,
+        token: &str,
+    ) -> Result<Value> {
```
brynary, line 221: The qlty-cloud Crate is intended to be the thing that talks to the API. It defines the QltyClient struct.

Feels like this very particular error handling code should be pushed down into qlty-cloud crate

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+            Err(ureq::Error::Transport(transport_error)) => bail!(
+                "Transport Error: {}: {:?}",
+                client.base_url,
+                transport_error
+            ),
+        }
+    }
+
+    /// Appends repository name to token if it is a workspace token
+    fn expand_token(&self, token: String) -> Result<String> {
```
brynary, line 269: This logic is specifically for dealing with Workspace coverage tokens. With Workspace tokens, we need to work out the repository name via other metadata since the token isn't project specific.

Therefore, this concern feels like it should naturally live in the qlty-coverage crate

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -0,0 +1,403 @@
...
+        assert_eq!(token, "qltcw_123/test");
+        Ok(())
+    }
+
+    #[test]
+    fn test_expand_token_workspace_with_env() -> Result<()> {
+        let token = complete(None).expand_token("qltcw_123".to_string())?;
+        assert!(token.starts_with("qltcw_123/"));
+
+        std::env::set_var("GITHUB_REPOSITORY", "");
```
brynary, line 364: Modifying the environment during test execution will create flaky tests. Bad robot!

marschattha:
@noahd1 @brynary let me know if it looks good to you guys now, I may of had gone a little over zealous with the refactor.

PR #2025, opened 2025-05-05
Add skip_full_branch directive

marschattha:
Allows skipping enabled plugins on full branch analysis, for plugins which you want only to run on upstream diff mode. 

Useful for slow plugins which can't batch targets on full branch analysis.


brynary:
We believe there is a better path forward for the use case which spawned this, so I am going to close this for now. I am a bit anxious about unintended side effects of this kind of behavior. So we will keep it in our pocket in case we need it, but hold off for the moment.

PR #2030, opened 2025-05-06
Use String config_files instead of PathBuf

marschattha:


lsegal:
Just saw this PR but came across this recently in the other direction when implementing #2044.

I think there is value to encoding PathBufs in the QltyConfig over Strings, as this allows us to denote different deserialization methods for those fields. This will be important as we refactor 2044 into a more generic implementation where we start deserializing all PathBufs with _absolute paths_ (using a configurable "relative base" via serde attributes) to avoid lookup issues later down the line.

For that reason I think we probably don't want to be making this change.

brynary:
I generally agree with Loren. We'll generally be able to create a more reliable and easy to iterate system if we lean on rich types rather than primitives.

PR #2035, opened 2025-05-07
fix(php): auto install dependencies

TartanLeGrand:
This pull request enhances the PHP tool in `qlty-check` by adding automatic installation capabilities, improving error handling, and providing OS-specific installation instructions. The most important changes include introducing a new `try_install_php` method, implementing OS-specific installation logic, and refining error messages for better clarity.

### Enhancements to PHP installation and verification:

* **Automatic PHP installation**: Added a new `try_install_php` method to attempt automatic PHP installation if verification fails. This includes detecting the OS and invoking appropriate installation commands for Linux and macOS.

* **Linux-specific installation logic**: Implemented methods for installing PHP on Linux using `apt` or `yum`, with support for adding repositories and handling Docker environments.

* **macOS-specific installation logic**: Added a method to install PHP on macOS using Homebrew, with fallback instructions if Homebrew is not available.

### Improvements to error handling and messaging:

* **Enhanced error messages**: Updated error messages to provide more actionable feedback, including instructions for manual installation and troubleshooting.

* **Debugging support**: Added debug logs to track installation attempts, failures, and environmental conditions (e.g., OS type, Docker detection).

### Minor adjustments:

* **Cloning environment variables**: Updated the `verify_installation` method to clone the environment map for safer reuse.

### DIFF
--- a/qlty-check/src/tool/php.rs
+++ b/qlty-check/src/tool/php.rs
@@ -73,7 +73,7 @@ impl Tool for Php {
 impl Php {
     fn verify_installation(&self, env: HashMap<String, String>) -> Result<()> {
         let cmd = cmd!("php", "--version")
-            .full_env(env)
+            .full_env(env.clone())
             .unchecked()
             .stderr_to_stdout()
             .stdout_capture();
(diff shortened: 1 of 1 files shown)


brynary:
Hi @TartanLeGrand -- Thank you for this PR!

Stepping back for a moment, with our installation of linters and their runtimes (like Ruby, Node, PHP), we are trying to achieve sandboxing. With sandboxing, the user can have multiple versions of a given tool installed (e.g., Ruby 3.1 and Ruby 3.3) on the same system. Also, the tools installed by Qlty CLI are intended to be isolated from system dependencies. If things work OK, we'll never modify or break your system. As a nice security benefit, we do not need to run anything as `root`.

The approach we use for installing language runtimes varies by tool:

- For NodeJS, we download precompiled binaries
- We have two ways of installing Ruby: source compilation and binaries. We are working on improving the reliability of these installs on systems in different states

We haven't gotten sandboxed installations working for PHP. So, currently, it is an exception, and the user must install PHP themselves so it is available on their `$PATH` before running Qlty. I believe this is because we were unsuccessful in finding a reliable, sandboxed installation method for PHP.

If I am reading correctly, this PR looks like it installs PHP globally using the system package manager. Given the philosophy above, would it work for your usage to use an installer script instead to install PHP and Qlty? (e.g. something like `apt get install x y z && curl https://qlty.sh/ | sh`)

PR #2036, opened 2025-05-08
Print the report URL after the complete command runs

noahd1:
I also added a section header before printing the response information. Note the below is slightly inaccurate -- it says "MARKING COMPLETE ... " now.

image

### DIFF
--- a/qlty-cli/src/commands/coverage/complete.rs
+++ b/qlty-cli/src/commands/coverage/complete.rs
@@ -75,7 +75,8 @@ impl Complete {
         print_authentication_info(&token, self.quiet);
 
         let timer = Instant::now();
-        let result = Self::request_complete(&plan.metadata, &token).context("Failed to complete coverage")?;
+        let result = Self::request_complete(&plan.metadata, &token)
+            .context("Failed to complete coverage")?;
         self.print_complete_success(timer.elapsed().as_secs_f32(), &result.url);
 
         CommandSuccess::ok()
@@ -119,11 +120,11 @@ impl Complete {
         }
 
         eprintln!("    Coverage marked as complete in {elapsed_seconds:.2}s!");
-        
+
         if !url.is_empty() {
             eprintln!("    {}", style(format!("View report: {url}")).bold());
         }
-        
+
         eprintln!();
     }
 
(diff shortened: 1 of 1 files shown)


qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -10,9 +10,13 @@ use qlty_coverage::{
     publish::{Plan, Planner, Settings},
     token::load_auth_token,
 };
-use serde_json::Value;
 use std::time::Instant;
 
+#[derive(Debug, Default, Clone)]
+pub struct CompleteResult {
+    pub url: String,
+}
```
brynary, line 18: A more Rust-y way to do this would be to apply the newtype pattern using `enum`: https://www.lurklurk.org/effective-rust/newtype.html

However, it seems like just using a String would also be fine. I would reach for the newtype pattern when there's more stuff to do with the data m

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -107,21 +114,36 @@ impl Complete {
...
         let legacy_api_url = get_legacy_api_url();
         let client = QltyClient::new(Some(&legacy_api_url), Some(token.into()));
-        client.post_coverage_metadata("/coverage/complete", metadata)
+        let response = client.post_coverage_metadata("/coverage/complete", metadata)?;
+
+        let url = response
+            .get("data")
+            .and_then(|data| data.get("url"))
+            .and_then(|url| url.as_str())
+            .unwrap_or_default();
```
brynary, line 143: Returning an empty string and testing it with `is_empty` is OK but a bit of a fib to the type system since it's a missing value. Slightly better would be to return `Option<String>` and don't print if `None` is returned.

qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -69,8 +73,11 @@ impl Complete {
         print_authentication_info(&token, self.quiet);
 
         let timer = Instant::now();
-        Self::request_complete(&plan.metadata, &token).context("Failed to complete coverage")?;
-        self.print_complete_success(timer.elapsed().as_secs_f32());
+        let result = Self::request_complete(&plan.metadata, &token)
+            .context("Failed to complete coverage")?;
+
+        self.print_section_header(" MARKING COMPLETE...");
```
brynary, line 79: I would simplify to just  `COMPLETING...`

Also I believe you need a trailing space in the string in order for the colored terminal output to shade the background color properly

brynary (approved):
LGTm with some take or leave it feedback

brynary (approved):

PR #2037, opened 2025-05-08
feat: Add support for `--sarif` flag to `qlty smells` command

ujlbu4:
This PR adds support for generating SARIF output using the --sarif flag when running the qlty smells command.

Notes:

I'm not an expert in Rust, and I encountered a few challenges while implementing this feature, so happy for any comments to improve it. 

The main issue was that the existing `SarifFormatter` only supported a single type: `qlty_check::Report`. Since Rust doesn't have inheritance, I resolved this by extracting a shared trait (`SarifTrait`) and implementing it for both `qlty_check::Report` and `qlty_analysis::Report`. The `SarifFormatter` was then updated to accept any type that implements this trait.

I'm not completely confident that all the structs and traits are placed in the most appropriate modules, so I'm open to feedback or suggestions about how to better organize them according to project conventions or best practices.

### DIFF
--- a/qlty-analysis/Cargo.toml
+++ b/qlty-analysis/Cargo.toml
@@ -27,6 +27,7 @@ path-absolutize.workspace = true
 pbjson-types.workspace = true
 qlty-config.workspace = true
 qlty-types.workspace = true
+qlty-formats.workspace = true
 rand.workspace = true
 rayon.workspace = true
 regex.workspace = true
--- a/qlty-analysis/src/report.rs
+++ b/qlty-analysis/src/report.rs
@@ -191,8 +191,6 @@ impl Report {
     }
 }
 
-// impl SarifFormatter for Report {}
-
 impl SarifTrait for Report {
     fn issues(&self) -> Vec<Issue> {
         self.issues.clone()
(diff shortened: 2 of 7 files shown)


fivemhostingserver:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemtebex:
Great implementation of this feature. I particularly like the approach to error handling.

brynary (commented):
Thanks for this PR!

I agree there's no good reason that `qlty check` supports `--sarif` but `qlty smells` does not, so I think this is a good feature.

In the future, we are going to be extracting the maintainability analysis that is in `qlty smells` and `qlty metrics` into a separate crate so that it can integrate more cleanly. This will have then effect of avoiding these kinds of issues. But the timeline for that is not set yet and this is helpful in the meantime.

As far as the design goes, I would try and avoid introducing a trait. What if we update the `SarifFormatter` struct so that instead of storing a `Report` as its fields it instead stores `Vec<Message>` and `Vec<Issue>`?

ujlbu4:
Hey @brynary, thanks for the suggestion! It really makes sense and actually simplifies the code. 🙂

Not sure why I didn’t go that route initially, probably didn’t want to touch the original interfaces. Anyway, I’ve pushed your suggestions. Hope it looks better now! )

ujlbu4:
hopefully fixed tests, added missed import

brynary (approved):
This looks great! Will merge shortly.

brynary (approved):

PR #2038, opened 2025-05-08
Centralize API URL retrieval and allow overriding with env var

noahd1:
Necessary / helpful for local development when you want to point the QLTY API to localhost.


brynary (commented):
Is it intentional that _both_ API URLs are using the same `$QLTY_API_URL` env var?

noahd1:
> Is it intentional that _both_ API URLs are using the same `$QLTY_API_URL` env var?

Kind of. My thought was that if you want to test something locally, it's probably likely just a single endpoint or two you're testing, and in those cases you don't want to fumble around trying to guess the name of the env var for the particular use case. But i could be convinced to change it.

brynary (approved):
OK, we can go with this

PR #2044, opened 2025-05-13
Package and export configuration files from plugins

lsegal:
This change allows plugins to package and declare exported configuration files using the syntax below:

```toml
[plugins.definitions.some-plugin]
# Paths are relative to this TOML file location (plugin.toml, source.toml, etc)
exported_config_paths = ["relative/path/to/config.yml", "other/config.alternate.yml"]

# ...
```

Declaring an exported configuration allows a configuration file to be automatically included at runtime as a fallback default if no other configuration files are present or detected by the workspace (including `.qlty/configs` directory). The configuration file will be symlinked just like any other configuration. 

### Disabling Export Behavior from a Downstream Consumer

Users who want to opt out of automatic importing of configuration files can either:

1. Define a default configuration file in their workspace source tree (or `.qlty/configs`), or,
2. Override the `plugins.definitions.<plugin>.exported_config_paths` as:

```toml
[plugins.definitions.some-plugin]
exported_config_paths = []
```

### Using with Sources

This behavior is can used with custom sources to automatically include custom configuration from a given layered source:

#### \> .qlty/qlty.toml

```toml
[[source]]
name = "default"
default = true

[[source]]
name = "configs"
directory = "./configuration" # this can also be a repository
```

#### \> configuration/source.toml

```toml
[plugins.definitions.rubocop]
exported_config_paths = ["rubocop/.rubocop.yml"]
```

#### \> configuration/rubocop/.rubocop.yml

```yaml
AllCops:
  Enable: true
# ...
```

Any project including the above source will automatically receive custom rubocop configuration that will automatically be applied when running `qlty check`.

### DIFF
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -292,11 +292,11 @@ impl Executor {
             }
         }
 
-        let exported_config_files = self
+        let exported_config_paths = self
             .plan
             .invocations
             .iter()
-            .flat_map(|invocation| invocation.plugin.exported_config_files.clone())
+            .flat_map(|invocation| invocation.plugin.exported_config_paths.clone())
             .unique()
             .collect_vec();
 
@@ -349,7 +349,7 @@ impl Executor {
             }
         }
 
-        for config_file in &exported_config_files {
+        for config_file in &exported_config_paths {
             if self.plan.workspace.root != self.plan.staging_area.destination_directory {
                 // for formatters
                 let loaded_config_file = load_config_file_from_source(
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -29,6 +29,7 @@ use qlty_types::analysis::v1::{Issue, Message, MessageLevel};
 use rand::seq::SliceRandom;
 use rand::thread_rng;
 use rayon::prelude::*;
+use staging_area::load_config_file_from_source;
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering};
@@ -291,6 +292,14 @@ impl Executor {
             }
         }
 
+        let exported_config_files = self
+            .plan
+            .invocations
+            .iter()
+            .flat_map(|invocation| invocation.plugin.exported_config_files.clone())
+            .unique()
+            .collect_vec();
+
         debug!(
             "Walker found {} config and affects cache files in {:.2}s",
             repository_config_files.len(),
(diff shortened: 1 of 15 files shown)


qlty-check/src/executor/staging_area.rs
```diff
@@ -218,86 +218,71 @@ impl StagingArea {
...
+#[cfg(windows)]
+fn symlink(from: &Path, to: &Path) -> std::io::Result<()> {
+    std::os::windows::fs::symlink_file(from, to)
+}
 
-    Ok("".to_string())
+#[cfg(unix)]
+fn symlink(from: &Path, to: &Path) -> std::io::Result<()> {
+    std::os::unix::fs::symlink(from, to)
 }
```
brynary, line 286: I think we have this platform-aware symlink stuff in a few cases. Can we extract to a common utility function?

qlty-check/src/executor.rs
```diff
@@ -291,6 +292,14 @@ impl Executor {
             }
         }
 
+        let exported_config_files = self
+            .plan
+            .invocations
+            .iter()
+            .flat_map(|invocation| invocation.plugin.exported_config_files.clone())
+            .unique()
+            .collect_vec();
```
brynary, line 301: To better align with our current separatio of concerns, it feels like the `Planner` should figure out exactly which exported config files are going to be used and which ones are going to be ignored (uniquely), and then encode that data into the `Plan` so that the Executor does not need to touch `plugin.exported_config_files`

brynary:
> Declaring an exported configuration allows a configuration file to be automatically included at runtime as a fallback default if no other configuration files are present or detected by the workspace (including .qlty/configs directory). 

My concern with this rule is that some linters have multiple configs (e.g. a `.rubocop.yml` and a `.rubocopignore`). If the project adds a `.rubocop.yml`, it should prevent use of the exported `.rubocop.yml`. However, I think the project should be able to check in a `.rubocopignore` without losing the `.rubocop.yml`

It's possible there may be conflict issues where you do want to block the export, but we have a way to do that. With the rule as written, it would, be impossible for the project to contain a `.rubocopignore` while accepting an exported `.rubocop.yml`

lsegal:
> My concern with this rule is that some linters have multiple configs (e.g. a .rubocop.yml and a .rubocopignore). If the project adds a .rubocop.yml, it should prevent use of the exported .rubocop.yml. However, I think the project should be able to check in a .rubocopignore without losing the .rubocop.yml

I don't think the described behavior would occur. The wording might have been unclear, when I said "no other configuration files" I meant no other configuration file in the overridable locations (.qlty/configs, repo directory), but this is checked _per configuration file_. I believe we even have a specific test to override one config file but not another in a scenario where a source exports multiple files.

qlty-check/src/executor/staging_area.rs
lsegal: I think cleaning up symlink code should probably be a separate PR as it would probably necessitate a separate shared crate.

qlty-check/src/executor.rs
lsegal: @brynary I'm a little confused by this. This approach intentionally follows the same one to [collect config_files](https://github.com/qltysh/qlty/blob/8bbc6da9016b756857dc82344c9f75e24a66f88b/qlty-check/src/executor.rs#L266) so as to be consistent. If we changed this, should we also change the above line? If not, why the difference?

brynary: Ah, sorry for the confusion... I  think I would have the same observation about `collect_config_files` actually. However, I don't think we want to dig into that right now so let's leave this as is to confirm with the local code.

brynary (approved):

PR #2047, opened 2025-05-16
Only enable RadarLint for Ruby when a RadarLint config is found

brynary:
- Only enable RadarLint for Ruby when a RadarLint config is found
- Fix bug where we did not correctly set `suggested_mode = "comment"` for RadarLint


brynary:
@marschattha I totally missed it in review but we were setting a non-existant `suggeste_mode` property (typo). We need to figure out a way to turn an invalid / unknown config directive into a warning at runtime and an error at test time.

PR #2048, opened 2025-05-16
Get executable name from `env::args()` instead of `env::current_exe()`

lsegal:
On Linux, `env::current_exe()` uses `/proc/self/exe` to determine the executable location which may not be accessible on all Linux systems (where /proc is unmounted):

```
thread 'main' panicked at qlty-cli/src/commands/version.rs:21:51:
Unable to identify current executable: Error { kind: Uncategorized, message: "no /proc/self/exe available. Is /proc mounted?" }
```

Instead we can use `env::args().nth(0)` which reliably provides the name of the executable. Unlike current_exe, this path is not absolute, but fortunately we do not require this behavior (and we can resolve the path manually from `env::current_dir()` if needed).

Note that we do not change the API on Windows since /proc is not used on that platform.


brynary (approved):
👍 One thought, take or leave it... We could extract into a function and add a comment explaining why we do it for the benefit of. a future dev

brynary:
@lsegal given that we are no longer planning to chroot, do you want to revert this one?

PR #2051, opened 2025-05-16
tool: default extra_env_vars() impl should look at runtime().extra_env_vars() if available

lsegal:
This affects golangci-lint: env vars from go runtime are not being imported because the golangci-lint plugin is a GithubReleaseTool and not a GoPackageTool.


brynary (approved):
Seems OK. Any risk of side effects?

lsegal:
I wouldn't expect anything as env should be additive and only pulling from runtime which would be necessary anyway.

PR #2058, opened 2025-05-20
Fix loading deep config files into staging

marschattha:
We seem to have had a regression where loading deeply nested config files into staging directory was breaking due to parent dir not being present.


brynary (commented):
Thanks, @marschattha. I'm surprised by this regression. How could we trace it back to when it was introduced?

marschattha:
It was introduced by the refactor of staging_area.rs in https://github.com/qltysh/qlty/pull/2044

Added non-network integration test, lets see if it will work on windows CI

PR #2066, opened 2025-05-22
Fix formatter invocation directory when runs from target_directory

marschattha:
Since the computation for invocation directory is done in planning phase before files are staged (in the executor phase). 
Advanced directive like runs_from target_directory was not functioning as intended.


brynary (commented):
Thanks. A few questions...

How was this problem exposed?

Also is this a regression or something we believe never worked?

brynary (approved):

PR #2067, opened 2025-05-22
Add terraform plugin support

marschattha:
Depends on https://github.com/qltysh/qlty/pull/2066 for intended function


brynary (commented):
I think we need to add `.terraform.lock.hcl` as an `affects_cache` or config file and this otherwise LGTM!!

brynary (approved):

PR #2075, opened 2025-05-22
fix: Use system certificate roots for downloads

### LINKED ISSUE #2072: Bug: Downloads are not using system certificate roots
### What happened?

When trying to run Qlty CLI on OpenAI ChatGPT Codex, installing Go fails with this error:

```
 > Failed to install go@1.22.0
 > 
 > Caused by:
 >     0: Error installing go@1.22.0.
 >        
 >     1: Error downloading file from https://go.dev/dl/go1.22.0.linux-amd64.tar.gz
 >     2: https://go.dev/dl/go1.22.0.linux-amd64.tar.gz: Connection Failed: tls connection init failed: invalid peer certificate: UnknownIssuer
 >     3: invalid peer certificate: UnknownIssuer
```

The reason appears to be that we use `ureq` in `download.rs` to download files. However, per the [docs](https://docs.rs/ureq/latest/ureq/#root-certificates), by default ureq uses Mozilla's root TLS certificates. Since Codex requires routing traffic through a TLS proxy, the certificate is not trusted, ureq thinks it is getting man-in-the-middled and errors.

To fix this, we need to enable the `platform-verifier` feature for the Crate  as well as configure a ureq::Agent to use it. We need up update all the places we use ureq to use `RootCerts::PlatformVerifier`.

Qlty CLI would work with HTTPS proxies by respecting the system certificate roots.

Yes, consistently

### CLI Version

Latest

### Relevant log output

```shell

```

brynary:
Enable platform-verifier feature for ureq and create HTTP helper modules that use rustls_platform_verifier to respect system certificate stores.

This fixes downloads failing on OpenAI ChatGPT Codex and other environments that use TLS proxies with custom certificate authorities.

Changes:
- Added platform-verifier feature to ureq in workspace Cargo.toml  
- Created HTTP helper modules in all crates that use ureq
- Updated all ureq::get(), ureq::post(), and ureq::put() calls
- System certificate roots now used instead of Mozilla's built-in ones

Fixes #2072

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,7 +136,7 @@ tree-sitter-rust = "0.21.2"
 tree-sitter-swift = "0.5.0"
 tree-sitter-typescript = "0.21.2"
 trycmd = "0.15.8"
-ureq = { version = "2.12.1", features = ["json", "proxy-from-env"] }
+ureq = { version = "2.12.1", features = ["json", "proxy-from-env", "platform-verifier"] }
 url = "2.5.4"
 uuid = { version = "1.14.0", features = ["v4"] }
 walkdir = "2.5.0"
--- a/qlty-check/src/tool.rs
+++ b/qlty-check/src/tool.rs
@@ -2,6 +2,7 @@ pub mod command_builder;
 mod download;
 mod github;
 pub mod go;
+mod http;
 mod installations;
 pub mod java;
 pub mod node;
--- a/qlty-check/src/tool/download.rs
+++ b/qlty-check/src/tool/download.rs
@@ -1,3 +1,4 @@
+use super::http;
 use super::installations::initialize_installation;
 use super::installations::write_to_file;
 use super::Tool;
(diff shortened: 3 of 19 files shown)


brynary:
@claude the unit tests are failing. Please fix them

brynary:
@claude get this code compiling and passing tests

brynary:
@claude change the way that the http module is imported.

Instead of qlty_config::http::get which is verbose, add a use statement to bring in qlty_config::http so it can be called as http::get. Apply this change everywhere

Cargo.toml
```diff
@@ -84,6 +84,8 @@ rand = "0.8.5"
 rayon = "1.10.0"
 regex = "1.11.0"
 rust-embed = { version = "8.7.2", features = ["include-exclude"] }
+rustls = "0.23.19"
```
brynary, line 87: @claude the latest version of rusttls is v0.23.27. Let's use that please

brynary:
@claude format this with `qlty fmt`

PR #2076, opened 2025-05-22
build: Disable source fetch by default since fetch happens during qlty install

lsegal:


brynary:
@lsegal to clarify is the intention here to allow `qlty check` to disable the source fetch behavior or to change the default behavior of `qlty check`?

lsegal:
@brynary the default behavior of qlty build is what was changed. The opt out in qlty check is an added benefit since the flag was added to Settings.

lsegal:
Actually I think this can be reworked a bit.

lsegal:
@brynary check again, this makes more sense now.

PR #2078, opened 2025-05-23
Add log for ignored coverage paths

marschattha:
Currently if any file coverages are ignored there is no way to communicate that leading to confusion among customers and support alike. 

This PR adds log line when coverage files are ignored.

### DIFF
--- a/qlty-cli/src/commands/coverage/publish.rs
+++ b/qlty-cli/src/commands/coverage/publish.rs
@@ -381,6 +381,18 @@ impl Publish {
             }
         );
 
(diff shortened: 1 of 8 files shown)


brynary (approved):

brynary:
@claude can you please fix the tests for this? You need to run the tests with `TRYCMD=overwrite` to update the snapshots, and be sure to make sure the tests pass before you push. Only update the tests -- the code is fine. If you can't get it working, don't push anything.

brynary:
@marschattha hah, well Claude is going to be confused 😂

PR #2082, opened 2025-05-23
fix: Improve accuracy of field counting for Java

### LINKED ISSUE #1396: Bug: Incorrect field count in Java source file when multiple classes have the same field name
### What happened?

With a Java source file defined as follows (note that both classes define `foo` and `bar` as fields):

```Java
class BooleanLogic {
    int foo;
    int bar;
}

class BooleanLogic1 {
    boolean foo;
    boolean bar;
    boolean baz;
    boolean qux;
}
```

Running `qlty metrics --all` reports only 4 fields even though 6 fields are present.

```
 name           | classes | funcs | fields | cyclo | complex | LCOM | lines | LOC
----------------+---------+-------+--------+-------+---------+------+-------+-----
 SomeClass.java |       2 |     0 |      4 |     1 |       0 |    0 |    11 |  10
 TOTAL          |       2 |     0 |      4 |     1 |       0 |    0 |    11 |  10
```

Changing the source file to one with 6 uniquely named fields across the classes:

```Java
class BooleanLogic {
    int zee;
    int zoo;
}

class BooleanLogic1 {
    boolean foo;
    boolean bar;
    boolean baz;
    boolean qux;
}
```

correctly reports 6 fields.

```
 name           | classes | funcs | fields | cyclo | complex | LCOM | lines | LOC
----------------+---------+-------+--------+-------+---------+------+-------+-----
 SomeClass.java |       2 |     0 |      6 |     1 |       0 |    0 |    11 |  10
 TOTAL          |       2 |     0 |      6 |     1 |       0 |    0 |    11 |  10
```

I'm unsure if this bug is restricted to Java and the "fields" count metric. In Java, a source file can only have 1 _public_ class defined in it with the same name as the file. However, it is possible to define more than 1 class in the file if the other classes are not public. Unsure if this is common. If not, and if restricted to Java, this is a relatively minor bug.

Count all fields regardless of name

Yes, consistently

### CLI Version

qlty 0.466.0 macos-arm64 (354ff80 debug 2025-01-03)

### Relevant log output

```shell
See issue description
```

brynary:
Fixes the Java field counting bug where duplicate field names across classes were incorrectly deduplicated.

Previously, when multiple classes in the same Java file had fields with the same name, the field counting logic would deduplicate by field name across the entire file, leading to incorrect counts.

This change modifies the field counting logic to count each Java field declaration individually without deduplication while preserving existing behavior for other languages.

Fixes #1396


brynary:
@claude refactor as described in the code review feedback

noahd1:
Seems much improved from where this was last I looked (when it conditionalized based on the language).

PR #2090, opened 2025-05-24
Segregate Coverage and analysis ignores

marschattha:
Right now exclude_patterns get copied over into coverage.ignores for coverage and there is no good way to exclude files or folders for analysis and not coverage.

Actually thinking more about it the rollout might be tricky as this can be considered a breaking change, not sure if how many people will be affected by this.

### DIFF
--- a/qlty-cli/src/initializer/templates/qlty.toml
+++ b/qlty-cli/src/initializer/templates/qlty.toml
@@ -53,5 +53,33 @@ test_patterns = [
   "**/spec_*.*",
 ]
 
+coverage.ignores = [
+  "*_min.*",
+  "*-min.*",
+  "*.min.*",
+  "**/.yarn/**",
+  "**/*.d.ts",
+  "**/bower_components/**",
+  "**/build/**",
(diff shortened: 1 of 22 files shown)


marschattha:
> Second, to minimize backwards compatibility breakage issues, what if we translate exclude patterns into coverage exclude patterns temporarily with a warning?

But then that will basically undo what this PR aims to achieve, being able to separate coverage exclusions from analysis exclusions, as your analysis exclusions will become part of the coverage exclusions.

How about if there are no coverage exclusions we default to analysis exclusions and if any coverage exclusions are provided we use only those and skip analysis exclusions?

marschattha:
> @marschattha ah yes I understand what you mean about undoing this change. Do we have information on the _server side_ for when coverage data is being excluded due to global exclude patterns?

I don't think we have any data regarding that on serverside.

brynary:
@marschattha I think the conclusion from today is we are going to hold off on this for now until we collect more info, so closing this PR

PR #2099, opened 2025-05-28
fix: shellcheck considers tabs as 8 columns in location ranges

### LINKED ISSUE #2100: Bug: Shellcheck patch generation fails for lines containing tab characters
## Description

Shellcheck reports column positions that are incorrect for lines containing tab characters, causing patch generation to fail. Shellcheck treats each tab character as occupying 8 column positions when reporting coordinates, but the actual file contains only 1 character (the tab itself).

## Problem

When shellcheck analyzes a file with tab characters and suggests fixes, the column positions in the output are calculated as if each tab takes up 8 character positions. However, when qlty tries to apply these suggestions as patches, it uses the actual file content where tabs are single characters, causing a mismatch in positioning.

## Example

Given a shell script line:
```bash
	sudo su - ${username} -c whoami
```
Where `	` represents a tab character.

Shellcheck reports the variable `${username}` as being at columns 19-30, but in the actual file content, it's at positions 12-23 (tab + "sudo su - " = 12 characters).

## Impact

- Patch suggestions from shellcheck cannot be applied correctly
- Auto-fix functionality doesn't work for shell scripts with tabs
- Manual fixes are required even when shellcheck provides suggestions

## Root Cause

The column adjustment calculation needs to account for the difference between shellcheck's tab width (8 characters) and the actual tab representation (1 character) in the file.

## Solution

For each tab character before the reported column position, subtract 7 from the column number (8 - 1 = 7) to get the correct file position.

## Related

- Fixed by: #2099
- Referenced issue: #2073

marschattha:
https://github.com/qltysh/qlty/issues/2073

So it seems Shellcheck adds 8 to the column for every tab (\t) char which we need to adjust for. 
Unfortunately can't be done in parser since we don't have access to file contents at that point, so instead had to do this hackaround in the patch builder.

Fixes #2100

### DIFF
--- a/qlty-check/src/planner.rs
+++ b/qlty-check/src/planner.rs
@@ -13,6 +13,7 @@ use check_filters::CheckFilters;
 use console::style;
 use document_url_generator::DocumentUrlGenerator;
 use itertools::Itertools;
+use plugin_tab_column_width_transformer::PluginTabColumnWidthTransformer;
 use qlty_analysis::cache::{Cache, FilesystemCache, NullCache};
 use qlty_analysis::git::{compute_upstream, DiffLineFilter};
 use qlty_analysis::workspace_entries::TargetMode;
@@ -23,6 +24,7 @@ use qlty_types::analysis::v1::ExecutionVerb;
 use qlty_types::{category_from_str, level_from_str};
 use rayon::prelude::*;
 use std::collections::HashMap;
+use std::sync::Arc;
 use std::time::Instant;
 use tracing::{debug, info};
 
@@ -37,6 +39,7 @@ mod level_filter;
 mod plan;
 mod plugin;
 mod plugin_mode_transformer;
+mod plugin_tab_column_width_transformer;
 mod plugin_workspace_entry_finder_builder;
 pub mod source_extractor;
 pub mod target;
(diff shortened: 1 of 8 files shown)


brynary:
@claude create a Bug issue for the problem that this PR intends to solve

brynary:
@marschattha a couple of questions... does this seem like a shellcheck bug?

Does shellcheck emit information with character offsets instead of X/Y row/column points?

marschattha:
> @marschattha a couple of questions... does this seem like a shellcheck bug?
> 
> Does shellcheck emit information with character offsets instead of X/Y row/column points?

Yup this is a Shellcheck bug/feature.
No there is no offsets in the output, see [sample](https://github.com/qltysh/qlty/pull/2099/files#diff-86761789cbf0f9e89b7a42f1335941fce5c98a2d8d5bd9914464331d1653db74).

marschattha:
It is for issue locations as well.

qlty-check/src/planner.rs
```diff
@@ -268,6 +271,17 @@ impl Planner {
             plugins: self.config.plugin.clone(),
         }));
 
+        for plugin in &self.active_plugins {
+            if let Some(tab_column_width) = plugin.plugin.tab_column_width {
```
brynary, line 275: Rather than having this be nullable, which causes us to have to juggle Some/None, maybe we make it just always an int with a default `1` and then only add the transformer if the value is `> 1`. The reason I am suggesting this is because it seems like there is never a difference between `1` and `None`. WDYT?

brynary (approved):

qlty-check/src/planner.rs
marschattha: I thought using `1` might look more like a magic number, where as `None` makes it more clear that there isn't going to be a need for this directive while being more idiomatic in rust terms.

marschattha: Gonna merge this so we can release.

brynary:
@marschattha do we still need to update the shellcheck plugin definition?

PR #2111, opened 2025-05-30
fix: add more support for simplecov parser

marschattha:
https://github.com/qltysh/support/issues/98


brynary (approved):
This seems OK, but the code in the `SimplecovJson` is a bit hard to follow. It seems like we have full unit test coverage on it, so I wonder if the robot might have some ideas on how to  refactor it to be easier to read while keeping all of the tests passing.

brynary:
@marschattha can you please change the PR title to something in this form:

> fix: [description of the bug]

That will cause the `metadata / conventional` check to pass

PR #2132, opened 2025-06-04
feat: allow batching of build invocation files

marschattha:
Depends on https://github.com/qltysh/cloud/pull/6333


marschattha:
https://github.com/qltysh/qlty/pull/2149

PR #2133, opened 2025-06-05
fix: add limit to snippet size

marschattha:
We can tweak the `MAX_SNIPPET_LENGTH` but there should be one to prevent issues during ingestion.


marschattha:
Went with https://github.com/qltysh/qlty/pull/2135 instead

PR #2137, opened 2025-06-06
fix: ESLint plugin is missing support for TypeScript eslint configuration (eslint.config.ts)

kolarski:
ESLint fails when TypeScript config is used with: 
```
 Value for 'config' of type 'path::String' required.  You're using eslint.config.js, some command line flags are no longer available. Please see https://eslint.org/docs/latest/use/command-line-interface for details.eslint  Error  Exited with code 2 in 0.20s  .qlty/out/invoke-sEJm5a.yaml

```
Because it generates wrong command: 
```
script: sh -c "eslint --config  --output-file /tmp/invocation-out-sEJm5a.txt --format json src/'"

```

The config parameter is missing. 

Ideally If config is not found it should not include --config parameter at all. The current fix adds support for TypeScrtipt ESLint

Issue: https://github.com/qltysh/qlty/issues/1542

### DIFF
--- a/qlty-plugins/plugins/linters/eslint/plugin.toml
+++ b/qlty-plugins/plugins/linters/eslint/plugin.toml
@@ -21,6 +21,7 @@ error_codes = [2]
 success_codes = [0, 1]
 config_files = [
   "eslint.config.js",
+  "eslint.config.ts",
   "eslint.config.mjs",
   "eslint.config.cjs",
 ]


kolarski:
Issue: https://github.com/qltysh/qlty/issues/1542

brynary (approved):

brynary:
Thanks, this looks good and we will include it in the next release.

(In the meantime, you can override this field in your qlty.toml file if you need it right away)

kolarski:
Oh, I didn't know I can do that!
I was looking at [https://docs.qlty.sh/qlty-toml#plugin](https://docs.qlty.sh/qlty-toml#plugin) and didn't find that overwrite was possible.  

This solved it for me and I confirm it is Qlty is now working correctly with the latest eslint with TypeScript config.
Here is the workaround for anyone interested. 
```
[[plugin]]
name = "eslint"
version = "9.28.0"
package_file = "package.json"
config_files = [
  "eslint.config.ts",
]
```

Thanks for your quick response! When the new release comes out I'll test it.

PR #2151, opened 2025-06-18
Allow publish and complete command to run Git free

marschattha:
The test harness for this was more complicated that the code itself.
If run in the same directory it picks the git directory of the main qlty repo, so needs to be run in a separate temp directory without git.


brynary (commented):
@marschattha this is a great start. Thanks for spiking this.

To accurately review and safely integrate this, I think we should break this into multiple PRs. Can we start with a PR which just does the `--override-author-time` behavior?

brynary:
Closing this one as replaced

PR #2167, opened 2025-06-23
feat: allow coverage publish to run without Git when using --override-commit-time

### LINKED ISSUE #2166: Allow coverage publish to run without Git when using --override-commit-time
Now that we have `--override-commit-time` flag in https://github.com/qltysh/qlty/pull/2155
coverage publish command needs to be able run without Git when `--override-commit-time` flag is provided, right now it errors when run without git on line https://github.com/qltysh/qlty/blob/c2fd37745f654a61a5cb72c1e709f3eaefa0027f/qlty-coverage/src/git.rs#L17

marschattha:
When --override-commit-time is provided, coverage publish now works without Git by:
- Making retrieve_commit_metadata() return Option<CommitMetadata> instead of failing
- Providing sensible defaults for author/committer info when Git is unavailable
- Requiring --override-commit-time when Git is not available

Resolves #2166


brynary:
@marschattha before we talk about the code, can we talk about the requirements? If a Git repository is not available on disk, what do we require instead? (This PR is focused specifically on `--override-commit-time` but I think there is other data we require as well.)

I recommend we list out the different _valid_ scenarios and _invalid_ scenarios. It seems like we have 3 potential sources of data: Git (now optional), the CI system's environment, CLI args (always overrides).

What is the minimum set of variables that we need for things to work?

marschattha:
So I believe the required variables right now are, branch name, commit SHA, build id and commit time. We get rest from CI env or override flags but commit time was the important one being extracted from git (we extract others but not required).

PR #2178, opened 2025-06-26
fix: treat exceeding MAX_ISSUES as a fatal error

### LINKED ISSUE #2177: Bug: Treat exceeding MAX_ISSUES as a fatal error
### What happened?

We have a limit MAX_ISSUES of 50,000. When this limit is exceeded, the system will stop invoking plugins and throw away any excess data. Today this is treated as a warning but not an error. However, when the system is behaving like this it is inherently randomized. Every build will produce a different set of 50,000 issues.

The proposal is we treated exceeding MAX_ISSUES as a fatal error and exit with a non-zero exit code in this case, which will cause the analysis to be considered FAILED.

Yes, consistently

### CLI Version

Latest

### Relevant log output

```shell

```

brynary:
Changes behavior when MAX_ISSUES (50,000) limit is exceeded:
- Previously: issued warnings and continued processing
- Now: treats as fatal error and exits with non-zero code

This ensures consistent analysis results instead of randomized truncation of issues.

Fixes #2177

### DIFF
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -193,7 +193,7 @@ impl Executor {
             &self.plan.hits,
             &invocations,
             self.plan.settings.skip_errored_plugins,
-        );
+        )?;
         let formatted = Self::build_formatted(&invocations);
 
         let messages = invocations
@@ -603,7 +603,7 @@ impl Executor {
         cache_hits: &[IssuesCacheHit],
         invocations: &[InvocationResult],
         skip_errored_plugins: bool,
-    ) -> Vec<Issue> {
+    ) -> Result<Vec<Issue>> {
         let mut issues = vec![];
 
         for cache_hit in cache_hits {
@@ -611,18 +611,14 @@ impl Executor {
                 issues.push(issue.to_owned());
 
                 if issues.len() >= MAX_ISSUES {
-                    warn!(
-                        "Maximum issue count of {} reached in cache, skipping further issues.",
-                        MAX_ISSUES
-                    );
-                    return issues;
+                    bail!("Maximum issue count of {} reached in cache. Execution halted. Please adjust your configuration to reduce the number of issues generated.", MAX_ISSUES);
                 }
             }
         }
 
         let mut errored_plugins = HashSet::new();
 
-        'invocation_loop: for invocation in invocations {
+        for invocation in invocations {
             if skip_errored_plugins && invocation.status() != InvocationStatus::Success {
                 errored_plugins.insert(invocation.invocation.plugin_name.clone());
             }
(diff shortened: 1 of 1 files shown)


brynary:
@davehenton the problem with the 50k limit as we discussed is that when you exceed it the results will become seemingly random (you'll get a slightly different 50k every time), which is confusing

This PR makes it a fatal error when we hit the 50k limit meaning we will consider the Build failed and will not process the results.

Therefore the debugging steps will involve using the CLI and modifying the config to reduce the issue count.

What do you think?

(As a follow up, we can probably improve the user messaging to include, for example, a summary of issue counts by tool and category to make it easier to know what to adjust)

brynary:
@claude when we hit this MAX_ISSUES error, it would be helpful to provide a summary of the count of issues per tool to the user so they understand what they need to adjust to revolve the problem

Can you please add a user friendly message which shows the count of issues by tool, sorted by count descending, to the error message?

Also let's include a docs URL: https://qlty.sh/d/too-many-issues

qlty-check/src/executor.rs
```diff
@@ -603,26 +603,22 @@ impl Executor {
...
             for issue in &cache_hit.issues {
                 issues.push(issue.to_owned());
 
                 if issues.len() >= MAX_ISSUES {
-                    warn!(
-                        "Maximum issue count of {} reached in cache, skipping further issues.",
-                        MAX_ISSUES
-                    );
-                    return issues;
+                    bail!("{}", Self::format_max_issues_error(&issues, " in cache"));
```
brynary: We can simplify this -- we don't need to use a different message based on the issues being in the cache. Use the same message every time.

qlty-check/src/executor.rs
```diff
@@ -665,6 +663,29 @@ impl Executor {
 
         Ok(())
     }
+
+    fn format_max_issues_error(issues: &[Issue], context: &str) -> String {
```
brynary: Remove the context arg

qlty-check/src/executor.rs
```diff
@@ -665,6 +663,29 @@ impl Executor {
...
+        let mut tool_summary: Vec<_> = tool_counts.into_iter().collect();
+        tool_summary.sort_by(|a, b| b.1.cmp(&a.1));
+
+        let tool_summary_text = tool_summary
+            .iter()
+            .map(|(tool, count)| format!("  {} ({} issues)", tool, count))
+            .join("\n");
+
+        format!(
+            "Maximum issue count of {} reached{}. Execution halted.\n\nIssue count by tool:\n{}\n\nPlease adjust your configuration to reduce the number of issues generated.\nFor more information: https://qlty.sh/d/too-many-issues",
```
brynary: Localize the max issues number with commas (so `50,000` instead of `50000`)

qlty-check/src/executor.rs
```diff
@@ -665,6 +663,29 @@ impl Executor {
...
+        for issue in issues {
+            *tool_counts.entry(issue.tool.clone()).or_insert(0) += 1;
+        }
+
+        let mut tool_summary: Vec<_> = tool_counts.into_iter().collect();
+        tool_summary.sort_by(|a, b| b.1.cmp(&a.1));
+
+        let tool_summary_text = tool_summary
+            .iter()
+            .map(|(tool, count)| format!("  {} ({} issues)", tool, count))
```
brynary: Localize the counts with commas as thousands separators

brynary (commented):
@claude address the code review feedback and provide an example of the output

PR #2189, opened 2025-07-08
Add non sandbox directive for node packages

marschattha:


brynary:
@marschattha I am going to close this for now because I don't think we are ready to commit to supporting non-sandbox mode

If/when we do decide to add support for that, we can refer to this.

For reference, does this work? If we wanted to support non-sandbox mode in the CLI, what other tasks would we need to complete besides this code?

PR #2219, opened 2025-07-14
update URLs for paths fixing doc

davehenton:
This updates the outdates URLs in our error output to point to the correct docs page (https://docs.qlty.sh/coverage/path-fixing)

### DIFF
--- a/qlty-cli/src/commands/coverage/publish.rs
+++ b/qlty-cli/src/commands/coverage/publish.rs
@@ -459,7 +459,7 @@ impl Publish {
 
             eprintln!(
                 "    {}",
-                style("https://qlty.sh/d/coverage-path-fixing").dim()
+                style("https://docs.qlty.sh/coverage/path-fixing").dim()
             );
 
             eprintln!();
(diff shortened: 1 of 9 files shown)


davehenton:
this fixes #2221

marschattha:
Actually rather than this, we should maybe fix it on the cloud side by updating this line
https://github.com/qltysh/cloud/blob/main/remix/app/routes/d.%24/route.tsx#L36

https://qlty.sh/d/coverage-path-fixing is supposed to be a shorter url that redirects to the correct one, which is https://docs.qlty.sh/coverage/path-fixing.

Although we could avoid using the `/d/` path altogether as it is not all that short.

davehenton:
closing this out -- now resolved

brynary (changes requested):
The intention is that for any URLs we are emitting from tools like the CLI, we want to use "jump links" in the form of: https://qlty.sh/d/SLUG

The reason is that we want to be able to re route the URLs to any destination we choose without waiting for everyone to upgrade the CLI

Therefore the fix is to update the jump link routing stored in the Cloud app

PR #2220, opened 2025-07-14
feat: enable validation by default for qlty coverage publish

### LINKED ISSUE #2218: add validation for qlty coverage publish by default
Currently, the `qlty coverage publish` command will not validate unless you pass `--validate` as an argument on the command line.

We should change this behavior so that coverage reports are validated by default (without needing to pass in that  `--validate` argument). Here are the requirements:

1. The `qlty coverage publish` command should validate by default
2.  The `--validate` argument is ignored (even though that argument will no longer be needed, we shouldn't fail if it is still being called) 
3.  Add a new `--no-validate` argument which can be used to _disable_ validation.

davehenton:
Add validation by default for qlty coverage publish command

## Summary
- Validation is now enabled by default for `qlty coverage publish`
- Added `--no-validate` flag to disable validation
- Deprecated `--validate` flag (still accepted with warning)

## Changes
- Modified Publish struct to validate by default
- Added should_validate() method returning !self.no_validate
- Added deprecation warning for --validate usage
- Updated tests to maintain existing behavior with --no-validate

## Test Plan
- [x] Verify `qlty coverage publish` validates by default
- [x] Verify `--validate` shows deprecation warning but works
- [x] Verify `--no-validate` disables validation
- [x] All existing tests updated and passing

Closes #2218


brynary:
@davehenton this is very close. Claude went a bit overboard on some of the testing changes, which we can clean up by hand, but the guts of this change is accurate

brynary (approved):

PR #2222, opened 2025-07-15
feat: Point to debug logs on lint error

marschattha:


brynary (changes requested):
@marschattha can you please change this to a "jump link":

https://qlty.sh/d/lint-error

Then we can re route it anytime

brynary (approved):
Looks like the test snapshot needs to be updated but LGTM generally

PR #2240, opened 2025-07-23
fix: honor ZDOTDIR environment variable in install.sh

### LINKED ISSUE #2241: Bug:
### What happened?

When using the qlty install script with a custom zsh configuration directory (set via `$ZDOTDIR`), the installer writes the `PATH` configuration to the wrong .zshrc file. The script hardcodes `$HOME/.zshrc` instead of respecting the `$ZDOTDIR` environment variable.

This means users who have customized their zsh setup using `$ZDOTDIR` will not have qlty added to their `PATH`, and the installation appears to succeed but qlty won't be available in new shell sessions.

The install script should respect the `$ZDOTDIR` environment variable and write to `$ZDOTDIR/.zshrc` when it's set, falling back to `$HOME/.zshrc` when it's not set.

Yes, consistently

### CLI Version

0.556.0 macos-x64 (9ed3b48 2025-07-17)

### Relevant log output

```shell

```

voltechs:
fixes #2241

## Summary
- Updates the zsh configuration path in install.sh to respect the `$ZDOTDIR` environment variable
- Uses the pattern `"${ZDOTDIR%/:-$HOME}"/.zshrc` instead of hardcoded `$HOME/.zshrc`
- Maintains backward compatibility by falling back to `$HOME` when `$ZDOTDIR` is not set

## Problem
The installer script hardcoded the zsh configuration file path to `$HOME/.zshrc`, which doesn't work for users who have customized their zsh configuration directory using the `$ZDOTDIR` environment variable.

## Solution
Updated the zsh configuration path to use `"${ZDOTDIR%/:-$HOME}"/.zshrc` which:
- Uses `$ZDOTDIR/.zshrc` when the variable is set (removing any trailing slash)
- Falls back to `$HOME/.zshrc` when `$ZDOTDIR` is not set or empty
- Follows standard shell practices for handling directory environment variables

## Test plan
- [x] Verified the change compiles with `cargo check`
- [x] Ran full test suite with `cargo test` - all 519 tests pass
- [x] Follows conventional commit format per contribution guidelines
- [x] No functional changes to Rust codebase, only shell script improvement

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -252,7 +252,7 @@ fish)
 zsh)
 	# Install completions, but we don't care if it fails
 	SHELL=zsh $exe completions --install >/dev/null 2>&1 || :
-	zsh_config=$HOME/.zshrc
+	zsh_config="${ZDOTDIR%/:-"$HOME"}/.zshrc"
 	tilde_zsh_config=$(tildify "$zsh_config")
 	{
 		printf '\n# qlty\n'


voltechs:
@brynary @noahd1 Not sure who takes the lead on these things. Just wanted to throw this on your radar. Pretty low hanging fruit, IMO.

Hope all is well!

brynary (approved):

PR #2265, opened 2025-07-31
feat: add zizmor plugin

marschattha:
https://github.com/qltysh/qlty/issues/2264

### DIFF
--- a/qlty-plugins/plugins/linters/zizmor/fixtures/__snapshots__/basic_v1.11.0.shot
+++ b/qlty-plugins/plugins/linters/zizmor/fixtures/__snapshots__/basic_v1.11.0.shot
@@ -4,7 +4,7 @@ exports[`linter=zizmor fixture=basic version=1.11.0 1`] = `
 {
   "issues": [
     {
-      "category": "CATEGORY_LINT",
+      "category": "CATEGORY_VULNERABILITY",
       "documentationUrl": "https://docs.zizmor.sh/audits/#excessive-permissions",
       "level": "LEVEL_MEDIUM",
       "location": {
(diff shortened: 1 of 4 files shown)


qlty-plugins/plugins/linters/zizmor/plugin.toml
```diff
@@ -0,0 +1,22 @@
...
+file_types = ["github-workflow"]
+latest_version = "1.11.0"
+known_good_version = "1.11.0"
+version_command = "zizmor --version"
+config_files = ["zizmor.yml"]
+description = "Static analysis tool for GitHub Actions"
+security = true
+
+[plugins.definitions.zizmor.drivers.lint]
+script = "zizmor --format=sarif ${target}"
```
brynary, line 15: I think Zizmore has a few different modes like offline vs. online.. I think we should use offline mode by default (faster, rate limits)

marschattha:
One thing I realized while testing this, if it gets an empty.yml file or a file which does not pass its github action validation it throws exit code 1 which is the same as all other error codes so if someone has an empty file, it is going to crash.

qlty-plugins/plugins/linters/zizmor/fixtures/__snapshots__/basic_v1.11.0.shot
```diff
@@ -0,0 +1,42 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`linter=zizmor fixture=basic version=1.11.0 1`] = `
+{
+  "issues": [
+    {
+      "category": "CATEGORY_LINT",
```
brynary: Can we please override these to Vulnerability category?

PR #2269, opened 2025-08-01
chore: Infer referenceType based on CoverageMetadata

noahd1:


noahd1:
I'm not sure why the tests are failing but when I run the tests locally, my Cargo.lock is dirty. Not sure why it's dirty or what the process should look like when it's in this state.

brynary (approved):

PR #2270, opened 2025-08-01
chore: Add ability to override git tag

noahd1:
In cases where there's no git tag (or the wrong git tag) from the CI environment, this override-git-tag option allows a user to specify an explicit value.

Also prints the git tag in the metadata section of the output and the override to the settings section.


qlty-cli/src/commands/coverage/complete.rs
```diff
@@ -43,6 +43,10 @@ pub struct Complete {
     /// Override the commit time from git metadata. Accepts a Unix timestamp (seconds since epoch) or RFC3339/ISO8601 format
     pub override_commit_time: Option<String>,
 
+    #[arg(long)]
```
brynary, line 46: Should this be marked as conflicts_with override-branch and override-pull-request-number?

noahd1: I added it but then reverted it because, I believe it would create an unusable state at the moment due to our current branch requirement.

I was going to consider validations, including the one you mentioned, as my next (and hopefully last) PR on this topic.

PR #2273, opened 2025-08-02
fix: Remove branch requirement, instead require branch, tag OR pull request

noahd1:
We now support sending through a git tag if one is detected from CI or provided with a `--override-git-tag` command line argument. This allows us to remove a branch as a requirement, and instead require either a branch, git tag or pull request as a reference.

Closes https://github.com/qltysh/cloud/issues/6677


brynary (approved):
Does this still allow for the common case of PR + branch?

noahd1:
> Does this still allow for the common case of PR + branch?

That’s the intention yes. The only “illegal” combinations at the moment is git tag with either branch or pull request.

PR #2283, opened 2025-08-04
chore: Introduce no-validate option for safe deploy of GitHub Action

noahd1:
Accepts (but for now does nothing) with a `--no-validate` option (current behavior is to not validate by default)


brynary (approved):
I might add a conflicts_with but your call

PR #2307, opened 2025-08-11
chore: fix for Rust 1.89 and lint fixes

marschattha:


noahd1 (approved):
I gave it a cursory glance -- Think we have to lean on our test suite here

PR #2310, opened 2025-08-11
fix: Merge enabled plugins by name and prefix

marschattha:
https://github.com/qltysh/qlty/issues/2309

### DIFF
--- a/qlty-config/src/config/builder.rs
+++ b/qlty-config/src/config/builder.rs
@@ -428,7 +428,7 @@ fn compute_unique_merged_enabled_plugins(plugins: &[EnabledPlugin]) -> Vec<Enabl
 }
 
 // sort the plugins by name, prefix, and version for consistency
-fn sort_enabled_plugins(plugins: &mut Vec<EnabledPlugin>) {
+fn sort_enabled_plugins(plugins: &mut [EnabledPlugin]) {
     plugins.sort_by(|a, b| {
         let name_cmp = a.name.cmp(&b.name);
         if name_cmp != std::cmp::Ordering::Equal {


brynary (commented):
Thanks @marschattha ! I have three questions and one request:

Question #1: How robust is this behavior? For example, if instead of _disabling_ the enabled plugin in the repository TOML, instead I change the mode from `block` to `comment`. Will that work? What about other changes to other EnabledPlugin settings? Ideally this PR would be a comprehensive resolution for these behaviors.

Question #2: This logic is in qlty-check but it feels like this concern makes more sense in qlty-config because qlty-config is responsible for config merging, right?

Question #3: Related, how will this be represented in `qlty config show`? Will the multiple enabled plugins be compressed together in the list?

Request: I feel like we are leaning too much on integration tests for small CLI behaviors. While they have some benefits, I find them a bit cumbersome to add, and a bit hard to follow what specific behavior they are intending to cover.

Can we try to change this from whole-CLI integration tests to state-based unit tests which exercise the ? This may benefit from some adjustments to the

brynary:
I'm also not seeing any references to `prefix` in the Rust code changes and I'm wondering if this is properly cascading when the prefixes match and _not_ cascading when the prefixes don't match.

marschattha:
> Question https://github.com/qltysh/qlty/issues/1: How robust is this behavior? For example, if instead of disabling the enabled plugin in the repository TOML, instead I change the mode from block to comment. Will that work? What about other changes to other EnabledPlugin settings? Ideally this PR would be a comprehensive resolution for these behaviors.

So the problem is since `disabled` is an enum it can hold only one value, since this enabled plugin merging happens after tomls are merged, there is no good way to know which mode should be prioritized for a given plugin right now. So right now it will keep whichever mode pops up first for a given plugin name and prefix.

> Question https://github.com/qltysh/qlty/issues/2: This logic is in qlty-check but it feels like this concern makes more sense in qlty-config because qlty-config is responsible for config merging, right?

Its in qlty-config/src/config/builder.rs right now.

> Question https://github.com/qltysh/qlty/issues/3: Related, how will this be represented in qlty config show? Will the multiple enabled plugins be compressed together in the list?

Yes it will show the merged/compressed form of the enabled plugin.

qlty-config/src/config/builder.rs
```diff
@@ -355,9 +358,97 @@ impl Builder {
...
+}
+
+fn compute_unique_merged_enabled_plugins(plugins: &[EnabledPlugin]) -> Vec<EnabledPlugin> {
+    let mut merged_unique_enabled_plugins: HashMap<(String, String), EnabledPlugin> =
+        HashMap::new();
+
+    for plugin in plugins.iter() {
+        let key = (
+            plugin.name.clone(),
+            plugin.prefix.clone().unwrap_or_default(),
```
marschattha, line 416: Prefix is selected here for generating the keying for unique for plugin name and prefix.

qlty-config/src/config/builder.rs
```diff
@@ -355,9 +358,97 @@ impl Builder {
...
+        triggers: existing
+            .triggers
+            .iter()
+            .chain(new.triggers.iter())
+            .cloned()
+            .collect(),
+        skip_upstream: existing.skip_upstream.or(new.skip_upstream),
+        package_file: existing.package_file.clone().or(new.package_file.clone()),
+        fetch: existing
+            .fetch
```
marschattha: So the merging behaviour is a bit tricky for strings and enums that can hold only one value, problem being which one is to be selected, since this happens after toml merging, right now no way to know which came from where, also someone can have two different enabled plugins blocks for same plugin name and prefix defined in their `qlty.toml` file similar to the integration test right now.

For array attributes I just merged the arrays.

marschattha: We may want to introduce a new priority variable? Or move this up somehow so it knows which config is coming from qlty.toml or sources, but that might be not so straight forward.

marschattha:
Todo: change the integration test to state-based unit tests.

brynary:
> So the problem is since disabled is an enum it can hold only one value, since this enabled plugin merging happens after tomls are merged, there is no good way to know which mode should be prioritized for a given plugin right now. So right now it will keep whichever mode pops up first for a given plugin name and prefix.

I think we need to solve this.

The EnabledPlugins are stored as a Vec, right?

Does it work yo prioritize later elements of the vector more so than earlier elements?

marschattha:
Yup the sequence is maintained, I prioritized the attributes of the last occurrence's attributes over previous, even changed array merging to instead of concatenating to use from latest if present.

PR #2330, opened 2025-08-18
feat: unify plugin enabler logic to check package files and filters

### LINKED ISSUE #2319: Bug: When enabling plugin through qlty plugins enable, package file and filters are not checked
Right now if you enable a plugin through `qlty plugins enable <plugin_name>` command, it does not account for package_file and package_filters which right now only happens when running `qlty init`, as part of the code which scans the codebase to determine which plugins should added to the generated config (qlty-cli/src/initializer/scanner.rs).

We need to unify the logic for enabling plugins. Ideally init is simply determining which plugins to enable, the runs an Enabler. The Enabler should be the exact same as we run for `qlty plugins enable`

marschattha:
Fixes #2319

Unified the logic for enabling plugins so that both `qlty init` and `qlty plugins enable` check for package files and filters.

## Changes
- Created PluginEnabler that scans for package files and filters when enabling plugins
- Updated qlty plugins enable command to use the unified enabler
- Now both commands check package.json/Gemfile for relevant packages
- Added package_file, package_filters, and prefix fields to plugin config when found
- Enhanced version detection from lockfiles (package-lock.json, yarn.lock, Gemfile.lock)

### DIFF
--- a/qlty-cli/src/commands/plugins/enable.rs
+++ b/qlty-cli/src/commands/plugins/enable.rs
@@ -1,14 +1,12 @@
 use crate::{
-    initializer::Settings, Arguments, CommandError, CommandSuccess, Initializer, QltyRelease,
+    initializer::{PluginEnabler, Settings},
+    Arguments, CommandError, CommandSuccess, Initializer, QltyRelease,
 };
 use anyhow::{Context, Result};
 use clap::Args;
 use console::style;
-use qlty_config::{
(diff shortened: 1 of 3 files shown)


marschattha:
@claude, can you refactor this and the existing code used by qlty init to reduce duplicate logic and ensure same code path is used by both?

marschattha:
https://github.com/qltysh/qlty/pull/2334

PR #2336, opened 2025-08-20
feat: automatically add local rubocop requires to config_files

marschattha:


qlty-cli/src/initializer/renderer.rs
```diff
@@ -149,6 +150,14 @@ impl Renderer {
             toml.push_str(&format!("mode = \"{}\"\n", plugin.mode.to_str()));
         }
 
+        if !plugin.config_files.is_empty() {
+            toml.push_str("config_files = [\n");
+            for config_file in plugin.config_files.iter().sorted() {
+                toml.push_str(&format!("  \"{config_file}\",\n"));
+            }
+            toml.push_str("]\n");
+        }
```
brynary, line 159: Can we please get a unit test for this in this file?

qlty-cli/src/initializer/initializer.rs
```diff
@@ -58,6 +58,30 @@ impl Initializer {
...
+                    if let Ok(custom_cops) =
+                        super::scanner::rubocop_config::extract_custom_cop_files(
+                            &self.settings.workspace.root,
+                            &plugin_def.config_files,
+                        )
+                    {
+                        config_files = custom_cops;
+                    }
+                }
+            }
```
brynary: I agree hard coding a RuboCop specific block into this file is problematic. We can discuss some ways to resolve tomorrow.

My initial idea is that we should handle this similar to the way that we handle customer _parsers_ for plugins....

We can define a trait like `CustomPluginEnabler` with a TBD interface. Then we define a `RuboCopEnabler impl CustomPluginEnabler` for the RuboCop specific logic.

To bring it all together we would then add a new property to the TOML plugin definition like this:

```
custom_enabler = "rubocop" # Use custom enabling logic
```

This would open the door for us to add custom enablers for any plugin where it is valuable without breaking the abstraction.

qlty-cli/src/initializer/initializer.rs
```diff
@@ -58,15 +59,33 @@ impl Initializer {
...
                 version: Some(installed_plugin.version.clone()),
                 package_file: installed_plugin.package_file.clone(),
                 package_filters: installed_plugin.package_filters.clone(),
                 prefix: installed_plugin.prefix.clone(),
                 mode: installed_plugin.mode,
-            });
+                ..Default::default()
+            };
+
+            // Get the plugin definition from scanner's sources_only_config
```
brynary: This comment matched the code in that it gets the plugin definition from the sources_only_config. Why do we do that?

(Ideally the comment should focus on the _why_ and the code is self-explanatory for the _what_)

PR #2338, opened 2025-08-21
feat(coverage): add automatic path fixing for CI/CD environments

brynary:
## Summary

- Implements automatic path fixing for code coverage reports in CI/CD environments
- Adds DefaultPathFixer transformer that removes common CI/CD path prefixes
- Automatically applies path fixing when manual options are not specified

## Details

This PR introduces the DefaultPathFixer transformer to automatically handle path normalization for code coverage reports generated in various CI/CD environments. When users don't specify manual path fixing options (`--path-replace` or `--path-prefix`), the system will automatically remove common CI/CD path prefixes to ensure coverage paths match the repository structure.

### Supported CI/CD Systems

The DefaultPathFixer recognizes and removes path prefixes from:
- CircleCI (`/home/circleci/`)
- GitHub Actions (`/home/runner/work/`, `/Users/runner/work/`)
- Travis CI (`/home/travis/build/`)
- Jenkins (`/var/lib/jenkins/workspace/`)
- GitLab CI (`/builds/`)
- Bitbucket Pipelines (`/opt/atlassian/pipelines/agent/build/`)
- AWS CodeBuild (`/codebuild/output/`)
- Azure Pipelines (`/home/vsts/work/`)

### Changes

1. **Added DefaultPathFixer transformer** (`qlty-coverage/src/transformer.rs`)
   - Implements automatic detection and removal of CI/CD path prefixes
   - Includes comprehensive unit tests for all supported CI systems

2. **Updated publish planner** (`qlty-coverage/src/publish/planner.rs`)
   - Applies DefaultPathFixer when no manual path options are provided
   - Maintains backward compatibility with existing manual configuration

3. **Updated transform planner** (`qlty-coverage/src/transform/planner.rs`)
   - Applies DefaultPathFixer when no manual path options are provided
   - Ensures consistent behavior across both publish and transform commands

## Test Plan

- [x] Unit tests for DefaultPathFixer with all supported CI path prefixes
- [x] Integration tests for publish command with DefaultPathFixer
- [x] Integration tests for transform command with DefaultPathFixer
- [x] Verified backward compatibility with manual path options
- [ ] Manual testing in actual CI/CD environments
- [ ] Documentation update for new automatic path fixing feature

### DIFF
--- a/qlty-coverage/src/transformer.rs
+++ b/qlty-coverage/src/transformer.rs
@@ -240,18 +240,11 @@ impl DefaultPathFixer {
     }
 
     fn clean_windows_prefix(&self, path: &str) -> String {
-        // Remove common Windows drive prefixes like C:/, D:/, etc.
-        if let Some(stripped) = path.strip_prefix("C:/") {
-            return stripped.to_string();
-        }
-        if let Some(stripped) = path.strip_prefix("D:/") {
-            return stripped.to_string();
-        }
-        // Handle any single letter drive
         if path.len() >= 3 && path.chars().nth(1) == Some(':') && path.chars().nth(2) == Some('/') {
-            return path[3..].to_string();
+            path[3..].to_string()
+        } else {
+            path.to_string()
         }
-        path.to_string()
     }
 }
 


noahd1:
@brynary I did a risk assessment of this pull request and also some thoughts on possible enhancements. Take away is that this is probably good to go as is.

# Risks

## Risk: over-eager pattern matching which transforms paths that shouldn't get transformed

These are the patterns:

```
            r"^/?home/circleci/project/",
            r"^/?(home|Users)/runner/work/[^/]+/[^/]+/",
            r"^/?(home|Users)/travis/build/[^/]+/[^/]+/",
            r"^/?(home|Users)/jenkins/jobs/[^/]+/workspace/",
            r"^/?Users/distiller/[^/]+/",
            r"^/?(home|Users)/[^/]+/workspace/[^/]+/[^/]+/",
```

Of course, anything is possible, but these seem largely safe.

We _could_ introduce a flag like `--skip-default-transforms` but I'd wait until this is actually a real world problem before introducing it.

## Risk: doesn't work in practice (missed or incorrect patterns)

There are unit tests for the patterns which are good. The risk I guess is that these CI providers have different paths than the ones we've specified, and in which case, the behavior will be the same as today, which is that a strip prefix argument must be supplied. Seems ok.

## Risk: isn't backwards compatible -- doesn't properly apply a customers add or strip prefix transformation

The tests in `qlty-coverage/src/publish/planner.rs` seem pretty good. There isn't an end to end test however. We could consider adding that.

# Enhancements

Think this can largely wait, but was curious on your take:

- Denote in debug output that paths are getting auto-transformed (bonus: which transformation was applied)
- Print a deprecation warning if the strip prefix provided matches an established pattern

brynary:
Thanks this is super helpful!

We can release on a weekend to limit any thundering herd of problems if we make a mistake.

One thing we should add is output like a line like this:

`Auto-path fixing: Enabled`

I don't think that has to be dynamic. In other words I think it's fine to render Enabled if we are attempting auto path fixing regardless of whether any paths are adjusted.

(Since path fixing is applied per path, the question of "was auto path fixing applied?" is not binary anyway. We could in theory print the counts but I don't think that's needed yet.)

noahd1:
> (Since path fixing is applied per path, the question of "was auto path fixing applied?" is not binary anyway. We could in theory print the counts but I don't think that's needed yet.)

Depends on perspective I guess -- it's binary if _any_ path fixing was applied. That said, it's certainly simpler to state that Auto-Path fixing is enabled.

brynary:
@noahd1 we need to add a path fix for: `github.com/SOMETHING/SOMETHING` which is a form that I've been seeing lately. (This isn't even a system path on disk it's virtual but we can fix it)

brynary:
> Depends on perspective I guess -- it's binary if any path fixing was applied. That said, it's certainly simpler to state that Auto-Path fixing is enabled.

Yeah, better than just "enabled" would be "Enabled but not applied" vs. "Enabled and applied"

PR #2342, opened 2025-08-22
feat: add configurable action timeout for external operations

brynary:
## Summary

- Adds a configurable `--action-timeout` CLI argument to control timeouts for operations that rely on external resources
- Implements timeout handling for network requests, script executions, tool installations, and other external operations
- Provides user control over timeout duration based on their network conditions and requirements

### CLI Arguments
- Added `--action-timeout` flag to `init`, `install`, `check`, and `fmt` commands
- Accepts duration formats like "5m" (5 minutes), "300s" (300 seconds), "10m" (10 minutes)
- Defaults to 5 minutes if not specified

### Core Implementation
- Created duration parsing utility in `qlty-cli/src/duration.rs` that handles various time formats
- Added `action_timeout` field to Settings struct with 5-minute default
- Applied timeout to:
  - HTTP downloads in `qlty-config/src/http.rs`
  - Tool installations in `qlty-check/src/tool/download.rs`
  - Script-based hooks and invocations in `qlty-check/src/executor.rs` and `qlty-check/src/executor/driver.rs`

### Error Handling
- Timeout errors respect the `--skip-errored-plugins` flag
- Provides clear error messages when operations time out

## Benefits

This feature prevents the CLI from hanging indefinitely when:
- Network resources are slow or unresponsive
- External scripts enter infinite loops
- Tool downloads stall due to connectivity issues

Users can adjust the timeout based on their specific needs:
- Increase for slower network connections
- Decrease for faster feedback in CI/CD environments

## Test Plan

- [ ] Test with default timeout (5 minutes)
- [ ] Test with custom timeout values (e.g., `--action-timeout=30s`, `--action-timeout=10m`)
- [ ] Test timeout triggering with slow network simulation
- [ ] Test that `--skip-errored-plugins` properly handles timeout errors
- [ ] Verify timeout applies to all external operations (downloads, scripts, installations)
- [ ] Test invalid duration formats produce helpful error messages

## Note

There is one test failure in `tests/cmd/check/timeout.toml` that needs investigation. The test expects a timeout to occur but with the new implementation it's not triggering as expected. This may need adjustment to align with the new timeout behavior.

### DIFF
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -75,7 +75,7 @@ impl Executor {
             self.plan.tools(),
             self.plan.jobs,
             self.progress.clone(),
-            Some(self.plan.settings.action_timeout),
+            self.plan.settings.action_timeout,
         );
 
         for installation_result in installation_results {
@@ -106,7 +106,7 @@ impl Executor {
         tools: Vec<(String, Box<dyn Tool>)>,
         jobs: usize,
         progress: Progress,
-        timeout: Option<std::time::Duration>,
+        timeout: std::time::Duration,
     ) -> Vec<(String, Result<()>)> {
         let timer = Instant::now();
         let pool = rayon::ThreadPoolBuilder::new()
@@ -225,7 +225,7 @@ impl Executor {
         name: String,
         tool: Box<dyn Tool>,
         progress: Progress,
-        _timeout: Option<std::time::Duration>,
+        _timeout: std::time::Duration,
     ) -> Result<()> {
         // Timeout is now passed through ToolBuilder to the tools that need it
         let task = progress.task(&name, "Installing...");
--- a/qlty-check/src/settings.rs
+++ b/qlty-check/src/settings.rs
@@ -2,6 +2,8 @@ use qlty_config::config::CheckTrigger;
 use qlty_types::analysis::v1::{Issue, Level};
 use std::{fmt::Formatter, path::PathBuf, time::Duration};
 
+const DEFAULT_ACTION_TIMEOUT_SECS: u64 = 600; // 10 minutes
+
 #[derive(Debug, Clone)]
 pub struct Settings {
     pub root: PathBuf,
@@ -54,7 +56,7 @@ impl Default for Settings {
             skip_errored_plugins: false,
             emit_existing_issues: false,
             auth_token: None,
-            action_timeout: Duration::from_secs(600), // Default 10 minutes
+            action_timeout: Duration::from_secs(DEFAULT_ACTION_TIMEOUT_SECS),
         }
     }
 }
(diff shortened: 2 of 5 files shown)


qlty-check/src/executor.rs
```diff
@@ -71,8 +71,12 @@ impl Executor {
 
     pub fn install(&self) -> Result<Vec<Message>> {
         let mut install_messages = vec![];
-        let installation_results =
-            Self::install_tools(self.plan.tools(), self.plan.jobs, self.progress.clone());
+        let installation_results = Self::install_tools(
+            self.plan.tools(),
+            self.plan.jobs,
+            self.progress.clone(),
+            self.plan.settings.action_timeout,
```
brynary, line 78: This is a message chain smell.

In keeping with  should add the action timeout to the Plan so that the Executor does not need to reach through the Plan to the Settings.

qlty-check/src/executor.rs
```diff
@@ -216,7 +221,13 @@ impl Executor {
         Ok(Results::new(messages, invocations, issues, formatted))
     }
 
-    fn install_tool(name: String, tool: Box<dyn Tool>, progress: Progress) -> Result<()> {
+    fn install_tool(
+        name: String,
+        tool: Box<dyn Tool>,
+        progress: Progress,
+        _timeout: std::time::Duration,
```
brynary, line 228: Can we remove this _timeout arg?

qlty-check/src/executor/driver.rs
```diff
@@ -127,7 +127,8 @@ impl Driver {
         let timer = Instant::now();
         let handle = cmd.start()?;
         let pids = handle.pids();
-        let timeout = plan.driver.timeout;
+        // Use the minimum of action_timeout and driver timeout
+        let timeout = std::cmp::min(plan.settings.action_timeout.as_secs(), plan.driver.timeout);
```
brynary, line 131: The calculation of the effective timeout should be moved from the executor to the planner

qlty-check/src/executor/driver.rs
```diff
@@ -519,12 +520,42 @@ impl Driver {
         let timer = Instant::now();
         let invocation_label = plan.invocation_label();
 
-        let output = cmd.run().with_context(|| {
+        // Use action_timeout from settings
+        let timeout = plan.settings.action_timeout;
+
+        // Start the command
+        let handle = cmd.start().with_context(|| {
```
brynary, line 527: Can we move the timeout handling down into the cmd code? It looks like we already have a CmdWrapper of some sort.

qlty-check/src/settings.rs
```diff
@@ -1,6 +1,8 @@
 use qlty_config::config::CheckTrigger;
 use qlty_types::analysis::v1::{Issue, Level};
-use std::{fmt::Formatter, path::PathBuf};
+use std::{fmt::Formatter, path::PathBuf, time::Duration};
+
+const DEFAULT_ACTION_TIMEOUT_SECS: u64 = 600; // 10 minutes
```
brynary, line 5: Are there any plugins which should get a higher timeout than 10 mins by default?

PR #2352, opened 2025-08-25
feat: add ktlint plugin

marschattha:
https://github.com/qltysh/qlty/issues/2351

### DIFF
--- a/qlty-plugins/plugins/linters/ktlint/plugin.toml
+++ b/qlty-plugins/plugins/linters/ktlint/plugin.toml
@@ -1,0 +1,28 @@
+config_version = "0"
+
+[plugins.definitions.ktlint]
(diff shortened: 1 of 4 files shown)


qlty-plugins/plugins/linters/ktlint/plugin.toml
```diff
@@ -0,0 +1,28 @@
...
+cache = true
+suggested = "targets"
+
+[plugins.definitions.ktlint.drivers.format]
+script = "java -jar ${linter}/ktlint -F ${target}"
+success_codes = [0, 1]
+output = "rewrite"
+batch = true
+driver_type = "formatter"
+suggested = "targets"
```
brynary: I've been moving formatters down from `suggested = "targets"` to `suggested = "config"` lately, and I think we should do the same here.

PR #2354, opened 2025-08-27
fix: ensure consistent output messages regardless of cache state

brynary:
## Summary

This PR fixes a bug where the caching mechanism in `qlty check` was causing inconsistent output messages.

## Problem

When all targets were cached, the output would incorrectly show:
```
No modified files for checks were found on your branch
```

Instead of the expected:
```
✔ No issues
```

This inconsistency was confusing because the same scenario with `--no-cache` would show the correct "✔ No issues" message.

## Solution

The fix tracks the total number of targets (both cached and uncached) throughout the pipeline, ensuring that the cache remains an implementation detail without affecting user-visible output.

### Changes Made

- Added `total_targets_count` field to `Plan`, `Planner`, and `Report` structs to track the total count
- Calculate total targets from all driver planners (includes both cached and uncached targets)
- Updated `Report::targets_count()` to use the tracked total instead of computing from invocations
- Updated test in `sarif.rs` to include the new field

## Testing

- All existing tests pass
- The output message is now consistent regardless of cache state
- Cache functionality remains unchanged (it still works as expected)

## Test Plan

- [ ] Run `qlty check` on a clean branch (should show "✔ No issues")
- [ ] Run `qlty check` again with cache enabled (should still show "✔ No issues", not the "No modified files" message)
- [ ] Run `qlty check --no-cache` (should show "✔ No issues")
- [ ] Verify cache is still working by checking execution times

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
+
+[[package]]
 name = "ascii"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


brynary:
@noahd1 while I think the PR description is a bit off, I think the fix is correct to unify the behavior

brynary:
@claude resolve the merge conflicts. Cargo.lock is a generated file

brynary:
@noahd1 can you please review?

noahd1:
@brynary The code change is minimal, but as someone less familiar with the cache and generally this code, I'd probably look for either a new test to exercise some behavior or a changed test, but I don't see that here. Seems like a good test would be running the same check twice, once without a cache and once with a cache and verifying that the results are the same.

brynary:
@claude can you please try to address Noah's feedback by adding a test which demonstrates the fix that this PR implements?

PR #2378, opened 2025-09-12
feat: improve coverage publish output

brynary:


noahd1:
@brynary I fixed the test on this and added printing of the `JACOCO_SOURCE_PATH` env var to the output

qlty-cli/src/commands/coverage/utils.rs
```diff
@@ -93,6 +93,19 @@ pub fn print_settings(settings: &Settings) {
...
     }
 
+    // Print JACOCO_SOURCE_PATH if defined
+    if let Ok(jacoco_source_path) = std::env::var("JACOCO_SOURCE_PATH") {
+        if !jacoco_source_path.is_empty() {
+            let paths: Vec<&str> = jacoco_source_path.split_whitespace().collect();
+            if !paths.is_empty() {
+                eprintln!("    JACOCO_SOURCE_PATH (from environment):");
+                for path in paths {
+                    eprintln!("      - {}", path);
```
brynary: I would remove the `- ` here. I understand it's trying to print a bulleted list but I think for this case I prioritize clean output

noahd1: :) AI wins here. I _added_ that `-`

PR #2402, opened 2025-09-23
fix: interaction of exported_config_paths and copy_configs_into_tool_install directives

marschattha:
Right now configs files exported from sources using export_config_files directive do not behave correctly with copy_configs_into_tool_install directive used for certain plugins such as eslint 9, this PR fixes that by adding the path of the config to be exported to the config files list during planning phase (important for it to be inserted into the invocations details and to be used in `--config` flag correctly) and copying them into the root from sources first before anything else, such as copying them into tool install directory.

### DIFF
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -310,6 +310,29 @@ impl Executor {
 
         let mut loaded_config_files = vec![];
 
+        // load exported config paths before anything else
+        for config_file in &exported_config_paths {
+            if self.plan.workspace.root != self.plan.staging_area.destination_directory {
+                // for formatters
+                let loaded_config_file = load_config_file_from_source(
+                    config_file,
+                    &self.plan.staging_area.destination_directory,
+                )?;
+
+                if !loaded_config_file.is_empty() {
+                    loaded_config_files.push(loaded_config_file);
+                }
+            }
+
+            // for linters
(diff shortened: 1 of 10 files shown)


marschattha:
> I reviewed. It's mostly moving code around, not new code, so I think it's fine.
> 
> I did notice that the function `stage_workspace_entries` is quite large. Maybe rust code in part has a way of being verbose. One area of improvement to that function is how it deals with formatters vs linters -- we might want to either handle those cases completely differently OR treat them the same, but we're in the middle.

I see what you mean, I did try and take a crack at it but it quickly ballooned up the scope of this PR, as there are a lot of things we can refactor and cleanup in there, maybe that can be done in a followup PR, that should not have any behavioural change?

noahd1:
> > I reviewed. It's mostly moving code around, not new code, so I think it's fine.
> > I did notice that the function `stage_workspace_entries` is quite large. Maybe rust code in part has a way of being verbose. One area of improvement to that function is how it deals with formatters vs linters -- we might want to either handle those cases completely differently OR treat them the same, but we're in the middle.
> 
> I see what you mean, I did try and take a crack at it but it quickly ballooned up the scope of this PR, as there are a lot of things we can refactor and cleanup in there, maybe that can be done in a followup PR, that should not have any behavioural change?

I think it's reasonable to defer to a follow up PR

PR #2404, opened 2025-09-24
Fix: add the glob for the gradle metadata.xml

### LINKED ISSUE #2403: Bug: osv-scanner and gradle needs a different lockfile
### What happened?

I think this is the right way to support the version lockfile on gradle: https://github.com/google/osv-scanner/issues/915

I have noticed that the plugin.toml for osv-scanner does not include that path.

Should it include: `"**/gradle/verification-metadata.xml"` in the globs?

It's generated by this: https://docs.gradle.org/current/userguide/dependency_verification.html

Happy to make a PR, but I'm not sure yet if that's the right way to do it.

osv-scanner should run against a gradle project.

Yes, consistently

### CLI Version

qlty 0.574.0 macos-arm64 (450dd70 2025-09-12)

### Relevant log output

```shell
❯ qlty check --all --filter osv-scanner --verbose --no-cache
     [0/1] 🤔  Planning...  0.05s
     [1/1] 🔍  Analyzing all targets...
 JOBS: 0

Plugin  Result  Targets  Time  Debug File

✔ No issues


The project has a build.gradle.kts file, and I've created a metadata verification:

❯ ls gradle/
 libs.versions.toml   verification-metadata.xml   wrapper


But the qlty auto-detection bits don't work.
```

dkowis:
Resolves #2403

Is there documentation or other things I need to add?

### DIFF
--- a/qlty-plugins/plugins/linters/osv-scanner/plugin.toml
+++ b/qlty-plugins/plugins/linters/osv-scanner/plugin.toml
@@ -15,6 +15,7 @@ globs = [
   "**/go.mod",
 
   # Gradle (Android/Java/Kotlin)
+  "**/gradle/verification-metadata.xml",
   "**/buildscript-gradle.lockfile",
   "**/gradle.lockfile",
   "**/pom.xml",


brynary:
@dkowis thanks for this contribution. I believe this should be all that is needed for osv-scanner to scan those files. 

Have you tried this locally? You can use the same block of TOML in your project's qlty.toml to override the plugin definition.

dkowis:
I did. I built the binary and used it directly, since I didn't know I could override that value in the plugin toml. It did pick up the thing and give me dependency results :D

image

PR #2407, opened 2025-09-25
refactor: qlty check executor staging_workspace_entries

marschattha:
Followup from https://github.com/qltysh/qlty/pull/2402

Tried to clean up the `staging_workspace_entries` function without making any behavioural/logical change.

Also added some tests, although they are not so straight forward due to the sheer amount of mocking required, I think they should be a step in the right direction.


marschattha:
Closing in favour of https://github.com/qltysh/qlty/pull/2416

PR #2415, opened 2025-09-26
feat: Ensure read important env vars for BuildKite for coverage publishing

noahd1:
Most crucially, we were missing the pull number.


dzunk:
FYI @noahd1 this is a breaking change for us as the `BUILDKITE` env var was previously not required. Since we're running `qlty` in a Docker container inside Buildkite, we explicitly pass all the necessary env vars to the container in our build configuration.

PR #2416, opened 2025-09-26
refactor: config discovery logic from executor to planner

marschattha:


qlty-check/src/planner/config_files.rs
```diff
@@ -15,6 +16,27 @@ pub struct PluginConfigFile {
...
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub enum ConfigOperationType {
+    /// Copy config from repository to staging area
+    CopyToStagingArea,
+    /// Copy config from repository to workspace root
+    CopyToWorkspaceRoot,
+    /// Load config from qlty directory to destination
+    LoadFromQltyDir,
```
brynary: @marschattha why does this one start with `Load` but the others start with `Copy`?

qlty-check/src/planner/config_files.rs
```diff
@@ -15,6 +16,27 @@ pub struct PluginConfigFile {
...
+    /// Copy config into tool install directory
+    CopyToToolInstall,
+    /// Download and place fetched file
+    FetchFile,
```
brynary: When the operation type is `FetchFile`, then where is the URL to fetch from stored in the plan?

qlty-check/src/planner/config_files.rs
```diff
@@ -15,6 +16,27 @@ pub struct PluginConfigFile {
...
+    pub destination_path: PathBuf,
+    pub operation_type: ConfigOperationType,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub enum ConfigOperationType {
+    /// Copy config from repository to staging area
+    CopyToStagingArea,
+    /// Copy config from repository to workspace root
+    CopyToWorkspaceRoot,
```
brynary: Is there ever a case where we want to `CopyToStagingArea` but we don't want to `CopyToWorkspaceRoot` or vice versa? In other words, are these two operation types always going to be used together?

qlty-check/src/planner/config_files.rs
```diff
@@ -148,3 +176,568 @@ pub fn plugin_configs(planner: &Planner) -> Result<HashMap<String, Vec<PluginCon
...
+        });
+    }
+
+    // Add operations for plugin fetches
+    for active_plugin in &plugins {
+        if !active_plugin.plugin.fetch.is_empty() {
+            for fetch in &active_plugin.plugin.fetch {
+                // Plugin fetch downloads to both workspace root and staging area
+                operations.push(ConfigStagingOperation {
+                    source_path: PathBuf::from(&fetch.path), // URL/source identifier
```
marschattha: FetchFile URL is stored as `source_path`

qlty-check/src/planner/config_files.rs
marschattha: https://github.com/qltysh/qlty/pull/2416/files#r2383172699

qlty-check/src/planner/config_files.rs
marschattha: Yes, if there are no formatters involved, there can be only `CopyToWorkspaceRoot` directives, if there are no linters involved (qlty fmt) there can be only `CopyToStagingArea`, if both are involved then both.
Also differs from plugin to plugin.

qlty-check/src/planner/config_files.rs
marschattha: > The verbs reflect the helper functions we call in the executor: for configs shipped inside the qlty directory we invoke load_config_file_from_qlty_dir, so the operation got named LoadFromQltyDir

Changing it to `CopyFromQltyDir` for consistency

qlty-check/src/planner/plan.rs
```diff
@@ -27,6 +28,7 @@ pub struct Plan {
     pub transformers: Vec<Box<dyn IssueTransformer>>,
     pub invocations: Vec<InvocationPlan>,
     pub fail_level: Option<Level>,
+    pub config_staging_operations: Vec<ConfigOperation>,
```
brynary: Would this field be better named `config_operations` or is the "staging" part important?

qlty-check/src/executor.rs
```diff
@@ -245,223 +246,169 @@ impl Executor {
...
-        }
-
-        let exported_config_paths = self
-            .plan
-            .invocations
-            .iter()
-            .flat_map(|invocation| invocation.plugin.exported_config_paths.clone())
-            .unique()
-            .collect_vec();
+        let loaded_config_files = self.execute_config_staging_operations()?;
```
brynary: Would `.stage_configs()` convey all of the meaning from `.execute_config_staging_operations()` but shorter?

qlty-check/src/executor.rs
```diff
@@ -245,223 +246,169 @@ impl Executor {
...
 
         debug!(
-            "Walker found {} config and affects cache files in {:.2}s",
-            repository_config_files.len(),
+            "Staged {} config files in {:.2}s",
+            loaded_config_files.len(),
```
brynary: We seem to have a collision where sometimes we refer to _"loaded"_ configs and other times we refer to _"staged"_ configs. Would it be appropriate to use "staged" everywhere?

qlty-check/src/executor.rs
```diff
@@ -245,223 +246,169 @@ impl Executor {
...
-                let loaded_config_file = load_config_file_from_qlty_dir(
-                    config_file,
-                    &self.plan.workspace,
-                    &self.plan.staging_area.destination_directory,
-                )?;
-
-                if !loaded_config_file.is_empty() {
-                    loaded_config_files.push(loaded_config_file);
-                }
+    fn execute_single_config_operation(
```
brynary: Starting here we have the code to execute the ConfigOperations, which splits based on the config source (download or file). This seems this code would be better off not being in `executor.rs` which is fairly big.

Config file are an important concept. In the planner, we have `planner/config_files.rs` which makes sense to me. I would recommend we move the code for executing config operations to `executor/config_files.rs`.

This creates a _symmetry_ between the planner and executor.

qlty-check/src/executor.rs
```diff
@@ -774,20 +721,28 @@ fn run_invocation(
...
-            if let Ok(dent) = result {
-                dents.lock().unwrap().push(dent);
-            }
-            WalkState::Continue
-        })
-    });
+#[cfg(unix)]
+fn create_symlink(from: &Path, to: &Path) -> std::io::Result<()> {
+    std::os::unix::fs::symlink(from, to)
+}
```
brynary: It feels like the symlink stuff and `ensure_parent_exists` is ripe to be isolated into a file like `fs.rs` somewhere where it is reusable and out of the way

qlty-check/src/planner/config_files.rs
```diff
@@ -148,3 +175,614 @@ pub fn plugin_configs(planner: &Planner) -> Result<HashMap<String, Vec<PluginCon
...
+
+pub fn compute_config_staging_operations(planner: &Planner) -> Result<Vec<ConfigOperation>> {
+    let plugins = enabled_plugins(planner)?;
+    let all_config_paths = collect_all_config_paths(&plugins);
+
+    let mut operations = Vec::new();
+    operations.extend(repository_config_operations(planner, &all_config_paths)?);
+    operations.extend(exported_config_operations(planner, &plugins)?);
+    operations.extend(qlty_config_operations(planner, &plugins)?);
+    operations.extend(fetch_operations(planner, &plugins));
```
brynary, line 191: This approach of having 4 different functions which each produce operations seems OK for now, but I would watch it carefully to make sure it does not get out of hand.

I think the most important thing is to add code comments for these four functions which explain why the exist, to help keep it clear what each of the four functions is doing.

qlty-check/src/planner/config_files.rs
```diff
@@ -148,3 +175,614 @@ pub fn plugin_configs(planner: &Planner) -> Result<HashMap<String, Vec<PluginCon
...
+        }
+    }
+
+    operations
+}
+
+pub fn compute_tool_install_config_operations(
+    planner: &Planner,
+    invocations: &[crate::planner::InvocationPlan],
+) -> Result<Vec<ConfigOperation>> {
```
brynary: I was expecting that all of the config operations would be exposed through the `compute_config_staging_operations` which is public.

However, this function is also public, which suggests something else in the codebase is calling it directly.

This seems to break the abstraction that we are providing one function (`compute_config_staging_operations`) which produces _all_ of the config operations.

Why is `compute_tool_install_config_operations` exposed seprately?

marschattha:
The new implementation is cleaner, and so far testing hasn't surfaced any issues. Given the code’s criticality, will proceed with the release and monitor closely.

PR #2426, opened 2025-10-02
feat: Add install only flag to check

marschattha:


qlty-cli/src/commands/check.rs
```diff
@@ -123,6 +123,10 @@ pub struct Check {
 
     /// Files to analyze
     pub paths: Vec<PathBuf>,
+
+    /// Install tools only, do not run checks
+    #[arg(long)]
+    install_only: bool,
```
brynary, line 129: What do you think of the`--install` vs `--install-only`?

qlty-check/src/executor.rs
```diff
@@ -59,6 +59,10 @@ impl Executor {
 
     pub fn install_and_invoke(&self) -> Result<Results> {
         let install_messages = self.install()?;
+        if self.plan.settings.install_only {
```
brynary: I feel like rather than having the Executor reach _through_ the Plan up to the Settings that the Planner should be responsible for determining what to do an encoding that information into the Plan.

The Planner reads Settings (and other things like files) and turns it a Plan. The Executor runs the Plan.

Do we have other cases today where the executor reads from `self.settings.plan`?

qlty-cli/src/commands/check.rs
marschattha: `--install` sounds like you are giving permission to install the tools but there is ambiguity wether its actually going to run checks, `--install-only` implies it is only going to install and nothing else which is closer to the behaviour of the flag.

qlty-check/src/executor.rs
marschattha: Yes there are a total of 6 other instances of `self.plan.settings` in the executor right now.

brynary: OK, what I would suggest is that we take a look at what would be needed to address those.

The reason that I would do that before this is because this increases the number of `self.plan.settings` references when I think we need to drive that down to zero

brynary (approved):

PR #2438, opened 2025-10-06
feat: add warning for unexpected config

marschattha:
Add warnings for unexpected config, there were a few in out own plugin definitions as well.


brynary:
We had the same idea, this was a PR that I was experimenting with earlier:

https://github.com/qltysh/qlty/pull/2381

Should this PR replace that PR? If so we can close that one and focus on this one

marschattha:
Yes this PR has the tests and current plugin definitions cleaned up a bit better.

brynary (approved):

PR #2446, opened 2025-10-14
fix: warn on excluding config_file

marschattha:


qlty-cli/tests/cmd/fmt/config_copying.in/.qlty/qlty.toml
```diff
@@ -5,12 +5,12 @@ file_types = ["shell"]
 
 [plugins.definitions.sample.drivers.format]
 prepare_script = "mkdir ${linter} && echo dir %2 > ${linter}/ls.cmd || echo dir %2 > ${linter}/ls.cmd"
-script = "ls -l ${config_file}"
-config_files = ["config.json"]
+script = "ls -l \"deep/nested/config.json\" || ls -l \"deep\\nested\\config.json\""
```
marschattha, line 8: Changed this test so that it checks 
1. That the config file is copied over to the staging directory using relative path, as `${config_file}` is now replace with an absolute path to the config file in the workspace directory.
2. Ensures config_files from `[[plugin]]` blocked are read and correctly accounted for.

PR #2451, opened 2025-10-16
chore: Generate id for analysis

noahd1:
Generates UUIDv7 id for an analysis.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,7 +141,7 @@ tree-sitter-typescript = "0.21.2"
 trycmd = "0.15.8"
 ureq = { version = "2.12.1", features = ["json", "proxy-from-env"] }
 url = "2.5.7"
-uuid = { version = "1.14.0", features = ["v4"] }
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -141,7 +141,7 @@ tree-sitter-typescript = "0.21.2"
 trycmd = "0.15.8"
 ureq = { version = "2.12.1", features = ["json", "proxy-from-env"] }
 url = "2.5.7"
-uuid = { version = "1.14.0", features = ["v4"] }
+uuid = { version = "1.14.0", features = ["v4", "v7"] }
```
brynary, line 144: Is there a UUIDv4 in here somewhere we can switch to v7?

noahd1: Looks like there's 3 references:
- buildId
- telemetry message id
- "original state" AppState part of auth_flow

Are they completely swappable? Either way seems appropriate for a separate PR.

PR #2465, opened 2025-10-20
fix: plugins enable behaviour interaction with suggested directive

marschattha:
Follow up for https://github.com/qltysh/qlty/pull/2334, fixes plugins enable behaviour to bypass suggested directive.

### DIFF
--- a/qlty-cli/src/initializer/scanner.rs
+++ b/qlty-cli/src/initializer/scanner.rs
@@ -7,7 +7,7 @@ mod package_json;
 use super::{Renderer, Settings, SourceSpec};
 use anyhow::Result;
 use driver_candidate::DriverCandidate;
-use driver_initializer::{ConfigDriver, DriverInitializer, TargetDriver};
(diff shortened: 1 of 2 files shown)


brynary:
@marschattha can the Rust changes be isolated from the other changes into two PRs?

marschattha:
@brynary, if we can merge this first then we can update the workflow file and snapshots in a followup PR.

brynary:
@marschattha I feel like the Scanner is at risk of getting overloaded but I don't have any clear ideas for small improvements that we could make to it.

The name "Scanner" implies that it is focused on walking folders on disk, but the code in here does a lot more than that. It's more like an initialization planner, perhaps.

I'm comfortable enough to go with this PR, but wanted to check if you had any feelings or suggestions on this

marschattha:
It is definitely getting overloaded with all the responsibilities, might be a good chance to refactor.

But since this PR fixes a regression and blocks the plugins tests, I feel like we should merge this fix and circle back to refactoring as that might take more time.

PR #2489, opened 2025-10-28
fix: repository config file handling

marschattha:
After investigating https://github.com/qltysh/qlty/issues/2483, it seems there was a regression introduced in https://github.com/qltysh/qlty/pull/2416 which may cause deletion of plugin config files present in the project directory.

Added the patch to update the planner logic as well as the executor logic as a secondary safeguard, and refined the planner tests as well.

### DIFF
--- a/qlty-check/src/planner/config_files.rs
+++ b/qlty-check/src/planner/config_files.rs
@@ -677,20 +677,7 @@ mod tests {
             "Should have two operations for exported configs"
         );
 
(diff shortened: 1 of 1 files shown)


brynary:
@marschattha can you please add to the Pr description with what is happening, and why, and is this a regression, or a behavior which has been like this for awhile?

marschattha:
https://github.com/qltysh/qlty/pull/2490

PR #2509, opened 2025-11-14
chore: opt-in feature to merge file coverage reports to make zip payload smaller

noahd1:
For very large coverage reports with duplicate file paths, merging the paths on the client before upload is overall more efficient than incurring the cost of server side merging.

This is a first step towards client side merging: if a flag `--merge` is passed, file coverages are merged based on the same rules that we currently use in the cloud. The rules mostly cover edge case scenarios when coverage hits arrays disagree in incompatible ways. In these cases, there's no "right" answer, so we aim to just be consistent.

If no flag is passed the results should be the same as they are today.

Rules are:

- when coverage hits arrays differ in size for the same file (aka, line counts differ), return a hits array the length of the smallest hits array
- when merging non-negative hit counts, simply sum them (compatible)
- when merging negative and non-negative (incompatible), report the sum as -1

More to discuss will hit up separately.

### DIFF
--- a/qlty-coverage/src/publish/metrics.rs
+++ b/qlty-coverage/src/publish/metrics.rs
@@ -1,6 +1,5 @@
 use qlty_types::tests::v1::FileCoverage;
 use serde::Serialize;
-use std::collections::HashMap;
 
 #[derive(Debug, Clone, Default, Serialize)]
 pub struct CoverageMetrics {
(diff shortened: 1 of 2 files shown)


noahd1:
Going to open a separate PR that will make discussion of key decisions easier.

PR #2510, opened 2025-11-15
Extract and change merging logic when calculating total coverage

noahd1:
Originally I tried to create a behavior-neutral PR which introduced an new opt-in behavior: merging coverage reports for the same file on the client. However, I discovered that there was some pre-existing coverage merging behavior that I wasn't aware of before raising the question of how and/or if to do this in a behavior neutral way.

I've opened this PR to change the pre-existing logic for merging coverage reports which is used solely for calculating total coverage numbers and is not behavior neutral. However, I believe this change will cause the total coverage numbers calculated on the server to match those reported in the client in more cases.

The specific decisions regarding merging that are tricky is when you have incompatible hits arrays -- differing lengths for example, or when one hits array reports coverable and another reports uncovered/covered. That's specifically what's in question here and this PR matches the decisions we make when performing server side merging.

- Should client side merging logic match what we perform on the server?
- Should the merging logic used to calculate total coverage numbers match the client side (not yet performed) merging logic which will return file coverages unique by path?

### DIFF
--- a/qlty-coverage/src/lib.rs
+++ b/qlty-coverage/src/lib.rs
@@ -3,6 +3,7 @@ mod env;
 pub mod export;
 pub mod formats;
 pub mod git;
+mod merging;
 pub mod parser;
 pub mod print;
 pub mod publish;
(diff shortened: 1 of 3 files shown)


brynary:
Can I suggest we break this into two PRs: A vanilla refactor to extract the merging logic and adds tests with no changes to the logic, followed by a PR with the proposed behavioral changes?

PR #2515, opened 2025-11-20
fix: block auto-downgrades

brynary:


noahd1:
I tested:

- attempting to downgrade which was disallowed
- attempting to upgrade which was allowed

noahd1 (approved):
Tested and it appropriately blocked downgrades.

PR #2516, opened 2025-11-20
move temp data to $HOME/.qlty/tmp

brynary:
Instead of system `/tmp` at root


noahd1:
Claude seems to have hullicinated a connection between Library::create and telemetry and I believe that comment is inaccurate.

noahd1 (approved):
Tested locally worked.

PR #2517, opened 2025-11-20
feat: require HTTPS for all outgoing HTTP requests

brynary:
`QLTY_INSECURE_ALLOW_HTTP` can be set to `true` for development / testing needs to override this behavior

### DIFF
--- a/qlty-coverage/src/ci/bitrise.rs
+++ b/qlty-coverage/src/ci/bitrise.rs
@@ -1,7 +1,5 @@
-use crate::{
-    ci::CI,
-    env::{EnvSource, SystemEnv},
-};
+use crate::ci::CI;
+use qlty_config::env::{EnvSource, SystemEnv};
 
 #[derive(Debug)]
 pub struct Bitrise {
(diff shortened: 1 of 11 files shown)


qlty-cli/tests/cmd/check/exported_config_files.in/results.txt
```diff
@@ -1 +0,0 @@
-sample.sh:2:1: OVERRIDE valid line
```
brynary, line 1: Need to look at why this was deleted

qlty-config/src/http.rs
```diff
@@ -13,16 +16,46 @@ static AGENT: Lazy<ureq::Agent> = Lazy::new(|| {
...
+
+    if let Some(host) = url.host_str() {
+        if host == "localhost"
+            || host == "::1"
+            || host.starts_with("127.")
+            || host.starts_with("[::1]")
+        {
+            return Ok(());
+        }
+    }
```
brynary: I want to look at removing this exception for localhost / loopback hosts

PR #2613, opened 2025-12-25
Detect Jenkins environment variables

noahd1:


brynary (approved):
Approving, but please address robot feedback (either by confirming/editing or refuting) before merge

PR #2691, opened 2026-02-12
feat: rename NullTool to ShellTool and add install_script support

noahd1:
## Summary

- Rename `NullTool` to `ShellTool` and add `install_script` field to `PluginDef`
- When `install_script` is set, `ShellTool` creates the tool directory and runs the script via `sh` during tool setup; when unset, preserves existing no-op behavior
- Install script paths from source `plugin.toml` files are resolved to absolute paths during source parsing; relative paths in inline `qlty.toml` definitions are resolved against the project root at install time

## Test plan

- [x] Unit tests for ShellTool: setup with install script, no-op without script, failing script, idempotent setup
- [x] Integration test (`install_script`) demonstrating a plugin that uses an install script to create a checker used by the driver
- [x] `cargo check` passes
- [x] All existing unit tests pass


qlty-config/src/config/plugin.rs
```diff
@@ -411,6 +411,9 @@ pub struct PluginDef {
 
     #[serde(default = "default_tab_column_width")]
     pub tab_column_width: usize,
+
+    #[serde(default)]
+    pub install_script: Option<String>,
```
brynary, line 416: Since this is on `PluginDef`, I think it implies that any plugin (including non-ShellTool-based) can define an `install_script` -- which probably makes sense as a useful hook but we would need to define the semantics / ordering behavior.

To make sure we don't paint ourselves into a corner, we should probably map out all the hooks we intend to have and how they fit together

qlty-cli/tests/cmd/check/install_script.stdout
```diff
@@ -0,0 +1,7 @@
+ JOBS: 2
+
+Plugin   Result   Targets   Time   Debug File
+greeter  Success  1 target  [..]s  .qlty/out/invoke-[..].yaml
+greeter  Success  1 target  [..]s  .qlty/out/invoke-[..].yaml
```
brynary, line 5: For the purposes of integration testing it would be nice if we can strength the tests...

Overall, we could use more visibility into the install process. This just shows data about what happens post-install during runtime. I think this is a pre-existing deficiency.

The debug files (invocation files) are the gold standard, so ideally we would write out a debug file for the `install_script` and be able to assert against it (or snapshot it)

qlty-cli/tests/cmd/check/install_script.in/install.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+echo '#!/bin/sh' > greet.sh
+echo 'grep -q "hello" "$1" && exit 0 || exit 1' >> greet.sh
+chmod +x greet.sh
```
brynary, line 4: Nice!

qlty-cli/tests/cmd/check/install_script.in/.qlty/qlty.toml
```diff
@@ -0,0 +1,14 @@
+config_version = "0"
+
+[plugins.definitions.greeter]
+file_types = ["shell"]
```
brynary, line 4: This works because `shell` will cause it to self-lint but is a bit obscure. Might warrant a comment.

qlty-cli/tests/cmd/check/install_script.in/.qlty/qlty.toml
```diff
@@ -0,0 +1,14 @@
...
+install_script = "install.sh"
+
+[plugins.definitions.greeter.drivers.lint]
+script = "sh ${linter}/greet.sh ${target}"
+success_codes = [0]
+output = "pass_fail"
+
+[[plugin]]
+name = "greeter"
+version = "1.0.0"
```
brynary, line 14: I think `version` is optional but we should confirm. Certainly it's not doing anything here.

qlty-check/src/tool/shell_tool.rs
```diff
@@ -0,0 +1,178 @@
...
+            self.install(task)?;
+        }
+        Ok(())
+    }
+
+    fn install(&self, _task: &ProgressTask) -> Result<()> {
+        if let Some(ref script_path) = self.plugin.install_script {
+            let script_path =
+                std::fs::canonicalize(script_path).unwrap_or_else(|_| PathBuf::from(script_path));
+            self.run_command(duct::cmd!("sh", script_path))?;
```
brynary, line 75: If this logic is lifted from elsewhere it would make want to think about extracting duplication to avoid subtle bugs later.

qlty-config/src/config/plugin.rs
brynary: Alternatively we could block `install_script` for any non-ShellTool on the basis that our regular installation _is_ the one true install (e.g. `npm install` and force those plugins to use a `before_install_script` or `after_install_script` as hooks to be clear what order things run in). That may be less prone to plugin author errors.

PR #2693, opened 2026-02-12
fix: auto-repair broken cache symlinks in .qlty directory

noahd1:
Noah's note: The impetus for this was that I deleted my ~/.qlty/cache directory and qlty check started to error when it had to check the cache. I'd assumed this was safe, with the thinking being that if a cache is not there, it will be simply re-created. However, that wasn't the case -- the cache works via symlinks, and if the cache is removed, the symlinks to that cache are broken. This causes the CLI to error with no user facing output.

## Summary

- Fix `try_symlink_if_missing()` to detect and repair broken symlinks using `fs::symlink_metadata()` instead of `Path::exists()` which follows symlinks
- Remove broken symlinks before recreating them, with a `warn!` log message for visibility
- Properly propagate errors with `.with_context()` instead of silently swallowing them

## Context

When `~/.qlty/cache/repos/{hash}/` is deleted externally (e.g. manual cleanup, disk tools), the symlinks in `.qlty/` (`out`, `logs`, `results`, `plugin_cachedir`) become broken. The previous code used `Path::exists()` which follows symlinks — a broken symlink returned `false`, the code tried to create a new symlink, but failed because the broken symlink file still physically existed. The error was silently logged, and downstream writes to `.qlty/out/` failed with "No such file or directory".

## Test plan

- [x] `cargo check -p qlty-config` passes with no warnings
- [x] `cargo test -p qlty-config` — all 100 tests pass including 3 new unit tests:
  - `broken_symlink_is_repaired`: creates symlink → deletes target → calls method → verifies new symlink is valid
  - `valid_symlink_is_left_alone`: creates valid symlink → calls method → verifies unchanged
  - `missing_symlink_is_created`: no existing symlink → calls method → verifies created
- [ ] Manual verification: create broken symlinks in `.qlty/`, run `qlty check`, confirm self-healing

### DIFF
--- a/qlty-config/src/library.rs
+++ b/qlty-config/src/library.rs
@@ -1,11 +1,11 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
 #[cfg(unix)]
 use std::os::unix::fs::PermissionsExt;
 use std::{
     env, fs,
     path::{Path, PathBuf},
 };
-use tracing::error;
+use tracing::warn;
 use walkdir::WalkDir;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


noahd1:
I can't replicate this. Going to close for now.

PR #2710, opened 2026-03-03
fix(check): prevent deletion of pre-existing config files during cleanup

marschattha:
## Summary
- `load_config_file_from` previously returned the destination path even when the file already existed in the workspace, causing `cleanup_config_files` to delete the user's original config file
- Changed return type from `Result<String>` to `Result<Option<String>>` so only files actually created (symlinked) by the executor are tracked for cleanup
- Added `error!` log when a source config file doesn't exist

## Test plan
- [x] Existing unit tests updated and passing (`test_load_config_file_already_exists`, `test_load_config_file_nonexistent`)
- [x] All tests pass
- [x] Typecheck, format, and lint all pass


noahd1:
@marschattha I would use warn! instead of error! for the case where the config file already exists at that location

PR #2711, opened 2026-03-03
fix(check): skip exported config for formatters when workspace file exists

marschattha:
## Summary
- For formatters, the exported config was always symlinked into the staging area (since it starts empty), ignoring any pre-existing file in the workspace root
- This was inconsistent with linter behavior, where a pre-existing workspace file takes precedence
- Now checks if the file exists in the workspace root before symlinking the exported config, so the workspace file wins for both linters and formatters

## Test plan
- [x] Typecheck, format, and lint all pass
- [x] Tests pass


noahd1 (commented):
Not saying this is an issue per se .. but just wanting to understand:

- if you previously were using formatting with a shared config, the shared config would win.
- now, your local config file will win, and formatting may look different for you than previously
- how does a customer know which config file "won" when these situations occur? Do we save it to any of our tables / output?

marschattha:
> * if you previously were using formatting with a shared config, the shared config would win.

Correct

> * now, your local config file will win, and formatting may look different for you than previously

Correct, if there is someone who has a shared config and repo based config that is different it can look different based on said config, and there is also the situation where a plugin has both linters and formatters, prior to this PR, formatter would be using the exported config and linter would be using the repo level config.

> * how does a customer know which config file "won" when these situations occur? Do we save it to any of our tables / output?

There is a lack of visibility there I believe since we added supported for a number of different options but haven't made it clear in one place, the debug logs can add some light to it I believe as it logs when it symlinks/copies files.

noahd1:
> > * how does a customer know which config file "won" when these situations occur? Do we save it to any of our tables / output?
> 
> There is a lack of visibility there I believe since we added supported for a number of different options but haven't made it clear in one place, the debug logs can add some light to it I believe as it logs when it symlinks/copies files.

So for some unknown subset of users who have both local configs and shared configs, all their PRs may start failing with formatting errors when a new config takes precedence where it hadn't previously. I think we may need to hold off on this change for the moment until we are wide eyed about the potential impact, even if it's the right change.

brynary (commented):
Can we please start with a test that would fail without this fix?

PR #2732, opened 2026-03-17
fix(check): support literal targets in formatter rewrite diff

marschattha:
## Summary
- When a formatter driver uses a literal target (e.g., `./...` for Go tools), the rewrite diff logic could not resolve actual file paths for comparison
- Falls back to `workspace_entries` when the target type is not `File` in both `handle_output_rewrite` (for `qlty fmt`) and `create_file_result_for_autofmts` (for `qlty check`)

## Test plan
- [x] Verified with golangci-lint v2 format driver using `./...` literal target
- [x] `qlty check` now correctly reports `LEVEL_FMT` issues for literal-target formatters
- [x] `qlty fmt` now correctly rewrites files for literal-target formatters
- [x] All existing tests pass

### DIFF
--- a/qlty-check/src/executor/invocation_result.rs
+++ b/qlty-check/src/executor/invocation_result.rs
@@ -290,16 +290,16 @@ impl InvocationResult {
     fn handle_output_rewrite(&mut self) -> Result<()> {
         let mut formatted = vec![];
 
-        let paths_to_check: Vec<String> =
-            if self.plan.driver.target.target_type == TargetType::Literal {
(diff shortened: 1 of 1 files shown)


marschattha:
Unfortunately there are no unit tests for this (not so straightforward at this point)
I split the changes into smaller PRs by concept boundaries for easier review, the real test for this one will be the one in https://github.com/qltysh/qlty/pull/2733

We can have a dedicated integration test but it would just be redundant and slow

PR #2754, opened 2026-04-14
feat(plugins): add system install mode for Node and PHP package tools

marschattha:
## Summary

Adds `system = true` plugin config option that installs Node/npm and PHP/Composer package tools into the repository workspace instead of qlty's isolated cache directory. This allows tools like `phpstan` to access the project's real autoload graph and dependency tree.

**User-facing config:**
```toml
[[plugin]]
name = "phpstan"
version = "2.1.0"
system = true

[[plugin]]
name = "eslint"
version = "9.7.0"
system = true
prefix = "frontend"
```

## What's covered

- `system` field on `EnabledPlugin` and `PluginDef` with validation (rejects unsupported runtimes, rejects `system + package_filters`)
- `system` field merging in `merge_enabled_plugins()`
- `cache_directory()` method on `Tool` trait to split infrastructure files (donefile, lockfile, logs) from the install directory
- `system_install_directory()` helper: resolves to `parent(package_file)` or `{workspace_root}/{prefix}`
- `directory()` override on `NodePackage` and `PhpPackage` for system mode
- `package_install()` with system-specific flags (`--no-save --package-lock=false` for npm; `--dev --with-all-dependencies --ignore-platform-reqs` for Composer)
- `package_file_install()` runs the package manager's install in the workspace to ensure project deps are present (`npm install` / `composer install`)
- `extra_env_paths()` / `extra_env_vars()` adjusted for system mode (PHP returns `vendor/bin`, drops Composer env overrides)
- Config-copy guard: `copy_configs_into_tool_install` disabled in system mode in both `executor.rs` and `invocation_script.rs`
- `plugin.system` and `tool.fingerprint()` added to issue cache key for proper invalidation
- `PhpPackage::update_hash()` removed — uses default trait impl (was only hashing `self.name()`)
- `Composer::phar_path()` extracted to reduce duplication
- Donefile-based installation detection via fingerprint invalidation (package file contents included in hash)

## What's NOT covered

- Python/pip, Ruby/gem, Go package tools (explicitly out of scope)
- Runtime binaries (node, php) still install into cache
- Binary/download tools unchanged
- No `is_installed()` override — uses default donefile check; if `node_modules`/`vendor` is deleted after install, user must clear cache to force reinstall
- No version validation for system + package_file (the package file dictates the resolved version)

## Test plan

- [x] `cargo check` passes
- [x] `cargo test` — all 966 tests pass
- [x] `qlty fmt` — clean
- [x] `qlty check --level=low --fix` — clean
- [ ] Manual: eslint with `system = true` in a repo root
- [ ] Manual: eslint with `system = true` and `prefix = "frontend"`
- [ ] Manual: phpstan with `system = true` using real project autoload

### DIFF
--- a/qlty-check/src/cache.rs
+++ b/qlty-check/src/cache.rs
@@ -166,6 +166,8 @@ impl InvocationCacheKey {
             digest.add("plugin.prefix", prefix);
         }
 
+        digest.add("plugin.system", &self.plugin.system.to_string());
+
         let driver = self.plugin.drivers.get(&self.driver_name).unwrap();
 
         digest.add("plugin.driver.script", &driver.script);
@@ -245,6 +247,7 @@ impl InvocationCacheKey {
 
         digest.add("qlty_version", &self.qlty_version);
         digest.add("tool", &self.tool.directory());
+        digest.add("tool_fingerprint", &self.tool.fingerprint());
         digest.add("driver_name", &self.driver_name);
 
         for config in self.configs.clone().iter().sorted() {
--- a/qlty-check/src/executor.rs
+++ b/qlty-check/src/executor.rs
@@ -417,7 +417,7 @@ impl Executor {
         loaded_config_files: &mut Vec<String>,
     ) -> Result<()> {
         for invocation in self.plan.invocations.iter() {
-            if invocation.driver.copy_configs_into_tool_install {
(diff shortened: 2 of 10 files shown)


marschattha:
Superseded by a redesigned approach using install_dir = project (enum) instead of system = true (bool). New PR coming.

PR #2759, opened 2026-04-16
feat: add VB.NET maintainability support

davehenton:
## Summary

Adds first-class VB.NET maintainability support so `.vb` files are discovered by the built-in smells and metrics pipeline, emit `LANGUAGE_VBDOTNET`, and preserve VB.NET's case-insensitive member semantics (`Me.Value` / `me.value` unify, `Sub New` / `sub new` are both constructors, etc.).

## Grammar source

Uses [CodeAnt-AI's `tree-sitter-vb-dotnet`](https://github.com/CodeAnt-AI/tree-sitter-vb-dotnet), pulled in as a rev-pinned git dep on a minimal fork at [**davehenton/tree-sitter-vb-dotnet** on the `abi-14` branch](https://github.com/davehenton/tree-sitter-vb-dotnet/tree/abi-14):

```toml
tree-sitter-vb-dotnet = { git = "https://github.com/davehenton/tree-sitter-vb-dotnet", rev = "f8be17908ce2f445b9a034eb0d3d4191231ccca3" }
```

The fork holds CodeAnt-AI's upstream grammar with two mechanical tweaks — `parser.c` regenerated for ABI v14 (upstream ships v15, which our `tree-sitter = "0.22.6"` runtime rejects) and the Rust binding rewritten to the old-style `language() -> Language` API. `grammar.js` is unchanged from upstream. Same pattern we used for the Kotlin grammar in qltysh/cloud#1737. Once we bump workspace tree-sitter to 0.25+ (separate PR), we can drop the fork and point at the crates.io release directly.

## Implementation highlights

- Added `normalize_identifier()` to the `Language` trait (default no-op; VB.NET lowercases) and threaded it through LCOM, cognitive complexity, and fields metrics
- Lazy function-name extraction across smells/metrics callers to handle VB.NET's `constructor_declaration` node (no `name` field); incidentally fixed a latent eager-evaluation crash in JS/TSX
- Full CLI integration suite: `build` snapshot (509-line verified JSON), `metrics --json` (must emit `LANGUAGE_VBDOTNET`), and `smells` with `Imports` duplication filtering
- 6 focused unit tests for case-insensitive normalization behavior in qlty-smells

## Test plan

- [x] `cargo check` passes
- [x] `cargo test` — 995 tests passed, 0 failed, 2 ignored (pre-existing network tests)
- [x] `cargo test -p qlty --test integration vbnet_tests` — PASS
- [x] `cargo test -p qlty --test integration c_tests` — PASS
- [x] `cargo test -p qlty --test integration cpp_tests` — PASS
- [x] `cargo test -p qlty --test integration csharp_tests` — PASS
- [x] `cargo test -p qlty --test integration swift_tests` — PASS
- [x] `cargo test -p qlty-smells vbnet` — 5 passed
- [x] `cargo test -p qlty-types --test language_enum` — PASS
- [x] `cargo test -p qlty-analysis vbnet` — 12 unit tests passed


marschattha:
Can we use https://github.com/CodeAnt-AI/tree-sitter-vb-dotnet?

davehenton:
@marschattha Yes — that's the grammar, now via a rev-pinned git dep on a trimmed fork at https://github.com/davehenton/tree-sitter-vb-dotnet/tree/abi-14. Fork only regenerates parser.c for ABI v14 since our workspace pin is tree-sitter =  "0.22.6" (CodeAnt-AI's crate ships v15). Same pattern as qltysh/cloud#1737. More in the updated PR description.

marschattha:
Pushed 314ad0c5 with four focused fixes from a code review pass — all P2 correctness issues that the existing snapshot/tests didn't catch.

**`FIELD_QUERY` capture inversion** (qlty-analysis/src/lang/vbnet.rs:28). `@name` was bound to the receiver (`object`) but `metrics::fields` reads `@name` as the field name. So `Me.Foo` and `Me.Bar` both deduped to `me`, and `numbers.Length` / `sortedNumbers.Length` deduped to two distinct "fields". Swapped to bind `@name` to the `member` identifier. `Identical.vb` snapshot updates `fields: 2 → 1`. The existing `mixed_case_fields_deduplicate_after_normalization` test passed under either query (both `Me.Foo` and `Me.foo` happen to share both receiver and member); added `distinct_members_on_same_receiver_count_separately` to actually exercise the fix.

**`CLASS_QUERY` missing Module/Structure** (qlty-analysis/src/lang/vbnet.rs:7). VB `Module` and `Structure` blocks are real grammar nodes (`module_block`, `structure_block`) that can contain methods, but only `class_block` and `interface_block` were captured. Methods inside modules/structures were missing from class stats and LCOM grouping. Added both to the query.

**`function_name_node` silent footgun** (qlty-analysis/src/lang/vbnet.rs:213). Constructor declarations have no `name` field; the override returned `parameter_list` so downstream `.utf8_text()` wouldn't panic. Any future caller treating the result as a name would get parameter-list source. Replaced with a `panic!` pointing at `function_name_from_node`. Also dropped the dead `VBNet::IDENTIFIER` constant.

**Case-sensitive boolean operator comparison** (qlty-smells .../boolean_logic.rs, cognitive.rs). Operators were compared against the canonical PascalCase strings `AndAlso`/`OrElse`/`And`/`Or` from raw source text, so `andalso`, `ANDALSO`, etc. wouldn't match. Lowercased the constants and normalized the operator via `normalize_identifier` at the call sites — defaults to no-op for other languages, so they're untouched.

One thing surfaced while testing the operator fix that's worth a follow-up issue rather than blocking this PR: the grammar fork's `parser.c` declares `field_operator` but doesn't bind it on the `binary_expression` production — `child_count == 2` and `child_by_field_name("operator") == None` for `a AndAlso b`. The operator is dropped from the AST entirely, which is why `BooleanLogic.vb` snapshot reads `complexity: 1` (the `If`) and the four `AndAlso`s contribute nothing. The normalization fix is forward-compatible and starts working as soon as the grammar exposes the operator. Didn't add a test that asserts non-zero issues there since it'd be a regression trap.

`cargo test`: 996 pass, 2 ignored (pre-existing network tests). 5 files changed, +111/-15.

marschattha (approved):

PR #2766, opened 2026-04-23
feat(cli): centralize config loading with --skip-source-fetch support

noahd1:
## Summary

Re-applies #2762 (reverted in #2764) but with a backwards-compatibility fix for `qlty coverage publish`. When config loading fails — e.g. unfetchable remote sources, network issues, or a missing cached source under `--skip-source-fetch` — the publish now warns and falls back to the default config instead of aborting.

## Why the re-apply

#2762 centralized source-fetch/config-load and fixed a bug where `coverage/utils.rs::load_config` discarded cached remote source config on fetch failure. It was reverted because it also made `qlty coverage publish` bail on any config load error, which broke customers with unfetchable custom remote sources — previously those errors were silently swallowed by `unwrap_or_default()`.

## What's different from #2762

`qlty-cli/src/commands/coverage/utils.rs::load_config` is now graceful:

- Returns `QltyConfig` (not `Result<QltyConfig>`).
- On any error from `workspace.config_exists()` or `workspace.load_config(skip)`:
  - Logs a warning via `tracing::warn!` (goes to the log file).
  - Prints a yellow `warning:` line to stderr (user-visible).
  - Falls back to `QltyConfig::default()`.
- Coverage publish proceeds rather than aborting.

Other commands (`check`, `fmt`, `metrics`, etc.) retain strict error behavior — only coverage publish is graceful, matching the pre-#2762 contract.

## Test plan

- [x] `cargo check` passes
- [x] `cargo test --workspace --lib` passes (all 996 lib tests)
- [x] `cargo fmt --all --check` clean
- [x] `qlty check --level=low` clean
- [ ] Manual: `qlty coverage publish` in a repo whose remote source is unfetchable — expect a yellow warning and the publish to still succeed with default config.
- [ ] Manual: `qlty coverage publish --skip-source-fetch` in a repo whose remote source is NOT cached — expect a warning and publish to proceed with default config.
- [ ] Manual: `qlty coverage publish --skip-source-fetch` in a repo whose remote source IS cached — expect success without warnings.
- [ ] Manual: `qlty coverage publish` (no flag) in a repo without `qlty.toml` — expect silent default config, no warning.


noahd1:

**Verdict:** risky — needs reviewer confirmation on the Qlty Cloud runner contract, otherwise merge-safe.

**Affected user segments:** CI users (segment 2), Qlty Cloud runner (segment 3). Local dev (segment 1) is unaffected by any finding here.

This PR re-applies #2762 with a direct fix for the regression that caused #2764 (the revert): `qlty coverage publish` no longer aborts when config loading fails against unfetchable remote sources. That fix is well-targeted — the graceful path is scoped only to `coverage/utils.rs::load_config` and other commands retain strict error behavior. Remaining BC concerns are around the new strict `validate_sources_cached()` check fanning out to every command that now passes through `prepare_sources()`.

### Findings

**[risky] `qlty-config/src/workspace.rs:56-72` + `qlty-config/src/sources/sources_list.rs:32-43` — new `validate_sources_cached()` on `--skip-source-fetch` / `build` default path**
- Before: `qlty check --skip-source-fetch`, `qlty fmt --skip-source-fetch`, and `qlty build` (without `--fetch-sources`) skipped the fetch and proceeded. Uncached git sources produced late/unclear failures or silently worked when config didn't exercise them.
- After: all three now call `prepare_sources(true)` → `validate_sources_cached()`, which `bail!`s with `"Source {...} is not available locally. Run without --skip-source-fetch, or run qlty sources fetch first."` if any `GitSource` entry's `local_origin_ref_path` doesn't exist. `DefaultSource` and `LocalSource` still report cached=true trivially, so default users are unaffected.
- Who breaks:
  - **Segment 2 (CI):** customers with a `[[source]] repository = ...` in `.qlty/qlty.toml` who invoke `qlty fmt --skip-source-fetch` / `qlty check --skip-source-fetch` / `qlty build` in CI without a prior `qlty sources fetch` step. They get exit non-zero where they previously got exit 0 (or a different downstream error).
  - **Segment 3 (Qlty Cloud runner):** `workflows/format_pull_request.yml` (and siblings) run `qlty fmt --skip-source-fetch`; the `qlty-build` action runs `qlty build`. Per the task contract the runner also invokes `qlty sources fetch` upstream, so in the happy path the cache is populated and this check passes. **But** any customer repo where the source fetch step silently no-ops (stale cache, partial clone, credential skew on a private repo) would previously have limped along and will now hard-fail with the new bail.
- Suggested mitigation: this strictness is arguably a fix, not a break — but worth (a) confirming with the runner team that `qlty sources fetch` is both required and reliably populates the `local_origin_ref_path` that `GitSource::is_cached` checks, and (b) noting the behavior change in the PR description / CHANGELOG so CI users with custom git sources know to add a fetch step before upgrading.

**[note] `qlty-cli/src/commands/coverage/utils.rs:14-49` — new stderr warning on publish config failures**
- Before (current `main`, post-revert): `load_config()` returned `QltyConfig::default()` via `unwrap_or_default()` on any error, silently.
- After: when `.qlty/qlty.toml` exists and `workspace.load_config(skip)` errors, `coverage publish` prints a yellow `warning: Failed to load qlty config: <msg>. Proceeding with default configuration.` line to stderr and logs `warn!`. When `.qlty/qlty.toml` is absent the behavior is still silent (per the second commit). Exit code is unchanged — the publish still succeeds with default config.
- Who sees the difference: **segment 2 (CI)** customers whose `qlty coverage publish` step previously ran silent against an unfetchable/misconfigured remote source will now see a stderr warning line. CI log scrapers / stderr-as-error guards could over-react. No exit code change, so no build failure.
- Suggested mitigation: none needed — this is the intended behavior of the re-apply and strictly improves observability. Call it out in release notes so log-scrapers aren't surprised.

**[note] `qlty-cli/src/commands/coverage/publish.rs:152-154` — new `--skip-source-fetch` flag on `qlty coverage publish`**
- Additive. Optional. Defaults to `false`, preserving the existing fetch-on-publish behavior. Does not shadow or conflict with any existing flag. `qlty coverage complete` is unchanged (it never called `load_config`). Safe.

- c. qlty.toml schema — no serde struct changes; only new methods on `Workspace`, `SourcesList`, and the `Source` trait (with a default impl so external `Source` implementors don't break).
- d. Output formats — no changes to `qlty-coverage/src/print.rs`, no JSON wire-format changes, no `qlty-types` proto changes.
- e. Exit codes — `coverage publish` still exits 0 on config-load failure (with new warning). New non-zero paths are gated behind the existing `--skip-source-fetch` opt-in or the existing `qlty build` default (addressed under finding #1).
- f. Environment variables — no `std::env::var` call sites touched; no `QLTY_*` changes; `qlty-coverage/src/token.rs` untouched.
- g. Source fetching / cache layout — no directory-layout changes under `.qlty/sources/`; `GitSource::is_cached()` only *reads* `local_origin_ref_path`.
- h. Plugin behavior — no `plugin.toml` schema or loader changes; `affects_cache` / `prepare_script` / download URLs untouched.
- i. Telemetry / logging defaults — new `tracing::info!("Reading qlty config from ...")` and `tracing::warn!` on failure are additive; nothing was downgraded or removed.

### Notes for reviewer

- **Please confirm with the runner team** (qltysh/runner) that `workflows/analyze_pull_request.yml`, `workflows/analyze_branch.yml`, and `workflows/format_pull_request.yml` always run `qlty sources fetch` successfully before any `--skip-source-fetch` invocation and before `qlty build`, and that the fetched path matches `GitSource::local_origin_ref_path`. If there's any path where the runner skips the fetch (e.g. `qlty-init`-only flows), customers with a custom `[[source]] repository = ...` will start failing on upgrade.
- Consider a CHANGELOG entry calling out (a) the new `--skip-source-fetch` flag on `coverage publish`, (b) the new stderr warning on publish config-load failures, and (c) that `--skip-source-fetch` / `qlty build` now validate the local source cache and fail fast when it's missing.
- The manual test plan in the PR description is the right shape — particularly the "remote source NOT cached" scenario, which is the one that would have surfaced the #2762 regression if it had been run. Worth completing those manual runs before un-drafting.

---

PR #2773, opened 2026-04-29
fix(editorconfig-checker): handle -1 line numbers in output parser

### LINKED ISSUE #2772: editorconfig-checker: parser crashes on -1 column values (u32 overflow)
## Summary

When `editorconfig-checker` reports an issue with a column number of `-1`, the Qlty output parser fails with:

```
editorconfig-checker/lint: Error parsing output: invalid value: integer `-1`, expected u32 at line 1 column N
```

The result is that all findings from the affected invocation are silently dropped — they never appear in the build output or dashboard.

## Reproduction

Run `qlty build` or `qlty check` against a repo where `editorconfig-checker` emits `-1` as a column value. This appears to happen on certain file types (e.g. binary-adjacent or non-UTF-8 adjacent files).

## Expected behavior

The parser should handle `-1` column values gracefully — either by clamping to `0`, treating it as "unknown column", or surfacing a warning without dropping all findings from the invocation.

## Actual behavior

The parser returns an error and drops all findings from the invocation silently.

## Relevant log output

```
ERROR qlty_check::executor::driver: editorconfig-checker/lint: Error parsing output: invalid value: integer `-1`, expected u32 at line 1 column 259
```

## Notes

- Affects `qlty_check::executor::driver` output parsing
- No action needed from the user — their `.editorconfig` file is not the cause
- Multiple invocations can be affected in a single build, each silently losing findings

laura-mlg:
## Summary

Fixes #2772

`editorconfig-checker` emits `-1` as a line number when it cannot determine the location of an issue. The previous `u32` fields in `CodeClimateLines` caused `serde_json` to fail deserializing the entire output payload, silently dropping all findings from the invocation.

## Changes

- Changed `begin` and `end` in `CodeClimateLines` from `u32` to `i64`
- Clamp to `0` on conversion to `u32` when constructing the `Range`, so issues with unknown locations are still reported
- Added a test case covering the `-1` input

## Test plan

- [ ] Existing snapshot test still passes
- [ ] New `parse_negative_line_number` test passes
- [ ] Run `qlty check` against a repo that previously triggered this error and confirm findings are no longer dropped

### DIFF
--- a/qlty-check/src/parser/editorconfig_checker.rs
+++ b/qlty-check/src/parser/editorconfig_checker.rs
@@ -31,9 +31,9 @@ struct CodeClimateLocation {
 
 #[derive(Debug, Deserialize)]
 struct CodeClimateLines {
-    begin: u32,
+    begin: i64,
     #[serde(default)]
-    end: u32,
+    end: i64,
 }
 
 #[derive(Debug, Default, Serialize, Deserialize, Clone)]
@@ -48,11 +48,13 @@ impl Parser for EditorconfigChecker {
             let rule_key = derive_rule_key(&message.description);
             // editorconfig-checker's codeclimate formatter sets `end` to
             // AdditionalIdenticalErrorCount (a relative count) rather than
-            // an absolute line number, so we compute: begin + end
+            // an absolute line number, so we compute: begin + end.
+            // The tool may emit -1 for unknown locations; clamp to 0.
+            let start_line = message.location.lines.begin.max(0) as u32;
             let end_line = if message.location.lines.end > 0 {
-                message.location.lines.begin + message.location.lines.end
+                (message.location.lines.begin + message.location.lines.end).max(0) as u32
             } else {
-                message.location.lines.begin
+                start_line
             };
 
             let issue = Issue {
@@ -64,7 +66,7 @@ impl Parser for EditorconfigChecker {
                 location: Some(Location {
                     path: message.location.path,
                     range: Some(Range {
-                        start_line: message.location.lines.begin,
+                        start_line,
                         end_line,
                         ..Default::default()
                     }),
@@ -115,6 +117,21 @@ fn severity_to_level(severity: &str) -> Level {
 #[cfg(test)]
 mod test {
     use super::*;
+
+    #[test]
+    fn parse_negative_line_number() {
+        let input = r###"
+[{"check_name":"editorconfig-checker","description":"Wrong character encoding","fingerprint":"abc123","severity":"minor","location":{"path":"foo.h","lines":{"begin":-1,"end":0}}}]
+        "###;
+
+        let issues = EditorconfigChecker::default().parse("editorconfig-checker", input);
+        assert!(issues.is_ok(), "should not fail on -1 line number");
+        let issues = issues.unwrap();
+        assert_eq!(issues.len(), 1);
+        let range = issues[0].location.as_ref().unwrap().range.as_ref().unwrap();
+        assert_eq!(range.start_line, 0);
+        assert_eq!(range.end_line, 0);
+    }
 
     #[test]
     fn parse() {


noahd1:
🤖 **Backwards-compatibility review in progress**

I'm running the `bc-check` skill against this PR to look for any behavior changes that might break existing qlty users on upgrade. I'll edit this comment with findings shortly.

---

noahd1:

**Target:** PR #2773 — `fix(editorconfig-checker): handle -1 line numbers in output parser`
**Verdict:** safe
**Affected user segments:** local devs (segment 1) — minor surface-level shift only

### Findings

**[note]** `qlty-check/src/parser/editorconfig_checker.rs:45` — Behavior changes on existing command paths
- **Before:** When `editorconfig-checker` emitted `-1` for any line number, `serde_json::from_str::<Vec<CodeClimateIssue>>` failed against the `u32` field types and the entire payload was discarded. The parser returned `Err` for that invocation, so users saw zero `editorconfig-checker` findings.
- **After:** `begin`/`end` deserialize as `i64` and are clamped to `0` when narrowing to `u32`. The payload deserializes cleanly and findings flow through.
- **Who breaks:** Only segment 1 (local devs running `qlty check`). Users whose `editorconfig-checker` output had been silently dropping will now see real findings appear — and in rare cases that flips `qlty check`'s exit status from 0 to non-zero on first run after upgrade. This is the *inverse* of the #2762 pattern (it surfaces output that was being silently dropped, rather than turning a soft error into a hard one), and the findings were always real. Segment 2 (CI / `qlty coverage publish|complete`) doesn't go through this parser. Segment 3 (Qlty Cloud, which invokes `qlty sources fetch`, `qlty config validate`, `qlty fmt --skip-source-fetch`, `qlty init`, `qlty install`, `qlty build`) doesn't exercise this code path.
- **Suggested mitigation:** None required. Worth a one-line CHANGELOG entry under "Fixed" so anyone who suddenly sees new editorconfig-checker findings after upgrading knows where they came from.

- CLI flag surface (`qlty-cli/src/commands/**`) — no flags touched
- `qlty.toml` schema (`qlty-config/src/config/**`) — no schema fields touched
- Output formats — `Range { start_line: u32, end_line: u32 }` wire shape preserved; the `i64` is purely an intermediate inside the parser
- Exit codes — no direct exit-code logic changed (downstream effect noted above)
- Environment variables — none touched
- Source fetching / cache layout — none touched
- Plugin behavior (`plugin.toml`, `drivers`, download URLs, `affects_cache`) — none touched
- Telemetry / logging defaults — none touched

### Notes for reviewer
- `CodeClimateLines` is a private struct used only inside this parser (verified: one definition site, no external `use`). Changing its field types from `u32` to `i64` has no API surface effect.
- The new test covers `begin: -1, end: 0`. Worth eyeballing `begin: -1, end: -1` (both negative): with `end > 0` false, the code falls into the `else` branch and uses `start_line` (clamped to `0`), so both ends land at `0`. Looks correct.


---

qlty-check/src/parser/editorconfig_checker.rs
```diff
@@ -48,11 +48,13 @@ impl Parser for EditorconfigChecker {
             let rule_key = derive_rule_key(&message.description);
             // editorconfig-checker's codeclimate formatter sets `end` to
             // AdditionalIdenticalErrorCount (a relative count) rather than
-            // an absolute line number, so we compute: begin + end
+            // an absolute line number, so we compute: begin + end.
+            // The tool may emit -1 for unknown locations; clamp to 0.
+            let start_line = message.location.lines.begin.max(0) as u32;
```
marschattha, line 53: Probably worth a double check but I think it should be `max(1)` here and below in endLine computation.
Line numbers are expected to start from 1, not 0.

marschattha (changes requested):

qlty-check/src/parser/editorconfig_checker.rs
davehenton: @marschattha per Claude: 
> .max(0) is actually correct here. Elsewhere in the pipeline (e.g. source_extractor.rs:80) start_line == 0 is treated as the "unknown / not provided" sentinel — proto3 u32 defaults to 0 too. Mapping -1 to 1 would attach the issue to the real first line of the file, turning an "unknown location" finding into a phantom line-1 finding. Keeping it at 0 lets downstream code recognize it as locationless, which matches the intent of editorconfig-checker emitting -1.

marschattha (approved):

qlty-check/src/parser/editorconfig_checker.rs
marschattha: Ah thanks approved.

PR #2782, opened 2026-05-01
fix: honor .qlty/configs/ for biome and knip plugins

laura-mlg:
## Summary

- Add `copy_configs_into_tool_install = true` and `config_script` to biome (lint + format) and knip (lint) drivers
- Add a trycmd integration test verifying a config in `.qlty/configs/` is copied to the tool install dir and accessible via `${config_file}`
- Closes the bug reported in https://github.com/qltysh/support/issues/296

## Root cause

Both plugins relied on the workspace-root symlink mechanism (`load_config_file_from_qlty_dir`, `executor.rs:350–370`) to make `.qlty/configs/` files visible. This mechanism silently breaks in two ways:

1. **Knip** uses `runs_from = target_directory` — its CWD is the directory containing `package.json`, not workspace root. The symlink placed at workspace root is invisible to knip's config discovery in any monorepo or subdirectory setup.
2. **Biome** runs from workspace root, so the symlink is technically in its CWD, but biome does not follow symlinks during config discovery.

## Fix

Follow the same pattern used by phpstan and eslint v9+:

- Add `config_script = "--config-path ${config_file}"` (biome) / `config_script = "--config ${config_file}"` (knip)
- Set `copy_configs_into_tool_install = true` on all affected drivers

`replace_config_script()` replaces `${config_script}` with the empty string when no config is found, so both tools continue to work for users without a config file. When a config is present (whether at repo root or in `.qlty/configs/`), it is physically copied into `tool.directory()` and the absolute path is passed to the tool via CLI flag — independent of CWD and symlink-following behaviour.

## Test plan

- [ ] New trycmd test `copy_config_to_tool_install_from_qlty_configs` — places `config.yml` in `.qlty/configs/`, runs a plugin with `copy_configs_into_tool_install = true` and `config_script`, verifies the config is accessible at the tool-dir path (i.e. the copy succeeded and `${config_file}` resolved correctly)
- [ ] Existing `copy_config_to_tool_install_with_exported_config` test continues to pass
- [ ] CI check tests pass

> **Note:** The expected stdout/stderr files for the new test were written to match the existing `copy_config_to_tool_install_with_exported_config` test pattern. If minor whitespace differences cause a snapshot mismatch, run `TRYCMD=overwrite cargo test -p qlty-cli copy_config_to_tool_install_from_qlty_configs` to regenerate them.

### DIFF
--- a/qlty-cli/tests/cmd/check/copy_config_to_tool_install_from_qlty_configs.in/.gitignore
+++ b/qlty-cli/tests/cmd/check/copy_config_to_tool_install_from_qlty_configs.in/.gitignore
@@ -1,0 +1,4 @@
+.qlty/results
+.qlty/logs
+.qlty/out
+.qlty/sources
--- a/qlty-cli/tests/cmd/check/copy_config_to_tool_install_from_qlty_configs.in/.qlty/configs/config.yml
+++ b/qlty-cli/tests/cmd/check/copy_config_to_tool_install_from_qlty_configs.in/.qlty/configs/config.yml
@@ -1,0 +1,1 @@
+configuration: true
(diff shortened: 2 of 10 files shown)


marschattha:
Breaking this into https://github.com/qltysh/qlty/pull/2788 and https://github.com/qltysh/qlty/pull/2789

PR #2791, opened 2026-05-08
Revert "fix(biome): pass --config-path explicitly to support .qlty/configs/"

marschattha:
Reverts qltysh/qlty#2788


noahd1:
🔍 **Backwards-compatibility check in progress…**

I'm analyzing this PR for behavior changes that could break existing qlty users. I'll update this comment with findings shortly.

---

noahd1:

**Target:** PR #2791 — *Revert "fix(biome): pass --config-path explicitly to support .qlty/configs/"*
**Verdict:** **risky — needs mitigation**
**Affected user segments:** local devs, CI users, Qlty Cloud (anyone running biome v2 with a `.qlty/configs/biome.json`)

### Findings

**[risky]** `qlty-plugins/plugins/linters/biome/plugin.toml:13-39` — plugin behavior (category h)

- **Before (current `main`, shipped in v0.629.0 via #2788):** Biome plugin has two version-gated driver definitions. For `version >= 2.0.0` it runs `biome lint --reporter=json --config-path ${config_file} ${target}` and sets `copy_configs_into_tool_install = true`, which writes the user's `.qlty/configs/biome.json` into the tool install directory (outside the project tree). This dodges biome v2's "nested root configuration" detection while still honoring the user's config.
- **After this revert:** A single driver definition runs `biome lint --reporter=json ${target}` for all versions. Biome v2's project-walk discovery sees both the workspace-root symlink to `.qlty/configs/biome.json` *and* the file at `.qlty/configs/biome.json` itself, and aborts with `× Found a nested root configuration, but there's already a root configuration` before reading either. Net effect: biome v2 users with a config in `.qlty/configs/` get a hard failure where v0.629.0 succeeds.
- **Who breaks:**
  - **Segment 1 (local devs):** Anyone on biome ≥ 2.0.0 with a checked-in `.qlty/configs/biome.json` will see `qlty check` fail on biome where it just started working in 0.629.0. This is the exact scenario from [support#299](https://github.com/qltysh/support/issues/299) — the bug the original PR fixed comes back.
  - **Segment 2 (CI users):** Same scenario — every CI build running `qlty check` against biome v2 with a qlty-managed config fails. CI segment baseline severity is blocker; flagging risky here only because users can downgrade biome to <2.0.0 or move the config to the project root as a workaround.
  - **Segment 3 (Qlty Cloud):** Qlty Cloud invokes `qlty fmt --skip-source-fetch` and `qlty check`-adjacent commands on hosted builds. Any customer whose config relies on biome v2 + `.qlty/configs/biome.json` will see hosted builds fail until they reconfigure or qlty downgrades biome.
- **Suggested mitigation:** Either (a) include a rationale in the PR body explaining why the fix is being reverted (regression, unintended side effect, etc.) so reviewers can weigh the trade-off; (b) keep the v2 driver block but narrow whatever broke about `copy_configs_into_tool_install`; or (c) hold the revert until a follow-up fix lands so 0.629.0 users don't experience a regression on the next upgrade. If this must merge as-is, call it out in the changelog under "regressions" so customers running biome v2 know to pin or downgrade.

- a. Behavior changes on existing command paths (no Rust changes)
- b. CLI flag surface (no `qlty-cli/src/commands` changes)
- c. `qlty.toml` schema (no `qlty-config` changes)
- d. Output formats (no print/serde changes; the `.shot` snapshot deletion is just removing a fixture for a test that no longer applies)
- e. Exit codes (no changes to exit logic)
- f. Environment variables (no `QLTY_*` changes)
- g. Source fetching / cache layout (no changes)
- i. Telemetry / logging defaults (no changes)

### Notes for reviewer
- The PR description is just `Reverts qltysh/qlty#2788` with no stated reason. Before merging, please document **why** the original fix is being pulled — there's a parallel revert pattern with #2789 (knip), so a shared rationale would help reviewers and customers tracking the next release notes.
- Verify the biome `1.9.4` `known_good_version` removal is intentional. It's just metadata, but it was added alongside the fix to pin the last-known-good v1 release.
- If the original fix had an unintended side effect (e.g., `copy_configs_into_tool_install` interacting badly with another config), capturing that in a follow-up issue would prevent a future re-fix from re-introducing the same problem.

---

PR #2797, opened 2026-06-23
fix(check): scrub env values from serialized invocation records

marschattha:
## Summary

`Invocation.env` recorded the full environment (names **and** values) of each plugin invocation. When a value was interpolated from a secret via `${env.NAME}`, that secret was serialized into invocation artifacts:

- `.qlty/builds/<build_id>/invocations-*.jsonl`
- `.qlty/builds/<build_id>/invocations.json.gz`
- `.qlty/out/invoke-*.yaml`

These artifacts get uploaded unredacted, since redaction only covers stdout/stderr — never serialized structs.

## Fix

Record only environment variable **names**, not their values, at the point `Invocation` is constructed (`invocation_result.rs`). Because `env` is write-only diagnostic data that is never read back, scrubbing at the source covers every serialization path (jsonl, gz, yaml outfile, `build --print`) with a single change.

Keys are preserved so the diagnostic value of "which vars were set" is retained — only the values are dropped.

## Escape hatch

Set `QLTY_INSECURE_KEEP_INVOCATION_ENV=1` (or `true`) to retain values for local debugging. Follows the existing `QLTY_INSECURE_ALLOW_HTTP` convention; default behavior is secure.

## CI

Separately, this branch pins the Rust-compiling Windows jobs (`cli.yml`, `build.yml`) to `windows-2022`. The `windows-latest` image rolled to VS 18 / MSVC 14.51, which the `aws-lc-sys` build (via rustls's `aws-lc-rs` backend) can't compile — `cmake 0.1.53` panics with "couldn't determine visual studio generator", and bumping cmake/cc is blocked by `tree-sitter-javascript 0.21.4` pinning `cc ~1.0.90`. This is pre-existing infra breakage, unrelated to the secret-scrubbing change.

## Tests

Added unit tests for the scrub transform. All 211 `qlty-check` tests pass; typecheck and `cargo fmt` clean.

Ref: qltysh/qltysh#635

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
 
           # Windows
           - target: x86_64-pc-windows-msvc
-            runner: windows-latest
+            runner: windows-2022
             archive_format: zip
     name: ${{ matrix.target }}
     runs-on: ${{ matrix.runner }}
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-15, windows-latest]
+        os: [ubuntu-latest, macos-15, windows-2022]
     name: ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     concurrency:
@@ -57,7 +57,7 @@ jobs:
         run: cargo llvm-cov --lcov --output-path target/lcov.info -- --include-ignored
 
       - uses: qltysh/qlty-action/coverage@f071a76ab07c05c4326f50342b7b477db89624f6 # node24-smoke / pre-release
-        if: ${{ matrix.os != 'windows-latest' }}
+        if: ${{ matrix.os != 'windows-2022' }}
         with:
           oidc: true
           files: target/lcov.info
--- a/qlty-check/src/executor/invocation_result.rs
+++ b/qlty-check/src/executor/invocation_result.rs
@@ -42,6 +42,28 @@ pub enum InvocationStatus {
     ParseError,
 }
 
+fn keys_only_env(env: HashMap<String, String>) -> HashMap<String, String> {
+    if env_scrubbing_disabled() {
+        return env;
+    }
+
+    scrub_env_values(env)
+}
+
+fn scrub_env_values(env: HashMap<String, String>) -> HashMap<String, String> {
+    env.into_keys().map(|key| (key, String::new())).collect()
+}
+
+fn env_scrubbing_disabled() -> bool {
+    match std::env::var("QLTY_INSECURE_KEEP_INVOCATION_ENV") {
+        Ok(value) => {
+            let normalized = value.trim().to_ascii_lowercase();
+            normalized == "1" || normalized == "true"
+        }
+        Err(_) => false,
+    }
+}
+
 impl InvocationResult {
     pub fn from_command_output(
         plan: &InvocationPlan,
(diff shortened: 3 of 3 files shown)


noahd1:
🔍 **Backwards-compatibility review in progress…**

A Claude routine is analyzing this PR for behavior changes that could break existing qlty users on upgrade. I'll edit this comment with the report when the check completes.

---

noahd1:

**Target:** PR #2797 — `fix(check): scrub env values from serialized invocation records`
**Verdict:** safe — intentional, security-driven behavior change with an opt-out escape hatch
**Affected user segments:** none broken; **Qlty Cloud (segment 3) should be informed** if any backend tooling reads `env` values from uploaded invocation artifacts

### Findings

**[note]** `qlty-check/src/executor/invocation_result.rs:107-117` — Output formats (category d)
- Before: `Invocation.env` was serialized into `.qlty/builds/<id>/invocations-*.jsonl`, `.qlty/builds/<id>/invocations.json.gz`, and `.qlty/out/invoke-*.yaml` with both names **and** real values (including secrets interpolated via `${env.NAME}`).
- After: the same fields are still present and the wire schema (`map<string, string>` in `qlty.analysis.v1.Invocation`, proto tag 12 / 13) is unchanged — only the **values** are replaced with empty strings. Keys are preserved, so "which vars were set" is still legible.
- Who breaks:
  - **Segment 1 (local devs):** no functional break. A developer inspecting `.qlty/out/invoke-*.yaml` for debugging will see empty values instead of real ones. The escape hatch `QLTY_INSECURE_KEEP_INVOCATION_ENV=1` restores the old view locally.
  - **Segment 2 (CI customers):** not affected. `qlty coverage publish` / `qlty coverage complete` do not consume or emit invocation env artifacts; exit codes, schema, and required inputs are unchanged.
  - **Segment 3 (Qlty Cloud):** Qlty Cloud invokes `qlty build` / `qlty fmt --skip-source-fetch`, both of which produce these invocation artifacts and upload them. If any Cloud backend code reads `invocation.env` values out of those artifacts (for a UI panel, debugging tools, search index, etc.), those values will be empty after this change. The qlty CLI codebase itself never reads the field back (`grep` for `invocation.env` / `inv.env` / `.env.get` over the repo turns up zero consumers), confirming the PR author's "write-only diagnostic data" claim within this repo — but Qlty Cloud is out-of-tree.
- Suggested mitigation: confirmed mitigations already in place: (1) wire schema unchanged, (2) keys preserved, (3) opt-out via `QLTY_INSECURE_KEEP_INVOCATION_ENV`. **Action item:** notify the Qlty Cloud team before/at release so any backend code reading `invocation.env` values stops relying on them (and so they understand the prior values were leaking secrets and shouldn't be re-introduced upstream).

**[note]** `qlty-check/src/executor/invocation_result.rs:55-63` — Environment variables (category f)
- Before: no such variable.
- After: new `QLTY_INSECURE_KEEP_INVOCATION_ENV` env var (`1` / `true` to disable scrubbing).
- Who breaks: nobody — purely additive, follows the existing `QLTY_INSECURE_ALLOW_HTTP` naming convention. Default is secure.
- Suggested mitigation: none needed. Worth a CHANGELOG/docs line so security-conscious users know the hatch exists.

- a. Behavior changes on existing command paths — exit codes, success/failure, validation: **unchanged**. No new `bail!`, no `?` propagation past a previously-swallowed result, no new network/IO on the hot path.
- b. CLI flag surface — no flags added, removed, renamed, or re-defaulted.
- c. `qlty.toml` schema — `qlty-config/` is untouched.
- e. Exit codes — no change to `CommandError` / `CommandSuccess` / `?` flow.
- g. Source fetching / cache layout — `.qlty/sources/`, `.qlty/cache/`, `~/.qlty/` all untouched.
- h. Plugin behavior — `plugin.toml` loader, drivers, `prepare_script`, `affects_cache`, download URLs: **unchanged**. The plugin's effective runtime env is **not** affected; only the diagnostic record of it is.
- i. Telemetry / logging defaults — no log-level changes.

### Notes for reviewer

- Verify the **proto field is preserved** so downstream consumers parsing the gz/jsonl don't trip on a missing field: confirmed at `qlty-types/src/protos/qlty.analysis.v1.rs:191-193` (`pub env: HashMap<String, String>`) — the field is still emitted, just with empty values, which `serde_json` round-trips as `""`. Existing parsers that handle the field as a map of strings continue to work.
- The escape-hatch env-var check happens at every `InvocationResult::from_command_output` call. That's the right place (called per invocation), but it does mean toggling `QLTY_INSECURE_KEEP_INVOCATION_ENV` mid-run can change behavior between plugin invocations. Almost certainly fine in practice (env doesn't usually mutate mid-process), but worth noting.
- Heads-up for whoever ships the next release notes / migration doc: a one-liner like "invocation artifacts no longer record environment variable values; set `QLTY_INSECURE_KEEP_INVOCATION_ENV=1` for local debugging" closes the loop for any user who notices the empty values.
- **Coordinated rollout with Qlty Cloud team is the only thing that could turn this from a `note` into a `risky`** — confirm with them before merging.

---
*Automated BC check by Claude routine. Scope: behavior changes affecting existing users on upgrade. Not a code-quality / correctness review.*

---

noahd1 (approved):

PR #2802, opened 2026-07-03
Add --selection to qlty coverage publish and complete

marschattha:
## Summary

Adds a bare `--selection` flag to `qlty coverage publish` **and** `qlty coverage complete` for uploads produced by a selective (subset) test run — e.g. only the tests affected by a pull request's changes. When set, the CLI sends `selection: "selected"`; when omitted, the field is absent (a full test-suite run). Labeled selections are not supported yet — the server accepts exactly `"selected"` — but the wire protocol carries a string so labels can be introduced later without protocol changes.

The server treats a non-null `selection` as selective coverage: it contributes to diff coverage only — no total coverage, no component coverage — and selected/unselected uploads for the same commit and tag assemble into **separate reports**. That last part is why `complete` needs the flag too: a complete call finalizes the report keyed by its selection, so it must be passed exactly when the publish it completes passed it. Without it, a selective multi-part (`--incomplete`) upload can never be finalized.

- `selection` (optional string, tag 44) on the CoverageMetadata proto + pbjson serde
- `publish`: `--selection` → `Settings` → `MetadataPlanner::compute()` → metadata.json
- `complete`: `--selection` → `Settings` → `MetadataPlanner::compute_minimal()` → the complete request body

### DIFF
--- a/qlty-cli/src/commands/coverage/complete.rs
+++ b/qlty-cli/src/commands/coverage/complete.rs
@@ -54,12 +54,11 @@ pub struct Complete {
     /// workspace and if it cannot be inferred from the git origin.
     pub project: Option<String>,
 
-    #[arg(long, num_args = 0..=1, require_equals = true, default_missing_value = "selected")]
-    /// Complete the uploads published with the same --selection value.
-    /// Selected and unselected uploads assemble into separate reports, so this
-    /// must match the value passed to qlty coverage publish. Defaults to
-    /// "selected"; pass --selection=<label> to name the selection.
-    pub selection: Option<String>,
+    #[arg(long)]
+    /// Complete the uploads published with --selection. Selected and
+    /// unselected uploads assemble into separate reports, so pass this
+    /// exactly when it was passed to qlty coverage publish.
+    pub selection: bool,
 
     #[arg(long)]
     /// Perform a dry-run without actually completing the coverage
@@ -143,7 +142,7 @@ impl Complete {
             tag: self.tag.clone(),
             quiet: self.quiet,
             project: self.project.clone(),
-            selection: self.selection.clone(),
+            selection: self.selection.then(|| "selected".to_string()),
             ..Default::default()
         }
     }
--- a/qlty-cli/src/commands/coverage/complete.rs
+++ b/qlty-cli/src/commands/coverage/complete.rs
@@ -54,6 +54,13 @@ pub struct Complete {
     /// workspace and if it cannot be inferred from the git origin.
     pub project: Option<String>,
 
+    #[arg(long, num_args = 0..=1, require_equals = true, default_missing_value = "selected")]
+    /// Complete the uploads published with the same --selection value.
+    /// Selected and unselected uploads assemble into separate reports, so this
+    /// must match the value passed to qlty coverage publish. Defaults to
+    /// "selected"; pass --selection=<label> to name the selection.
+    pub selection: Option<String>,
+
     #[arg(long)]
     /// Perform a dry-run without actually completing the coverage
     pub dry_run: bool,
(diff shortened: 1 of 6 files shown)


noahd1:
🔍 **Backwards-compatibility check in progress…**

I'm reviewing this PR for any behavior changes that could break existing qlty users on upgrade (CLI flags, `qlty.toml` schema, output formats, exit codes, etc.). I'll edit this comment with the findings when the check completes.

---

noahd1:

**Target:** PR #2802 — Add `--selective` flag to `qlty coverage publish`
**Verdict:** ✅ **safe**
**Affected user segments:** none — the change is purely additive and opt-in

### Findings

No backwards-incompatible changes found. Walk-through of the risk surface:

- **New CLI flag** (`qlty-cli/src/commands/coverage/publish.rs:152-156`): `--selective` is a new boolean flag that defaults to `false`. It does not shadow an existing flag, does not add a positional argument, and no existing flag was removed, renamed, or had its default changed. A user with a six-month-old invocation script sees identical behavior.
- **Settings struct** (`qlty-coverage/src/publish/settings.rs:29`): new `selection: Option<String>` field. `Settings` derives `Default`, and the only in-tree constructor (`Publish::build_settings` at `publish.rs:280-308`) sets it explicitly. `qlty-coverage` is an internal workspace crate, not a published-API surface, so adding a field is not a downstream break.
- **Proto / `metadata.json` output** (`qlty-types/src/protos/qlty.tests.v1.rs:91-92`, `qlty.tests.v1.serde.rs`): new optional field `selection` at proto tag 44. Adding a new optional protobuf field at a fresh tag is BC-compatible on the wire. On the JSON side, the serializer writes `selection` only when `is_some()` — when the flag isn't passed, `metadata.json` is byte-identical to today. Existing server-side consumers that don't know about the field are not exposed to it.
- **Command hot path** (`publish.rs:174-249`): no new `bail!`, `ensure!`, network fetch, or validation added. The `#2762`-style silent-failure-becomes-hard-error pattern is not present.
- **Exit codes, env vars, `qlty.toml` schema, source fetching, plugin behavior, telemetry**: not touched.

### Cross-segment check

- **Segment 1 (local devs)**: not affected — `qlty coverage publish` is a CI command; local flow (`qlty check`, `qlty fmt`, `.qlty/qlty.toml`) is untouched.
- **Segment 2 (CI customers)**: not affected on upgrade — existing pipelines that don't pass `--selective` produce the same `metadata.json` and hit the same server code paths. Only pipelines that explicitly opt in via `--selective` emit the new field.
- **Segment 3 (Qlty Cloud)**: not affected — Qlty Cloud invokes `qlty sources fetch`, `qlty config validate`, `qlty fmt --skip-source-fetch`, `qlty init`, `qlty install`, and `qlty build`. None of those touch `coverage publish` or the changed types.

- Behavior changes on existing command paths
- CLI flag surface
- `qlty.toml` schema
- Output formats (proto + JSON)
- Exit codes
- Environment variables
- Source fetching / cache layout
- Plugin behavior
- Telemetry / logging defaults

### Notes for reviewer

- **Server-side coordination (operational, not a BC break):** a user who upgrades the CLI and opts in with `--selective` will send `"selection": "selective"` in `metadata.json`. Confirm the ingest side accepts (or ignores) the new field on the target Qlty Cloud instance before this flag is documented publicly. If the server does strict field validation and hasn't shipped the counterpart, opt-in users would see upload failures. This is opt-in, so it is not a BC break on upgrade — but the rollout ordering is worth checking before promoting the flag in docs.
- **Field is a `String` not a `bool`** — as the PR description notes, this leaves room for future `--selection <label>` variants without a wire change. No BC concern; just calling out the deliberate design choice.


---

noahd1 (approved):
@marschattha If you pass `--selective` it should set `selection` to `selected` (a default that for now can't be overridden).

marschattha:
`selective`, `selection` and `selected`, do we want 3 different terms?

noahd1:
Another option would be `--selection` which defaults its `selection` to "selected" and could later take an argument.

noahd1 (approved):

PR #2805, opened 2026-07-07
fix(check): recreate dangling .qlty symlinks after cache deletion

noahd1:
## Problem

If `~/.qlty/cache/repos` is deleted (a natural troubleshooting step), the `.qlty/out` and `.qlty/logs` symlinks in every previously-checked workspace are left dangling. The next `qlty check` then fails with an opaque error:

```
 > FATAL error occurred running 7 invocations
```

The logs show why:

```
ERROR qlty_config::library: Failed to create symlink from .../cache/repos/<hash>/out to .../.qlty/out: File exists (os error 17)
ERROR qlty_check::executor::invocation_result: Failed to create directory: .../.qlty/out
ERROR qlty_check::executor: Invocation failed for actionlint/lint: Failed to write invocation results ... No such file or directory (os error 2)
```

## Root cause

`Library::try_symlink_if_missing` gates symlink creation on `!link.exists()`. `Path::exists()` follows symlinks, so a dangling symlink reports `false` — but the link node still exists, so `symlink()` fails with `EEXIST`. That error is only logged, so the run proceeds with the broken link in place and every invocation fails writing through it.

## Fix

Inspect the link node itself with `symlink_metadata()`. If it is a symlink whose target is gone, remove it and recreate it pointing at the current cache directory. Valid symlinks and real files/directories at the path are left untouched, preserving existing behavior.

## Testing

- Added unit tests for `try_symlink_if_missing` covering: link missing, dangling link (the bug), valid link, and a real directory at the link path. Gated `#[cfg(unix)]` since symlink creation on Windows CI requires elevated privileges.
- Verified end-to-end: reproduced the failure in a workspace whose `.qlty/out` and `.qlty/logs` pointed at a deleted cache hash — the released binary fails with the FATAL error above; the patched binary heals the symlinks and completes with `✔ No issues`.

### DIFF
--- a/qlty-config/src/library.rs
+++ b/qlty-config/src/library.rs
@@ -254,29 +254,39 @@ impl Library {
     }
 
     fn try_symlink_if_missing(&self, target: &Path, link: &Path) -> Result<()> {
-        if !link.exists() {
-            #[cfg(unix)]
-            {
-                if let Err(err) = std::os::unix::fs::symlink(target, link) {
-                    error!(
-                        "Failed to create symlink from {} to {}: {}",
-                        target.display(),
-                        link.display(),
-                        err
-                    );
-                }
+        // Path::exists() follows symlinks, so a dangling symlink (e.g. one left
+        // behind after the cache directory was deleted) reports false while still
+        // blocking symlink creation with EEXIST. Check the link node itself and
+        // remove it if it is a symlink whose target is gone.
+        if let Ok(metadata) = link.symlink_metadata() {
+            if metadata.file_type().is_symlink() && !link.exists() {
+                fs::remove_file(link)?;
+            } else {
+                return Ok(());
             }
+        }
 
-            #[cfg(windows)]
-            {
-                if let Err(err) = std::os::windows::fs::symlink_dir(target, link) {
-                    error!(
-                        "Failed to create symlink from {} to {}: {}",
-                        target.display(),
-                        link.display(),
-                        err
-                    );
-                }
+        #[cfg(unix)]
+        {
+            if let Err(err) = std::os::unix::fs::symlink(target, link) {
+                error!(
+                    "Failed to create symlink from {} to {}: {}",
+                    target.display(),
+                    link.display(),
+                    err
+                );
+            }
+        }
+
+        #[cfg(windows)]
+        {
+            if let Err(err) = std::os::windows::fs::symlink_dir(target, link) {
+                error!(
+                    "Failed to create symlink from {} to {}: {}",
+                    target.display(),
+                    link.display(),
+                    err
+                );
             }
         }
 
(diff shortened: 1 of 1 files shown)


noahd1:
🔎 **BC check started**

Running backwards-compatibility analysis on this PR. Results will be posted here when complete.

---

noahd1:

**Target:** PR #2805 — `fix(check): recreate dangling .qlty symlinks after cache deletion`
**Verdict:** safe
**Affected user segments:** none (fix is purely restorative)

### Findings

None. This change is a targeted bug fix scoped to `Library::try_symlink_if_missing` in `qlty-config/src/library.rs`. I walked through each observable case:

| State at `link` path | Before | After | Δ |
|---|---|---|---|
| Nothing there | `link.exists()` false → create symlink | `symlink_metadata()` Err → create symlink | same |
| Valid symlink | `link.exists()` true → no-op, return Ok | `symlink_metadata()` Ok, `is_symlink()` true, `link.exists()` true → else branch → return Ok | same |
| Real file/dir at link path | `link.exists()` true → no-op, return Ok | `symlink_metadata()` Ok, `is_symlink()` false → else branch → return Ok | same |
| **Dangling symlink** (the bug) | `link.exists()` false → try to create → EEXIST → error logged, function returns Ok but link still broken → downstream `FATAL error occurred running N invocations` | Detects dangling link via `symlink_metadata`, removes it, recreates pointing at current cache dir | **fixed** |

- **a. Behavior changes on existing command paths** — the only behavior change is that a dangling `.qlty/out` / `.qlty/logs` symlink is now healed instead of causing a FATAL downstream. No user could rely on the previous broken state being preserved.
- **b. CLI flag surface** — no CLI changes.
- **c. `qlty.toml` schema** — no config changes.
- **d. Output formats** — no output changes.
- **e. Exit codes** — one theoretical edge: `fs::remove_file(link)?` now propagates a hard error if removal of a dangling symlink fails (e.g. permission denied). That case previously reached the `FATAL error` path anyway, so it's not a real regression.
- **f. Environment variables** — none touched.
- **g. Source fetching / cache layout** — layout unchanged; the fix only reconciles broken links produced by a user deleting `~/.qlty/cache/repos`.
- **h. Plugin behavior** — unchanged.
- **i. Telemetry / logging** — unchanged.

### Segment analysis

- **Local devs** — benefit: `qlty check` no longer FATALs after `rm -rf ~/.qlty/cache/repos`.
- **CI users (`qlty coverage *`)** — cache dir is typically absent between runs; unaffected.
- **Qlty Cloud** — `qlty sources fetch`, `config validate`, `fmt --skip-source-fetch`, `init`, `install`, `build` all call `Library::create_local` at startup, so this code path runs. Behavior for the states Qlty Cloud actually reaches (no link / valid link / real dir) is byte-for-byte identical to before. Only the "dangling symlink" state changes, and that state can only arise from a prior local run whose cache was later deleted — not a state Qlty Cloud ever produces.

### Notes for reviewer

- Test coverage is well-matched to the fix: four `#[cfg(unix)]` unit tests cover missing / dangling / valid / real-directory states.
- The `?` on `fs::remove_file(link)` inside the dangling-detection block is fine — the surrounding `create_local` already propagates via `?`, and any failure there was previously masked by the silent `error!` log.

---

PR #2837, opened 2026-08-11
Prevent filtered checks from polluting the issue cache

brynary:
## Summary

- Include active check filters in issue cache keys.
- Use the same filter-aware key for cache reads and writes.
- Add a CLI regression test for a filtered formatter check followed by an unfiltered lint check.

## Root cause

Qlty applies issue filters before it writes a cache-enabled linter result. A formatter-only filter can remove the linter issues and cache the empty result. The cache key did not include the active filters, so a later unfiltered check could reuse that empty result and incorrectly report no issues.

The regression test reproduced this before the fix: the second, unfiltered check exited 0 and printed No issues. It now exits 1 and reports the lint issue.

## Verification

- cargo fmt --check --all
- cargo nextest run --locked --workspace --all-targets --all-features --no-fail-fast
  - 1,096 passed; 2 skipped
- Focused regression test failed before the implementation and passed after it

## Existing unrelated failures

- Clippy with Rust 1.95 reports existing warnings as errors in unchanged files, beginning at qlty-plugins/src/lib.rs:8.
- Workspace doctests fail in the existing qlty-analysis/src/utils/fs.rs example because join_path_string is not imported.

### DIFF
--- a/qlty-check/src/cache.rs
+++ b/qlty-check/src/cache.rs
@@ -1,6 +1,7 @@
 use crate::planner::config_files::PluginConfigFile;
 use crate::planner::target::Target;
 use crate::tool::Tool;
+use crate::CheckFilter;
 use anyhow::{Context, Result};
 use git2::{Repository, Status};
 use itertools::Itertools;
@@ -104,6 +105,7 @@ struct InvocationCacheKey {
     tool: Box<dyn Tool>,
     plugin: Arc<PluginDef>,
     driver_name: String,
+    filters: Vec<CheckFilter>,
     affects_cache: HashMap<PathBuf, String>,
     configs: Arc<Vec<PluginConfigFile>>,
 }
@@ -247,6 +249,10 @@ impl InvocationCacheKey {
         digest.add("tool", &self.tool.directory());
         digest.add("driver_name", &self.driver_name);
 
+        for filter in &self.filters {
+            digest.add(&format!("check.filter.{filter}"), "true");
+        }
+
         for config in self.configs.clone().iter().sorted() {
             digest.add(&config.path.to_string_lossy(), &config.contents);
         }
@@ -264,6 +270,7 @@ impl IssuesCacheKey {
         tool: Box<dyn Tool>,
         plugin: Arc<PluginDef>,
         driver_name: String,
+        filters: Vec<CheckFilter>,
         configs: Arc<Vec<PluginConfigFile>>,
         affects_cache: Vec<String>,
     ) -> Self {
@@ -290,6 +297,7 @@ impl IssuesCacheKey {
                 tool: tool.clone(),
                 plugin: plugin.clone(),
                 driver_name,
+                filters,
                 affects_cache: cache_busters,
                 configs,
             }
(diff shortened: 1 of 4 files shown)


noahd1:
🔎 **Backwards-compatibility check in progress**

I'm analyzing this PR for behavior changes that could break existing qlty users on upgrade. I'll edit this comment with the findings when I'm done.

---

noahd1:

**Target:** PR #2837 — _Prevent filtered checks from polluting the issue cache_
**Verdict:** safe
**Affected user segments:** none required to act; a narrow observable-behavior change is noted for reviewer awareness

### Findings

**[note]** `qlty-check/src/cache.rs:247` — behavior change on existing command paths (cache correctness fix)
- Before: `InvocationCacheKey::digest_key` did not include the active `--filter` list, so a filtered run (e.g. `qlty check --filter=<plugin>:fmt`) could write a cache entry with issues stripped by the filter, and a later unfiltered `qlty check` would reuse that entry and report no issues.
- After: the digest now folds each active `CheckFilter` into the hash (`digest.add("check.filter.{filter}", "true")`). Filtered and unfiltered runs of the same driver now have distinct cache keys, so the unfiltered run re-executes and surfaces the real issues.
- Who breaks: **no segment breaks in the BC sense.** A user whose CI happens to run a filtered `qlty check` followed by an unfiltered one on the same commit will, on upgrade, see the unfiltered run correctly re-report previously-suppressed issues — i.e. it may exit `1` where the buggy cache reuse used to exit `0`. That is a bug fix, not a regression. The digest is unchanged when `filters` is empty (the loop is a no-op), so users who never pass `--filter` see byte-identical cache keys and keep their existing cache entries.
- Suggested mitigation: none required. Optional callout in the release notes so anyone whose green CI was propped up by the stale-cache bug knows why the next run may light up. No coordinated Qlty Cloud update needed — the affected commands are on the `qlty check` cache path, not on the `qlty sources fetch` / `qlty config validate` / `qlty fmt --skip-source-fetch` / `qlty init` / `qlty install` / `qlty build` paths.

- CLI flag surface (`qlty-cli/src/commands/**`) — no `#[arg]` / `#[clap]` added, removed, renamed, or defaulted differently.
- `qlty.toml` schema (`qlty-config/src/config/**`) — untouched.
- Output formats (`qlty-coverage/src/print.rs`, JSON printers, `qlty-types` protos) — untouched.
- Exit codes — no new `bail!` / `ensure!` / `?` propagation into `main`; the only exit-code delta is the intended correctness fix noted above.
- Environment variables — no `std::env::var` changes; `QLTY_*` surface unchanged.
- Source fetching / cache layout — no changes to `.qlty/sources/`, `.qlty/cache/`, or `~/.qlty/` directory layout; the on-disk cache key hash for filtered runs changes, which invalidates prior filtered-run cache entries. That is the intended fix and does not break unfiltered users.
- Plugin behavior — no `plugin.toml`, driver, or download-URL changes.
- Telemetry / logging defaults — untouched.

### Notes for reviewer

- The filter iteration order comes from `Vec<CheckFilter>` as passed through `Settings.filters`. If two invocations pass logically-equivalent filter sets in different orders, they will hash to different keys. That is a very minor cache-efficiency footgun rather than a BC break, and it errs on the safe side (cache miss, never a stale hit). Consider sorting before hashing if you want stable keys across argument orderings.
- The regression test `filtered_check_does_not_pollute_unfiltered_issue_cache` spawns the built qlty binary via `CARGO_BIN_EXE_qlty`. That's a good end-to-end guardrail and I'd keep it.

---

Issue #2854, opened 2026-09-21
Gitleaks creates one invocation per workspace file and can cause build timeouts


laura-mlg:
## Summary

While investigating repeated default-branch build timeouts for Storecove's `datajust` project, I found that the built-in Gitleaks driver created one process invocation per target/file.

In the captured build:
- 6,253 total plugin invocations
- 5,922 Gitleaks invocations
- Gitleaks represented about 82% of cumulative plugin execution time (3,844 seconds)
- The normal runner failed during processing after about 1,043 seconds, with Qlty process memory reaching about 10 GB
- The same build completed only after temporarily increasing CPU and RAM

The current definition uses `batch = false` and the implicit default file target, so each workspace file becomes a separate command:

`gitleaks detect --no-git --source=<file> ...`

The customer has since removed Gitleaks (and TruffleHog) from their configuration, and their latest master build passes, but this reduces security coverage.

## Question

This is not suggesting every plugin should run exactly once: for example, ESLint can produce several invocations while batching many targets into each. The concern here is Gitleaks’ near-1:1 workspace-entry-to-process fan-out for a scanner that can scan a directory itself.

Is one Gitleaks invocation per file intended or preferred? Since Gitleaks supports scanning a directory, should the built-in driver use one repository-level target instead?

I tested this project-level override:

```toml
[plugins.definitions.gitleaks.drivers.lint]
target = { type = "literal", path = "." }
```

It produced a single `gitleaks detect --source=.` invocation and still reported a nested-file finding correctly.

One caveat is that root-target scanning may require exclusions to live in native `.gitleaks.toml` rather than relying on Qlty per-file exclusions.

## References

- Customer: Storecove / `datajust`
- Workspace ID: `c68f2f74-f1f6-48b9-9df8-c88e15a3bb7b`
- Project ID: `5f52201c-2717-4996-b45c-65125abde2a8`
- Failed build: `01a0b05e-2e42-72c5-a4ea-eb2bce8ad518`
- Successful high-resource diagnostic build: `01a0b39a-9be1-7650-a49a-3b6eb26759c5`
- Latest build after scanner removal: `01a0c2c6-e23b-74e3-a23c-b2f74a351c36`

marschattha:
Well this was one of the earliest plugins we added, it was intentionally not batched I believe.

PR #2856, opened 2026-09-21
perf(check): compute repository cache state once per run

marschattha:
## Problem

`IssuesCacheKey::new` opens the git repository and runs a full `git status` over the working tree (`cache.rs`, `collect_dirty_paths`). `InvocationResult::cache_issues` builds one of these **per invocation**, so the cost scales as O(invocations × working tree size).

Plugins with `batch = false` run one invocation per file, which makes this quadratic. gitleaks is the worst case: `file_types = ["ALL"]` plus `batch = false` means invocations equals file count.

The planner already does this correctly — it builds the key once per driver (`planner/driver.rs`) and then only calls the cheap `finalize()` per target. The executor rebuilds the expensive part every time.

Two aggravating details:

- `--no-cache` does not avoid it. It swaps in `NullCache`, but `cache_issues()` is gated only on `driver.cache_results`, so the key is still built and the walk still runs before the write becomes a no-op.
- On macOS, libgit2's filesystem iterator calls `iconv_open` per directory, which serializes on a process-wide rwlock. With `jobs = 8` this left only 0–2 linter processes alive at a time.

Profiling a 2,480 file repository with only gitleaks enabled put **85% of samples** on `run_invocation` → `cache_issues` → `IssuesCacheKey::new` → `collect_dirty_paths` → `git2::Repository::statuses`.

## Change

Hoist the repository facts into a `RepositoryState` computed once when the `IssueCache` is built, shared by `Arc`. Both key sites already hold an `IssueCache`, so nothing new is threaded through `InvocationPlan`.

Deliberately **not** a `OnceLock` static: qlty's tests build sample repos in temp dirs and run in parallel in one process, so a process-global memo would leak one test's repository state into another's.

## Results

Same source, two release binaries, 2,480 file repository, gitleaks only:

| | wall | user | sys | peak RSS | issues |
|---|---|---|---|---|---|
| baseline `--no-cache` | 125.2s | 138.4 | **255.9** | 1.72 GB | 2 |
| **patched** `--no-cache` | **31.2s** | 69.2 | **44.0** | 1.59 GB | 2 |
| baseline cached (cold) | 113.6s | 136.3 | **254.5** | 1.48 GB | 2 |
| **patched** cached (cold) | **33.2s** | 69.7 | **43.4** | 1.43 GB | 2 |

**4.0× and 3.4× faster wall clock, system time down 5.8×.** Identical output in all four runs.

## Behavior

Output and cache digests are unchanged. The hashed `InvocationCacheKey` contains no repository fields and `add_target_sha` is untouched, so **existing cache entries stay valid** — no invalidation on upgrade.

Two real changes worth reviewer attention:

1. **One status walk now always runs**, in `Planner::new()`. Paths that previously did zero — `qlty fmt`, `qlty validate`, `--no-cache`, literal-target-only configs — now do one. Measured cost in the test suite: nil. Avoidable with `Arc<OnceLock<RepositoryState>>` if preferred; note the naive `OnceLock<Arc<…>>` on a `Clone` struct would be wrong, since `IssueCache` is cloned per invocation.
2. **The snapshot moves from execute time to plan time.** This makes read keys (planner) and write keys (executor) consistent, where before they could disagree and write entries that would never be read back. The cost is that the window in which a file edited mid-run is missed grows from the tail of the run to the whole run.

Safe because the working tree does not change mid-run: `Patcher` runs in the processor after execution, and formatters write to a staging area rather than the workspace.

## Tests

`cache.rs` had no test module at all, so `add_target_sha` was uncovered. Adds 5 tests, each verified by mutation:

- `test_cloned_cache_shares_repository_state` — clones the cache *before* asking for state, so it catches a future lazy rewrite that recomputes per clone. Fails when `repository()` is made to recompute.
- Four tests pinning `add_target_sha`: clean path digests the tree sha; dirty path, dirty parent directory, and missing sha all fall back to mtime. Three fail when the dirty-paths guard is inverted.

Full suite: 1,100 passed, 0 failed. `cargo check`, `qlty fmt`, and `qlty check --level=low --fix` all clean.

## Scope

This is one of two independent quadratics. Memory barely moves (8%) because the other one is untouched: `Driver::parse` emits one `FileResult` per workspace entry per invocation and retains them all, which is what drives the ~10 GB in #2854. That needs its own change.

This also does not close the gap to a batched scanner — 31s for 2,480 invocations is still ~44× the 0.71s a single gitleaks directory scan takes. The gitleaks target-shape question in #2854 remains open on its own merits; this just means it is no longer load-bearing for whether large repositories can finish at all.

One pre-existing bug left alone: `RepositoryState::compute()` still uses `Workspace::new()`, which resolves from cwd rather than `settings.root`, so `qlty check --root /other/repo` keys off the wrong repository. Fixing it would change cache keys for `--root` users.

Refs #2854


noahd1:

**Target:** PR #2856 — `perf(check): compute repository cache state once per run`
**Verdict:** risky — needs mitigation (rebase before merge)
**Affected user segments:** local devs, CI users (any user running `qlty check --filter …`)

### Findings

**[risky]** branch is out-of-date with `main`; missing #2837 (`filters` in cache key)

- Before (current `main`, post #2837): `IssuesCacheKey` includes `filters: Vec<CheckFilter>` and hashes each active `--filter` into the digest. A formatter-only filtered run and an unfiltered lint run get different cache keys, so the empty filtered result cannot poison the unfiltered read.
- After (this PR merged as-is): the PR was branched from `0a789a1`, one commit before #2837 landed (`372a928`). `git diff origin/main…HEAD` (three-dot, what GitHub shows) does not include the filter code because the merge-base predates it, so the file rewrite in this PR reintroduces an `InvocationCacheKey` without any `filters` field. `git diff origin/main HEAD` (two-dot) confirms the delta: `-    filters: Vec<CheckFilter>,` and `-        for filter in &self.filters { digest.add(&format!("check.filter.{filter}"), "true"); }` are removed relative to current `main`. Cache keys stop discriminating on active filters, and the exact regression #2837 fixed (a filtered check writes an empty result under a key an unfiltered run then reads, and reports zero issues) comes back.
- Who breaks: **segment 1 (local devs)** and **segment 2 (CI users)** running `qlty check --filter …` or with any workflow that mixes filtered and unfiltered runs against the same cache — including the CLI regression test #2837 added. Cache-enabled linters affected; scenario: `qlty check --filter format` writes empty issues for a driver, then `qlty check` reads that entry and reports no lint issues on a file that actually has some.
- Suggested mitigation: rebase this branch on current `origin/main` and re-apply the perf refactor on top of #2837's `filters` field (add `filters: Vec<CheckFilter>` back to the new `InvocationCacheKey`, keep the `for filter in &self.filters { digest.add(...) }` loop in `build()`, and thread `self.settings.filters.clone()` / `self.plan.settings.filters.clone()` through both `IssuesCacheKey::new` call sites alongside the new `repository` arg). The three-dot diff hides this — reviewers should read `git diff origin/main HEAD` before merging.

**[note]** `qlty-check/src/cache.rs:111` — one `git status` walk now runs on command paths that previously did none

- Before: `qlty fmt`, `qlty validate`, `qlty check --no-cache`, and literal-target-only configs never constructed an `IssuesCacheKey`, so `collect_dirty_paths` never ran.
- After: `IssueCache::new` eagerly calls `RepositoryState::compute()`, which walks the working tree once regardless of whether any cache key will later be built.
- Who breaks: nobody functionally — same outputs, exit codes, and cache digests. Author measured cost as nil in the test suite. Flagging so reviewers know this is intentional and expected in the wall-clock of the affected subcommands.
- Suggested mitigation: none required. If ever measured as material, wrap `repository` in `Arc<OnceLock<RepositoryState>>` per the PR description (not the naive `OnceLock<Arc<…>>` on a `Clone` struct).

**[note]** `qlty-check/src/planner/driver.rs:165` — repository-state snapshot moves from execute-time to plan-time

- Before: `IssuesCacheKey::new` was called inside the executor, so `dirty_paths` and `tree_sha` reflected the working tree at execute time.
- After: the snapshot is computed once when `IssueCache` is built (planner startup) and shared by both plan-time reads and execute-time writes.
- Who breaks: no observable break. Read/write keys are now consistent (previously they could disagree and write entries that would never be read back — the PR calls this out as an improvement). The one behavior change: the window in which a file edited mid-run is missed grows from the tail of the run to the whole run. Safe because `Patcher` runs after execution and formatters write to a staging area.
- Suggested mitigation: none.

**[note]** `qlty-check/src/cache.rs:112` — pre-existing: `Workspace::new()` resolves from cwd, not `settings.root`

- Before: same behavior — the pre-hoist code also used `Workspace::new()`, so `qlty check --root /other/repo` already keyed off the wrong repository.
- After: unchanged. The PR explicitly calls this out and leaves it alone because fixing it would invalidate cache keys for `--root` users.
- Who breaks: no new regression from this PR; noting for completeness because the code moved.
- Suggested mitigation: none in scope for this PR.

- CLI flag surface (no `#[arg]` / `#[clap]` changes)
- qlty.toml schema (no serde changes)
- Output formats (PR verified identical output across 4 runs; no `qlty-types` / `print_*` changes)
- Exit codes (no `exit`, `CommandError`, or `?` propagation changes)
- Environment variables (no `std::env::var` sites touched)
- Source fetching / cache layout (`.qlty/cache/` on-disk format unchanged; hashed `InvocationCacheKey` fields unchanged relative to the PR's own base, though see finding 1 for the delta vs. current `main`)
- Plugin behavior (no `plugin.toml` or loader changes)
- Telemetry / logging defaults (no log-level changes; one `trace!` unchanged)

### Notes for reviewer
- Do not evaluate the diff GitHub renders in isolation — it uses three-dot semantics against the merge-base (`0a789a1`) and therefore hides that the PR predates #2837. Read `git diff origin/main HEAD -- qlty-check/src/cache.rs` locally before approving.
- If the resolution is "rebase and re-thread `filters`," add the CLI regression test from #2837 to the pre-merge validation list so a green run demonstrates that the pollution fix survived the perf refactor.
- Segment 3 (Qlty Cloud) is not in scope for this PR — none of `qlty sources fetch`, `qlty config validate`, `qlty fmt --skip-source-fetch`, `qlty init`, `qlty install`, or `qlty build` build an `IssueCache`, so the filter regression does not affect the hosted-build path.

---

noahd1 (approved):
Nice find. Approving to unblock but I would double check the note in the backwards compatibility check. Since it looks like code merged after this PR opened. Unsure of the impact of that.

marschattha:
> Nice find. Approving to unblock but I would double check the note in the backwards compatibility check. Since it looks like code merged after this PR opened. Unsure of the impact of that.

Yea rebased after that comment, I don't think it holds now.

